PR #10, opened 2014-06-04
initial BasicAuthExhibitorRestClient 

sclasen:
that can also skip ssl cert validation if required

Also might be nice to provide a convenience method in ExhibitorEnsembleProvider, something like

``` java
pubic static ExhibitorEnsembleProvider fromURL(URL vip, boolean validateSsl)
```

which would create the `Exhibitors` from the hostname, extract the userInfo, port and path from the url


tisonkun:

PR #50, opened 2014-10-29
CURATOR-154 - Modified the PersistentEphemeralNode handling

cammckenzie:
CURATOR-154 - Modified the handling for creating the ephemeral node so that if it already exists, an attempt will be made to set its data to match the data that the PersistentEphemeralNode has cached.


madrob:
Not a committer, but +1, LGTM.

Randgalt:
+1

PR #56, opened 2014-12-01
Curator 161

cammckenzie:
First cut of watcher removal. Still need to provide the ability to register for notification of watch removal.

### DIFF
--- a/curator-framework/src/main/java/org/apache/curator/framework/CuratorFramework.java
+++ b/curator-framework/src/main/java/org/apache/curator/framework/CuratorFramework.java
@@ -148,7 +148,7 @@ public interface CuratorFramework extends Closeable
      * Start a remove watches builder.
      * @return builder object
      */
-    public RemoveWatchesBuilder removeWatches();    
+    public RemoveWatchesBuilder watches();    
 
     /**
      * Returns the listenable interface for the Connect State
--- a/curator-framework/src/main/java/org/apache/curator/framework/CuratorFramework.java
+++ b/curator-framework/src/main/java/org/apache/curator/framework/CuratorFramework.java
@@ -216,7 +216,11 @@ public interface CuratorFramework extends Closeable
      * Call this method on watchers you are no longer interested in.
      *
      * @param watcher the watcher
+     * 
+     * @deprecated As of ZooKeeper 3.5 Curators recipes will handle removing watcher references
+     * when they are no longer used.
      */
+    @Deprecated
     public void clearWatcherReferences(Watcher watcher);
         
     /**
--- a/curator-framework/src/main/java/org/apache/curator/framework/CuratorFramework.java
+++ b/curator-framework/src/main/java/org/apache/curator/framework/CuratorFramework.java
@@ -24,7 +24,6 @@ import org.apache.curator.framework.api.*;
 import org.apache.curator.framework.api.transaction.CuratorTransaction;
 import org.apache.curator.framework.imps.CuratorFrameworkState;
 import org.apache.curator.framework.listen.Listenable;
-import org.apache.curator.framework.state.ConnectionState;
 import org.apache.curator.framework.state.ConnectionStateListener;
 import org.apache.curator.utils.EnsurePath;
 import org.apache.zookeeper.Watcher;
@@ -144,6 +143,12 @@ public interface CuratorFramework extends Closeable
      * @return builder object
      */
     public SyncBuilder sync();
+
+    /**
+     * Start a remove watches builder.
+     * @return builder object
+     */
+    public RemoveWatchesBuilder removeWatches();    
 
     /**
      * Returns the listenable interface for the Connect State
--- a/curator-framework/src/main/java/org/apache/curator/framework/api/BackgroundPathableQuietly.java
+++ b/curator-framework/src/main/java/org/apache/curator/framework/api/BackgroundPathableQuietly.java
@@ -1,0 +1,5 @@
+package org.apache.curator.framework.api;
+
+public interface BackgroundPathableQuietly<T> extends BackgroundPathable<T>, Quietly<BackgroundPathable<T>>
+{
+}
--- a/curator-framework/src/main/java/org/apache/curator/framework/api/BackgroundPathableQuietly.java
+++ b/curator-framework/src/main/java/org/apache/curator/framework/api/BackgroundPathableQuietly.java
@@ -1,5 +1,0 @@
-package org.apache.curator.framework.api;
-
-public interface BackgroundPathableQuietly<T> extends BackgroundPathable<T>, Quietly<BackgroundPathable<T>>
-{
-}
--- a/curator-framework/src/main/java/org/apache/curator/framework/api/DeleteBuilder.java
+++ b/curator-framework/src/main/java/org/apache/curator/framework/api/DeleteBuilder.java
@@ -18,6 +18,6 @@
  */
 package org.apache.curator.framework.api;
 
-public interface DeleteBuilder extends GuaranteeableDelete, ChildrenDeletable
+public interface DeleteBuilder extends GuaranteeableDeletable, ChildrenDeletable
 {
 }
--- a/curator-framework/src/main/java/org/apache/curator/framework/api/RemoveWatchesBuilder.java
+++ b/curator-framework/src/main/java/org/apache/curator/framework/api/RemoveWatchesBuilder.java
@@ -12,18 +12,18 @@ public interface RemoveWatchesBuilder
      * @param watcher
      * @return
      */
-    public RemoveWatchesType watcher(Watcher watcher);
+    public RemoveWatchesType remove(Watcher watcher);
     
     /**
      * Specify the watcher to be removed
      * @param watcher
      * @return
      */
-    public RemoveWatchesType watcher(CuratorWatcher watcher);
+    public RemoveWatchesType remove(CuratorWatcher watcher);
     
     /**
      * Specify that all watches should be removed
      * @return
      */
-    public RemoveWatchesType allWatches();
+    public RemoveWatchesType removeAll();
 }
--- a/curator-framework/src/main/java/org/apache/curator/framework/api/RemoveWatchesBuilder.java
+++ b/curator-framework/src/main/java/org/apache/curator/framework/api/RemoveWatchesBuilder.java
@@ -1,0 +1,29 @@
+package org.apache.curator.framework.api;
+
+import org.apache.zookeeper.Watcher;
+
+/**
+ * Builder to allow watches to be removed 
+ */
+public interface RemoveWatchesBuilder
+{
+    /**
+     * Specify the watcher to be removed
+     * @param watcher
+     * @return
+     */
+    public RemoveWatchesType watcher(Watcher watcher);
+    
+    /**
+     * Specify the watcher to be removed
+     * @param watcher
+     * @return
+     */
+    public RemoveWatchesType watcher(CuratorWatcher watcher);
+    
+    /**
+     * Specify that all watches should be removed
+     * @return
+     */
+    public RemoveWatchesType allWatches();
+}
--- a/curator-framework/src/main/java/org/apache/curator/framework/api/RemoveWatchesLocal.java
+++ b/curator-framework/src/main/java/org/apache/curator/framework/api/RemoveWatchesLocal.java
@@ -10,9 +10,8 @@ public interface RemoveWatchesLocal extends BackgroundPathable<Void>
     /**
      * Specify if the client should just remove client side watches if a connection to ZK
      * is not available.
-     * @param local
      * @return
      */
-    public BackgroundPathable<Void> local(boolean local);
+    public BackgroundPathable<Void> local();
     
 }
--- a/curator-framework/src/main/java/org/apache/curator/framework/api/RemoveWatchesLocal.java
+++ b/curator-framework/src/main/java/org/apache/curator/framework/api/RemoveWatchesLocal.java
@@ -1,0 +1,18 @@
+package org.apache.curator.framework.api;
+
+/**
+ * Builder to allow the specification of whether it is acceptable to remove client side watch information
+ * in the case where ZK cannot be contacted. 
+ */
+public interface RemoveWatchesLocal extends BackgroundPathable<Void>
+{
+   
+    /**
+     * Specify if the client should just remove client side watches if a connection to ZK
+     * is not available.
+     * @param local
+     * @return
+     */
+    public BackgroundPathable<Void> local(boolean local);
+    
+}
--- a/curator-framework/src/main/java/org/apache/curator/framework/api/RemoveWatchesLocal.java
+++ b/curator-framework/src/main/java/org/apache/curator/framework/api/RemoveWatchesLocal.java
@@ -9,7 +9,7 @@ public interface RemoveWatchesLocal extends BackgroundPathableQuietly<Void>
    
     /**
      * Specify if the client should just remove client side watches if a connection to ZK
-     * is not available.
+     * is not available. Note that the standard Curator retry loop will not be used in t
      * @return
      */
     public BackgroundPathableQuietly<Void> locally();
--- a/curator-framework/src/main/java/org/apache/curator/framework/api/RemoveWatchesLocal.java
+++ b/curator-framework/src/main/java/org/apache/curator/framework/api/RemoveWatchesLocal.java
@@ -4,7 +4,7 @@ package org.apache.curator.framework.api;
  * Builder to allow the specification of whether it is acceptable to remove client side watch information
  * in the case where ZK cannot be contacted. 
  */
-public interface RemoveWatchesLocal extends BackgroundPathableQuietly<Void>
+public interface RemoveWatchesLocal extends BackgroundPathableQuietlyable<Void>
 {
    
     /**
@@ -12,6 +12,6 @@ public interface RemoveWatchesLocal extends BackgroundPathableQuietly<Void>
      * is not available. Note that the standard Curator retry loop will not be used in t
      * @return
      */
-    public BackgroundPathableQuietly<Void> locally();
+    public BackgroundPathableQuietlyable<Void> locally();
     
 }
--- a/curator-framework/src/main/java/org/apache/curator/framework/api/RemoveWatchesLocal.java
+++ b/curator-framework/src/main/java/org/apache/curator/framework/api/RemoveWatchesLocal.java
@@ -4,7 +4,7 @@ package org.apache.curator.framework.api;
  * Builder to allow the specification of whether it is acceptable to remove client side watch information
  * in the case where ZK cannot be contacted. 
  */
-public interface RemoveWatchesLocal extends BackgroundPathable<Void>
+public interface RemoveWatchesLocal extends BackgroundPathableQuietly<Void>
 {
    
     /**
@@ -12,6 +12,6 @@ public interface RemoveWatchesLocal extends BackgroundPathable<Void>
      * is not available.
      * @return
      */
-    public BackgroundPathable<Void> local();
+    public BackgroundPathableQuietly<Void> local();
     
 }
--- a/curator-framework/src/main/java/org/apache/curator/framework/api/RemoveWatchesLocal.java
+++ b/curator-framework/src/main/java/org/apache/curator/framework/api/RemoveWatchesLocal.java
@@ -12,6 +12,6 @@ public interface RemoveWatchesLocal extends BackgroundPathableQuietly<Void>
      * is not available.
      * @return
      */
-    public BackgroundPathableQuietly<Void> local();
+    public BackgroundPathableQuietly<Void> locally();
     
 }
--- a/curator-framework/src/main/java/org/apache/curator/framework/api/RemoveWatchesType.java
+++ b/curator-framework/src/main/java/org/apache/curator/framework/api/RemoveWatchesType.java
@@ -6,7 +6,7 @@ import org.apache.zookeeper.Watcher.WatcherType;
  * Builder to allow the specification of whether it is acceptable to remove client side watch information
  * in the case where ZK cannot be contacted. 
  */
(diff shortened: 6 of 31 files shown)


curator-framework/src/main/java/org/apache/curator/framework/CuratorFramework.java
```diff
@@ -1,233 +1,238 @@
-/**
```
Randgalt: Any idea why this looks like a complete replacement?

cammckenzie: My only guess is that it's some sort of whitespace issue?

curator-framework/src/main/java/org/apache/curator/framework/api/RemoveWatchesBuilder.java
```diff
@@ -0,0 +1,29 @@
...
+     * @param watcher
+     * @return
+     */
+    public RemoveWatchesType watcher(CuratorWatcher watcher);
+    
+    /**
+     * Specify that all watches should be removed
+     * @return
+     */
+    public RemoveWatchesType allWatches();
```
Randgalt: I'd rename this to something like "all". Try to make these build sentences: client.removeWatches().all()

curator-framework/src/main/java/org/apache/curator/framework/CuratorFramework.java
```diff
@@ -1,233 +1,238 @@
...
+     *
+     * @return builder object
+     */
+    public SyncBuilder sync();
+
+    /**
+     * Start a remove watches builder.
+     * @return builder object
+     */
+    public RemoveWatchesBuilder removeWatches();
```
Randgalt: Rename this to something like "watches" so we can get nice sentences: client.watches().removeAll()

curator-framework/src/main/java/org/apache/curator/framework/api/RemoveWatchesBuilder.java
```diff
@@ -0,0 +1,29 @@
...
+ * Builder to allow watches to be removed 
+ */
+public interface RemoveWatchesBuilder
+{
+    /**
+     * Specify the watcher to be removed
+     * @param watcher
+     * @return
+     */
+    public RemoveWatchesType watcher(Watcher watcher);
```
Randgalt: I'd rename this to something like "remove". Try to make these build sentences: client.watches().remove(w).

curator-framework/src/main/java/org/apache/curator/framework/api/RemoveWatchesBuilder.java
```diff
@@ -0,0 +1,29 @@
...
+    
+    /**
+     * Specify the watcher to be removed
+     * @param watcher
+     * @return
+     */
+    public RemoveWatchesType watcher(CuratorWatcher watcher);
```
Randgalt: I'd rename this to something like "remove". Try to make these build sentences: client.watches().remove(w).

Randgalt:
I think this change should also use the remove watches APIs in the various recipes. Or should that be a separate issue?

cammckenzie:
Thanks for looking at this. I think that maybe a new JIRA for the recipes.
(I think there's already one for the node cache recipes).
cheers

Randgalt:
OK - please open an issue for this. If the Node Cache issue is appropriate, rename it and target it for 3.0.0

curator-framework/src/main/java/org/apache/curator/framework/api/RemoveWatchesLocal.java
```diff
@@ -0,0 +1,18 @@
...
+public interface RemoveWatchesLocal extends BackgroundPathable<Void>
+{
+   
+    /**
+     * Specify if the client should just remove client side watches if a connection to ZK
+     * is not available.
+     * @param local
+     * @return
+     */
+    public BackgroundPathable<Void> local(boolean local);
```
Randgalt: the boolean doesn't seem necessary. If local() isn't called then it's not local.

Randgalt: I'd name this "locally" so that it reads right. E.g. {{client.watches().remove(w).ofType(t).locally().forPath(...)}}

cammckenzie:
I have made the changes based on your suggestions.

Was wondering if we wanted to have some sort of way of registering for notification of a watch being removed, or if we just do it via the CuratorListener?

Currently you need to register a CuratorListener and then check

if(event.getType() == CuratorEventType.WATCHED && event.getWatchedVent().getType() == EventType.DataWatchRemoved)
{
}

(or EventType.ChildWatchRemoved).

Maybe a new type of listener interface? Or is this overkill?

cammckenzie:
There's actually another interesting thing that I hadn't thought about. The 'local' flag lets a client remove watches while its not actually connected to ZK, but the current way that callWithRetry() works is that it blocks until the client is connected or timed out prior to executing the Callable.

I guess I need to add something to Callable to indicate that it should be called even if there's no connection to ZK available. Does that sound reasonable?

Randgalt:
You can do this to get the ZooKeeper handle directly:

```
client.getZookeeperClient().getZooKeeper()
```

This does not use the retry mechanism. So, when you are in "local" mode, don't do a retry. Just get the ZK handle and make the call. IMPORTANT: getZooKeeper() can legally return <code>null</code> in some cases.

curator-framework/src/main/java/org/apache/curator/framework/api/RemoveWatchesBuilder.java
```diff
@@ -0,0 +1,29 @@
+package org.apache.curator.framework.api;
```
Randgalt: This will need a guaranteed() method similar to node deletion. Use the FailedDeleteManager as a model (maybe extend it?).

curator-framework/src/main/java/org/apache/curator/framework/imps/RemoveWatchesBuilderImpl.java
```diff
@@ -0,0 +1,192 @@
...
+    private void pathInForeground(final String path) throws Exception
+    {
+        RetryLoop.callWithRetry(client.getZookeeperClient(), 
+                new Callable<Void>()
+                {
+                    @Override
+                    public Void call() throws Exception
+                    {
+                        ZooKeeper zkClient = client.getZooKeeper();
+                        if(watcher == null)
```
Randgalt: Please review. if watcher is null, removeAllWatches is done. If not null, removeWatches.

curator-framework/src/main/java/org/apache/curator/framework/imps/RemoveWatchesBuilderImpl.java
```diff
@@ -0,0 +1,192 @@
...
+                trace.commit();
+                CuratorEvent event = new CuratorEventImpl(client, CuratorEventType.REMOVE_WATCHES, rc, path, null, ctx, null, null, null, null, null);
+                client.processBackgroundOperation(operationAndData, event);                
+            }
+        };
+        
+        ZooKeeper zkClient = client.getZooKeeper();
+        if(watcher == null)
+        {
+            zkClient.removeAllWatches(operationAndData.getData(), watcherType, local, callback, operationAndData.getContext());
```
Randgalt: Please review. if watcher is null, removeAllWatches is done. If not null, removeWatches.

curator-framework/src/main/java/org/apache/curator/framework/api/RemoveWatchesBuilder.java
Randgalt: We should also add a <code>quietly()</code> method or something to swallow <code>KeeperException.NoWatcher</code> exceptions. I'll bet most users don't care if the watcher doesn't exist.

curator-framework/src/main/java/org/apache/curator/framework/imps/RemoveWatchesBuilderImpl.java
cammckenzie: This is reasonable yes? If we are specifying a particular watcher to remove then do that, otherwise remove all watches from the path.

curator-framework/src/main/java/org/apache/curator/framework/api/RemoveWatchesBuilder.java
cammckenzie: I think quietly() is a good idea as its explicit. Also, guaranteed() will never tell the user that the remove has failed, whereas quietly would indicate failure (except for NoWatcher exceptions).

cammckenzie: Under what conditions can getZookeeperClient().getZookeeper() return null? Is it when we're trying to reconnect?

What should be done under these conditions? Presumably we just need to throw an exception back to the user because  we can't execute the command.

curator-framework/src/main/java/org/apache/curator/framework/api/RemoveWatchesLocal.java
```diff
@@ -0,0 +1,17 @@
...
+ */
+public interface RemoveWatchesLocal extends BackgroundPathableQuietly<Void>
+{
+   
+    /**
+     * Specify if the client should just remove client side watches if a connection to ZK
+     * is not available.
+     * @return
+     */
+    public BackgroundPathableQuietly<Void> local();
```
Randgalt: In context, this would be better named "locally". E.g.:

```
client.watches().removeAll().ofType(WatcherType.Any).locally().forPath(path);
```

curator-framework/src/main/java/org/apache/curator/framework/api/RemoveWatchesType.java
```diff
@@ -0,0 +1,19 @@
...
+ */
+public interface RemoveWatchesType
+{
+   
+    /**
+     * Specify the type of watcher to be removed.
+     * @param watcherType
+     * @return
+     */
+    public RemoveWatchesLocal ofType(WatcherType watcherType);
```
Randgalt: Can this be made optional with a default of WatcherType.Any?

curator-framework/src/main/java/org/apache/curator/framework/api/RemoveWatchesLocal.java
cammckenzie: Done

curator-framework/src/main/java/org/apache/curator/framework/api/RemoveWatchesType.java
cammckenzie: Done, it will now accept:
client.watches().removeAll().forPath("/");

curator-framework/src/main/java/org/apache/curator/framework/api/RemoveWatchesBuilder.java
cammckenzie: Looking at the local case again, I think it's actually just client.getZookeeper() that needs to be used. I notice that this is how it's used in all of the performBackgroundOperation calls, but there are no checks for null pointers in this case. Is there a reason for this?

Randgalt: Maybe I mis-remembered. If the rest of the code doesn't check you don't need to.

curator-framework/src/main/java/org/apache/curator/framework/imps/RemoveWatchesBuilderImpl.java
```diff
@@ -0,0 +1,214 @@
...
+        RetryLoop.callWithRetry(client.getZookeeperClient(), 
+                new Callable<Void>()
+                {
+                    @Override
+                    public Void call() throws Exception
+                    {
+                        try
+                        {
+                            ZooKeeper zkClient = client.getZooKeeper();
+                            if(watcher == null)
```
Randgalt: Don't forget to look at these lines. It seems wrong to me.

curator-framework/src/main/java/org/apache/curator/framework/api/RemoveWatchesBuilder.java
cammckenzie: There are additional problems when trying the local case in the background as the 'performBackgroundOperation()' method checks if the client is connected prior to executing the operation.

curator-framework/src/main/java/org/apache/curator/framework/imps/RemoveWatchesBuilderImpl.java
```diff
@@ -0,0 +1,214 @@
+package org.apache.curator.framework.imps;
```
Randgalt: Once the builder is committed via forPath() the watcher should also be removed from NamespaceWatcherMap. We should deprecate clearWatcherReferences() too.

curator-framework/src/main/java/org/apache/curator/framework/imps/RemoveWatchesBuilderImpl.java
```diff
@@ -0,0 +1,214 @@
...
+        this.watcherType = null;
+        this.local = false;
+        this.quietly = false;
+        this.backgrounding = new Backgrounding();
+    }
+    
+    @Override
+    public RemoveWatchesType remove(Watcher watcher)
+    {
+        this.watcher = watcher == null ? null : client.getNamespaceWatcherMap().getNamespaceWatcher(watcher);
```
Randgalt: <code> client.getNamespaceWatcherMap().getNamespaceWatcher(watcher)</code> will add a namespace watcher record if it doesn't exist. That doesn't seem right for removal. A new method in NamespaceWatcherMap is probably needed.

cammckenzie: I guess if we don't have a reference in the NamespaceWatcherMap then we pass the watcher through to the ZK client without it being wrapper in a NamespaceWatcher. The only reason I can think of a watcher not being in the map would be if it was registered directly with the ZK client without going through the ZK framework. Either way, if it doesn't exist it will result in a NoWatcherException

curator-framework/src/main/java/org/apache/curator/framework/imps/RemoveWatchesBuilderImpl.java
cammckenzie: I don't think that this is necessarily true is it? We don't want to remove the Watcher reference itself because it may still be watching other paths. If it is no longer referenced then the WeakReference in the NamespaceWatcherMaps internal cache should handle this.

curator-framework/src/main/java/org/apache/curator/framework/imps/RemoveWatchesBuilderImpl.java
cammckenzie: What is it that seems wrong? If no Watcher is specified this means that we're removing all watches for a given path. If a Watcher is specified then this means we're removing a specific watch from a given path.

curator-framework/src/main/java/org/apache/curator/framework/api/RemoveWatchesBuilder.java
Randgalt: Internally, it looks like "local" mode acts only if the connection is lost. Maybe you can test the connection state when local is set and don't do retries, etc.

curator-framework/src/main/java/org/apache/curator/framework/imps/RemoveWatchesBuilderImpl.java
Randgalt: I was wrong. Ignore this.

curator-framework/src/main/java/org/apache/curator/framework/api/RemoveWatchesBuilder.java
cammckenzie: Looking at the ZK code I think that 'local' mode occurs whenever there's any sort of error.

curator-framework/src/main/java/org/apache/curator/framework/imps/RemoveWatchesBuilderImpl.java
Randgalt: Good point. You're right.

cammckenzie:
All functionality is now implemented. Ready for a final review.

curator-framework/src/main/java/org/apache/curator/framework/api/DeleteBuilder.java
```diff
@@ -18,6 +18,6 @@
  */
 package org.apache.curator.framework.api;
 
-public interface DeleteBuilder extends Guaranteeable, ChildrenDeletable
+public interface DeleteBuilder extends GuaranteeableDelete, ChildrenDeletable
```
Randgalt: A nit, but the naming convention is actually: <Verb1>..<VerbN>able. So it would be GuaranteeDeletable.

curator-framework/src/main/java/org/apache/curator/framework/api/BackgroundPathableQuietly.java
```diff
@@ -0,0 +1,5 @@
+package org.apache.curator.framework.api;
+
+public interface BackgroundPathableQuietly<T> extends BackgroundPathable<T>, Quietly<BackgroundPathable<T>>
```
Randgalt: Naming convention nit: should be "BackgroundPathQuietlyable" - Yes, I'm probably going insane.

curator-framework/src/main/java/org/apache/curator/framework/imps/CuratorFrameworkImpl.java
```diff
@@ -807,7 +821,7 @@ private void performBackgroundOperation(OperationAndData<?> operationAndData)
     {
         try
         {
-            if ( client.isConnected() )
+            if ( !operationAndData.isConnectionRequired() || client.isConnected() )
```
Randgalt: Why this change?

curator-framework/src/main/java/org/apache/curator/framework/imps/FailedDeleteManager.java
```diff
@@ -19,45 +19,18 @@
...
-import org.slf4j.Logger;
-import org.slf4j.LoggerFactory;
 
-class FailedDeleteManager
+class FailedDeleteManager extends FailedOperationManager<String>
 {
-    private final Logger log = LoggerFactory.getLogger(getClass());
-    private final CuratorFramework client;
-    
-    volatile FailedDeleteManagerListener debugListener = null;
```
Randgalt: nit: mark @VisibleForTesting

curator-framework/src/main/java/org/apache/curator/framework/imps/FailedRemoveWatchManager.java
```diff
@@ -0,0 +1,56 @@
...
+    FailedRemoveWatchManager(CuratorFramework client)
+    {
+        super(client);
+    }
+
+    @Override
+    protected void executeGuaranteedOperationInBackground(FailedRemoveWatchDetails details)
+            throws Exception
+    {
+        if(details.watcher ==  null)
```
Randgalt: nit: formatting

curator-framework/src/main/java/org/apache/curator/framework/imps/NamespaceWatcherMap.java
```diff
@@ -70,6 +70,14 @@ NamespaceWatcher remove(Object key)
     {
         return map.remove(key);
     }
+    
+    /**
+     * Remove all watchers for a given path
+     * @param path
+     */
+    void removeAllForPath(String path) {
+
```
Randgalt: Why is this empty?

curator-recipes/src/test/java/org/apache/curator/framework/recipes/leader/ChaosMonkeyCnxnFactory.java
```diff
@@ -59,13 +59,6 @@ public void startup(ZooKeeperServer zks) throws IOException, InterruptedExceptio
         super.startup(new ChaosMonkeyZookeeperServer(zks));
     }
 
-    /**
```
Randgalt: Why is this removed?

curator-framework/src/main/java/org/apache/curator/framework/imps/NamespaceWatcherMap.java
cammckenzie: Looks like I started doing something with the NamespaceWatcherMap and then realised that I shouldn't. Will remove this method, it's not used anywhere.

curator-recipes/src/test/java/org/apache/curator/framework/recipes/leader/ChaosMonkeyCnxnFactory.java
cammckenzie: Not sure? I don't see this change in my revision history? Last update was 6 months ago in the upgrade to ZK 3.5.0-alpha by Ioannis?

curator-framework/src/main/java/org/apache/curator/framework/api/DeleteBuilder.java
cammckenzie: Ok, changed.

curator-framework/src/main/java/org/apache/curator/framework/api/BackgroundPathableQuietly.java
cammckenzie: Changed. I have never written Quitelyable before!

curator-framework/src/main/java/org/apache/curator/framework/imps/FailedRemoveWatchManager.java
cammckenzie: Removed the extra space.

Randgalt:
LGTM

cammckenzie:
Ok, thanks will merge into the 3.0.0 branch.

aaylward:
can this be closed?

PR #59, opened 2014-12-29
CURATOR-111: Add authorization(List<AuthInfo)

kambatla:
I was able to run the test through IntelliJ, but had trouble building and testing on command line.

### DIFF
--- a/curator-framework/src/main/java/org/apache/curator/framework/imps/CuratorFrameworkImpl.java
+++ b/curator-framework/src/main/java/org/apache/curator/framework/imps/CuratorFrameworkImpl.java
@@ -24,6 +24,7 @@ import org.apache.curator.CuratorConnectionLossException;
 import org.apache.curator.CuratorZookeeperClient;
 import org.apache.curator.RetryLoop;
 import org.apache.curator.TimeTrace;
+import org.apache.curator.framework.AuthInfo;
 import org.apache.curator.framework.CuratorFramework;
 import org.apache.curator.framework.CuratorFrameworkFactory;
 import org.apache.curator.framework.api.*;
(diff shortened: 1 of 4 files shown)


curator-framework/src/main/java/org/apache/curator/framework/imps/CuratorFrameworkImpl.java
```diff
@@ -65,7 +69,7 @@
     private final BlockingQueue<OperationAndData<?>>                    backgroundOperations;
     private final NamespaceImpl                                         namespace;
     private final ConnectionStateManager                                connectionStateManager;
-    private final AtomicReference<AuthInfo>                             authInfo = new AtomicReference<AuthInfo>();
```
kambatla: I didn't see any thread-safety issues using List<AuthInfo>

cammckenzie: No, I don't see any issues with it either. Just wondering why it was previously an AtomicReference. Anyway, I think this looks good to merge. Thanks.

Randgalt: Hmm - I don't know why it's an AtomicReference. It doesn't seem necessary.

madrob: It was probably an AtomicRef so that it could be marked final and referred to from inside of the anonymous inner class in makeZookeeperFactory.

Randgalt: As I've gotten older my mind has less and less room for information. I forget everything these days. I don't even remember if I wrote this code or someone else did. :(

PR #60, opened 2015-01-02
Use Curator in thread names.

swankjesse:


swankjesse:
https://issues.apache.org/jira/browse/CURATOR-177

madrob:
LGTM!

Randgalt:
It would be a lot simpler to make the change in ThreadUtils, right?

curator-framework/src/main/java/org/apache/curator/framework/imps/CuratorTempFrameworkImpl.java
```diff
@@ -98,7 +99,15 @@ private synchronized void openConnectionIfNeeded() throws Exception
         if ( cleanup == null )
         {
             ThreadFactory threadFactory = factory.getThreadFactory();
-            cleanup = (threadFactory != null) ? Executors.newScheduledThreadPool(1, threadFactory) : Executors.newScheduledThreadPool(1);
+
```
dragonsinth: Is there a ThreadUtils call that could replace?

dragonsinth:
+1 to @Randgalt suggestion

curator-framework/src/main/java/org/apache/curator/framework/imps/CuratorTempFrameworkImpl.java
dragonsinth: Maybe we want a ThreadUtils.newNonDaemonThreadFactory() to replace this?
But LGTM either way.

dragonsinth: But @Randgalt can you verify that this should be non-daemon?  Just wondering if it's a latent bug here that this will tie up process exit.

PR #62, opened 2015-01-13
CURATOR-111 - Modified the authorization handling so that the underlying

cammckenzie:
auth data is stored in a single list. Modified tests to have both a single
auth and multiple auth case.

### DIFF
--- a/curator-framework/src/main/java/org/apache/curator/framework/CuratorFrameworkFactory.java
+++ b/curator-framework/src/main/java/org/apache/curator/framework/CuratorFrameworkFactory.java
@@ -406,20 +406,20 @@ public class CuratorFrameworkFactory
         {
             switch ( authInfos.size() )
             {
-            case 0:
-            {
-                return null;
-            }
+                case 0:
+                {
+                    return null;
+                }
 
-            case 1:
-            {
-                return authInfos.get(0).getAuth();
-            }
+                case 1:
+                {
+                    return authInfos.get(0).getAuth();
+                }
 
-            default:
-            {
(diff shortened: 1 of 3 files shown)


curator-framework/src/main/java/org/apache/curator/framework/CuratorFrameworkFactory.java
```diff
@@ -380,16 +383,6 @@ public String getNamespace()
             return namespace;
         }
 
-        public String getAuthScheme()
```
Randgalt: I think these methods should stay and be marked as deprecated so this is not a breaking change. The old methods can either return return the first auth or throw an error if there is more than 1 auth.

curator-framework/src/main/java/org/apache/curator/framework/CuratorFrameworkFactory.java
```diff
@@ -163,8 +166,8 @@ public CuratorTempFramework buildTemp(long inactiveThreshold, TimeUnit unit)
          */
         public Builder authorization(String scheme, byte[] auth)
         {
-            this.authScheme = scheme;
-            this.authValue = (auth != null) ? Arrays.copyOf(auth, auth.length) : null;
+            this.authInfos = Lists.newArrayList();
```
Randgalt: This should call authorization(List<AuthInfo> authInfos). E.g.

```
return authorization(Lists.newArrayList(new AuthInfo(scheme, (auth != null) ? Arrays.copyOf(auth, auth.length) : null)));
```

PR #64, opened 2015-01-23
Port Codehaus Jackson to fasterxml Jackson.

ae6rt:
We believed there might be value in providing a port of Curator from legacy Codehaus Jackson to FasterXML Jackson.  This PR is that port.

One line of executable code changed, here, in

org.apache.curator.x.discovery.server.contexts.GenericDiscoveryContext#unMarshallJson

	@Override
    public T unMarshallJson(JsonNode node) throws Exception
    {
        T payload;
        ObjectMapper mapper = new ObjectMapper();
        //noinspection unchecked
        payload = (T)mapper.readValue(mapper.treeAsTokens(node),  // <<<<<<<<<<<<<<<<
        payloadType.getRawType());
        return payload;
    }

Jackson v2 has no method on ObjectMapper supporting readValue(JsonNode, Class).  However, Jackson v1 states the form I used
is a shortcut for that legacy method.

All other changes to the code and poms were simply replacing imports
and maven dependencies.

Jackson 2.3.2 was chosen to be compatible with the version already
being transitively provided through the dependency
io.dropwizard:dropwizard-configuration:jar:0.7.0:compile

### DIFF
--- a/curator-examples/src/main/java/discovery/InstanceDetails.java
+++ b/curator-examples/src/main/java/discovery/InstanceDetails.java
@@ -18,7 +18,7 @@
  */
 package discovery;
 
-import org.codehaus.jackson.map.annotate.JsonRootName;
+import com.fasterxml.jackson.annotation.JsonRootName;
 
 /**
  * In a real application, the Service payload will most likely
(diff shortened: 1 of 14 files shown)


ae6rt:
I neglected to file a JIRA issue and name the branch accordingly.  Closing in favor of that process.

PR #65, opened 2015-01-23
Port Codehaus Jackson to fasterxml Jackson.

ae6rt:
CURATOR-186

One line of executable code changed, here, in

org.apache.curator.x.discovery.server.contexts.GenericDiscoveryContext#unMarshallJson

	@Override
    public T unMarshallJson(JsonNode node) throws Exception
    {
        T payload;
        ObjectMapper mapper = new ObjectMapper();
        //noinspection unchecked
        payload = (T)mapper.readValue(mapper.treeAsTokens(node),  // <<<<<<<<<<<<<<<<
        payloadType.getRawType());
        return payload;
    }

Jackson v2 has no method on ObjectMapper supporting readValue(JsonNode, Class).  However, Jackson v1 states the form I used
is a shortcut for that legacy method.

All other changes to the code and poms were simply replacing imports
and maven dependencies.

Jackson 2.3.2 was chosen to be compatible with the version already
being transitively provided through the dependency
io.dropwizard:dropwizard-configuration:jar:0.7.0:compile

### DIFF
--- a/curator-examples/src/main/java/discovery/InstanceDetails.java
+++ b/curator-examples/src/main/java/discovery/InstanceDetails.java
@@ -18,7 +18,7 @@
  */
 package discovery;
 
-import org.codehaus.jackson.map.annotate.JsonRootName;
+import com.fasterxml.jackson.annotation.JsonRootName;
 
 /**
  * In a real application, the Service payload will most likely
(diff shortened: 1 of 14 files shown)


Randgalt:
This looks good to me. This won't affect existing implementations right?

madrob:
It will not be compatible with existing clients, no. The method `public T unMarshallJson(JsonNode node) throws Exception` goes from taking a `org.codehaus.jackson.JsonNode` to taking a `com.fasterxml.jackson.databind.JsonNode`.

Well, I guess it would be fine if that method is only used internally, but I'm not confident enough in that assertion.

PR #66, opened 2015-02-09
Curator 187

dkesler:
A fix for the problem described in https://issues.apache.org/jira/browse/CURATOR-187.  The ChildReaper continuously runs even when its Reaper is not the leader, resulting in a pileup of paths to reap in the Reaper's queue. 

This changeset allows the ChildReaper to create the LeaderLatch so that it too can avoid doing any work (i.e. adding paths to the Reaper's queue) if this application is not the leader.

### DIFF
--- a/curator-recipes/src/main/java/org/apache/curator/framework/recipes/locks/ChildReaper.java
+++ b/curator-recipes/src/main/java/org/apache/curator/framework/recipes/locks/ChildReaper.java
@@ -20,6 +20,8 @@ package org.apache.curator.framework.recipes.locks;
 
 import com.google.common.base.Preconditions;
 import com.google.common.collect.Sets;
+
+import org.apache.curator.framework.recipes.leader.LeaderLatch;
 import org.apache.curator.utils.CloseableUtils;
 import org.apache.curator.framework.CuratorFramework;
 import org.apache.curator.utils.CloseableScheduledExecutorService;
@@ -52,6 +54,7 @@ public class ChildReaper implements Closeable
     private final Reaper.Mode mode;
     private final CloseableScheduledExecutorService executor;
     private final int reapingThresholdMs;
+    private final LeaderLatch leaderLatch;
 
     private volatile Future<?> task;
 
@@ -109,7 +112,15 @@ public class ChildReaper implements Closeable
         this.mode = mode;
         this.executor = new CloseableScheduledExecutorService(executor);
         this.reapingThresholdMs = reapingThresholdMs;
-        this.reaper = new Reaper(client, executor, reapingThresholdMs, leaderPath);
+        if (leaderPath != null)
+        {
+            leaderLatch = new LeaderLatch(client, leaderPath);
+        }
+        else
+        {
+            leaderLatch = null;
+        }
+        this.reaper = new Reaper(client, executor, reapingThresholdMs, leaderLatch);
         addPath(path);
     }
 
@@ -136,7 +147,10 @@ public class ChildReaper implements Closeable
             reapingThresholdMs,
             TimeUnit.MILLISECONDS
         );
-
+        if (leaderLatch != null)
+        {
+            leaderLatch.start();
+        }
         reaper.start();
     }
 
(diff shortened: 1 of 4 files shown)


curator-recipes/src/main/java/org/apache/curator/framework/recipes/locks/ChildReaper.java
```diff
@@ -173,32 +191,40 @@ public boolean removePath(String path)
...
-    private static ScheduledExecutorService newExecutorService()
+    public static ScheduledExecutorService newExecutorService()
     {
         return ThreadUtils.newFixedThreadScheduledPool(2, "ChildReaper");
     }
 
     private void doWork()
     {
-        for ( String path : paths )
+        if (shouldDoWork())
```
cammckenzie: There is potential for a race condition here if leadership is lost while attempting to do the reaping. Not sure if this is an issue or not?

I would assume that it's not really an issue as:
1.) If leadership has been lost then the connection to ZK will be lost so the reaping won't work anyway.
2.) If the reaping fails due to 2 nodes concurrently attempting to delete / query children, then it will just run again next time anyway.

curator-recipes/src/main/java/org/apache/curator/framework/recipes/locks/Reaper.java
```diff
@@ -333,7 +355,20 @@ public void notLeader()
                 reapingIsActive.set(false);
             }
         };
-        localLeaderLatch.addListener(listener);
-        return localLeaderLatch;
+        leaderLatch.addListener(listener);
+
+        reapingIsActive.set(leaderLatch.hasLeadership());
```
cammckenzie: Another potential race condition here? If the Reaper gets passed a LeaderLatch in its constructor that has already undergone election (and been elected leader) then the isLeader() method on the listener will not fire until another election occurs. While the reapiningIsActive flag will be set to true, the schedule() method will not get called so no reaping will actually occur?

curator-recipes/src/main/java/org/apache/curator/framework/recipes/locks/Reaper.java
```diff
@@ -333,7 +355,20 @@ public void notLeader()
...
+    }
+
+    private static LeaderLatch makeLeaderLatchIfPathNotNull(CuratorFramework client, String leaderPath)
```
cammckenzie: Any reason for this to be static?

dkesler: It's because it's called from within a constructor that calls another constructor.

cammckenzie: Yep, good point, didn't look at where it was being called from.

curator-recipes/src/main/java/org/apache/curator/framework/recipes/locks/Reaper.java
dkesler: schedule() gets called on the individual paths in activePaths.  Since this method only gets called during the construction of the Reaper, there won't be any activePaths to schedule. When those paths get added later, they will get scheduled as expected.

cammckenzie: Ok.

curator-recipes/src/main/java/org/apache/curator/framework/recipes/locks/ChildReaper.java
dkesler: That race condition is already present with the leader election in the Reaper itself as it could lose leadership after scheduling a check for some time in the future, so this is at least no worse than the current state of things.  And the race condition doesn't matter too much anyway as the reapers don't use leader election for correctness, but rather because otherwise you get a bunch of logspam about trying to delete already deleted nodes.

cammckenzie: Yep, thought as much.

cammckenzie:
Looks good to me. Thanks for the fix! I will merge shortly.

PR #67, opened 2015-02-10
CURATOR-173

dkesler:
Adding the notion of a 'lock schema' to ChildReaper that enables it to reap both the direct children its watching and subnodes of those children.  This is necessary with InterProcessSemaphoreV2 as it creates multiple subnodes beneath its lock nodes and otherwise is unreapable with ChildReaper

I'm a little unsure of this solution.  It definitely doesn't make sense for ChildReaper to know about the locks and leases paths for InterProcessSemaphoreV2.  This solution allows a given lock implementation to provide its schema which the ChildReaper can use to clean up for that particular lock.  Currently IPSV2 is the only lock that needs to do so, however.  While more generic than childreaper knowing about lock implementations directly, this still seems a little bruteforcey.  It also wouldn't handle the case where the schema of a lock layout isn't known statically (although they all are currently).    

Another approach might have been to enable childreaper to work recursively, but then it needs to know to not reap the ephemeral nodes created by the lock implementation itself.  (Perhaps simply excluding ephemeral nodes  would work, but it kind of seems like that just "happens to make it work right" rather than because it makes sense for the general concept of child reaping to ignore ephemeral nodes).


dkesler:
Just wondering if anyone has had a chance to take a look at this yet or had any feedback.

Randgalt:
I haven't. You might start a discussion on @user

cammckenzie:
I've had a brief look at this, and I think it looks ok.
A quick thought about allowing for dynamic change to the schema, is it better to get the lock implementation instance to return a reference to an instance of the LockSchema object that is specific to that particular lock instance? The lock implementation could then update this reference with any dynamic content?

Obviously  this would add the need for synchronisation etc.

dkesler:
@cammckenzie The problem with requiring passing the lock instance to the reaper to get the lock schema is that you can end up leaking locks.  If the app that took the lock doesn't successfully clean it up then other processes won't know the schema of that particular lock instance since they need an instance of the lock in order to clean it up.

Not entirely sure if it's a huge concern, but I thought I'd point it out since I don't have insight as to possible future lock incarnations.

cammckenzie:
@dkesler That's a good point, and I guess is going to be an inherent flaw with any implementation that does not persist the schema somewhere. I guess that persisting the lock schema to ZK would be a possibility but it seems a bit over engineered. Perhaps your static approach is the best for now, and we can retrofit something else down the track if we have new locking implementations that have dynamic schemas.

I'll hopefully get another chance to look at your patch in a bit more detail today.

dkesler:
Yeah, if there really is a need for locks with dynamic schemas maybe the idea of a recursive childreaper that reaps any childless non-ephemeral node might be the way to go, although that would also impose certain restrictions on how locks behave.

curator-recipes/src/main/java/org/apache/curator/framework/recipes/locks/LockSchema.java
```diff
@@ -0,0 +1,22 @@
...
+import java.util.HashSet;
+import java.util.Set;
+
+import com.google.common.collect.Sets;
+
+public class LockSchema {
+    private final Set<String> paths;
+
+    public LockSchema() {
+        paths = new HashSet<String>();
```
cammckenzie: Just nit picking but this should probably be Sets.newHashSet() for consistency.

cammckenzie:
@dkesler Thanks for the patch, I think this is good to merge.
@Randgalt, @dragonsinth or @madrob I think that this fix is good, but I haven't used the semaphore stuff much, so I think a second opinion would be good prior to merging.

curator-recipes/src/main/java/org/apache/curator/framework/recipes/locks/LockSchema.java
```diff
@@ -0,0 +1,22 @@
...
+    }
+
+    public LockSchema(Set<String> paths) {
+        this.paths = Sets.newHashSet(paths);
+    }
+
+    public Set<String> getPaths() {
+        return Sets.newHashSet(paths);
+    }
+}
```
dragonsinth: This doesn't look like it adds a lot of value as a concrete class.  I would either just pass the set directly to the reaper, or else turn this into an interface.  You could also use Iterable instead of Set since you don't really depend on Set semantics.

dragonsinth:
LGTM with nit

dkesler:
@dragonsinth Thanks for the feedback.  I've pushed a change removing the LockSchema object and instead just using Set<String> directly.

dragonsinth:
LGTM

Randgalt:
@dragonsinth please merge it in if it's good

dragonsinth:
Will merge, sorry for the delay

PR #68, opened 2015-03-03
Make PathChildrenCache ignore old versions

cheddar:
Currently, PathChildrenCache, when set to cache data, can potentially cache stale versions of files.  In some client code, we are seeing CHILD_UPDATE events being delivered with out-of-order versions.

This PR is for a fix to the logic for what to do with old `Stat` objects.  That said, I looked over the unit tests and wasn't able to come up with a great strategy for verifying this behavior.  It would be nice to be able to mock something out so that I just feed out-of-order events to the method in question, but I'm not seeing that injection point.  Any ideas on how this could be easily UT'd?

### DIFF
--- a/curator-recipes/src/main/java/org/apache/curator/framework/recipes/cache/PathChildrenCache.java
+++ b/curator-recipes/src/main/java/org/apache/curator/framework/recipes/cache/PathChildrenCache.java
@@ -675,14 +675,27 @@ public class PathChildrenCache implements Closeable
         if ( resultCode == KeeperException.Code.OK.intValue() ) // otherwise - node must have dropped or something - we should be getting another event
         {
             ChildData data = new ChildData(fullPath, stat, bytes);
-            ChildData previousData = currentData.put(fullPath, data);
-            if ( previousData == null ) // i.e. new
-            {
-                offerOperation(new EventOperation(this, new PathChildrenCacheEvent(PathChildrenCacheEvent.Type.CHILD_ADDED, data)));
-            }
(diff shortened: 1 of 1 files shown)


drcrallen:
Unit testing for when ZK decides to be silly is a constant thorn in our sides.

curator-recipes/src/main/java/org/apache/curator/framework/recipes/cache/PathChildrenCache.java
```diff
@@ -675,14 +675,27 @@ private void applyNewData(String fullPath, int resultCode, Stat stat, byte[] byt
...
+                ChildData previousData = currentData.get(fullPath);
+                if ( previousData == null ) // i.e. new
+                {
+                    if (currentData.putIfAbsent(fullPath, data) == null) // is currentData still missing the key?
+                    {
+                        offerOperation(new EventOperation(this, new PathChildrenCacheEvent(PathChildrenCacheEvent.Type.CHILD_ADDED, data)));
+                        break;
+                    }
+                }
+                else if ( previousData.getStat().getVersion() < stat.getVersion() ) // Newer version of something that existed
```
Randgalt: This is the real change, right? The other code is just reformatting, etc.?

Randgalt:
Also, what is the referring Jira issue for this?

tisonkun:
Closed as no more activities. @cheddar if you'd like to proceed this patch, please resubmit with an associated ticket and addressing comments above.

PR #69, opened 2015-03-03
CURATOR-190 - Modified to always use the base name when creating protect...

cammckenzie:
Modified to always use the base name when creating protected ephemeral nodes. This prevents the current protected name being appended to the name each time the node is created, which in turn prevents multiple
nodes being created.


Randgalt:
LGTM

PR #72, opened 2015-04-10
CURATOR-164 prevent phantom services

madrob:
When ServiceDiscovery reregisters services after a reconnection, it's
possible to add back a service that was already deleted. This is why
we need to remove the service from the cache before deleting it from
zookeeper.

### DIFF
--- a/curator-x-discovery/src/main/java/org/apache/curator/x/discovery/details/ServiceDiscoveryImpl.java
+++ b/curator-x-discovery/src/main/java/org/apache/curator/x/discovery/details/ServiceDiscoveryImpl.java
@@ -207,9 +207,15 @@ public class ServiceDiscoveryImpl<T> implements ServiceDiscovery<T>
         {
             try
             {
-                CreateMode      mode = (service.getServiceType() == ServiceType.DYNAMIC) ? CreateMode.EPHEMERAL : CreateMode.PERSISTENT;
-                client.create().creatingParentsIfNeeded().withMode(mode).forPath(path, bytes);
-                isDone = true;
+                synchronized(service) {
+                    // Check that the service didn't get deleted while we were reconnecting
+                    if (services.containsKey(service.getId()))
+                    {
+                        CreateMode      mode = (service.getServiceType() == ServiceType.DYNAMIC) ? CreateMode.EPHEMERAL : CreateMode.PERSISTENT;
+                        client.create().creatingParentsIfNeeded().withMode(mode).forPath(path, bytes);
+                        isDone = true;
+                    }
+                }
             }
             catch ( KeeperException.NodeExistsException e )
             {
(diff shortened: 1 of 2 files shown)


curator-x-discovery/src/test/java/org/apache/curator/x/discovery/TestServiceDiscovery.java
```diff
@@ -297,4 +304,70 @@ public void testNoServerOnStart() throws Exception
...
         }
     }
+
+    // CURATOR-164
+    @Test
+    public void testUnregisterService() throws Exception {
+        final String name = "name";
+
+        final CountDownLatch restartLatch = new CountDownLatch(1);
+        final CountDownLatch serializeLatch = new CountDownLatch(1);
```
busbey: you made this latch, but then used a sleep call to wait for the serialization to finish. better ot switch to using this latch.

curator-x-discovery/src/main/java/org/apache/curator/x/discovery/details/ServiceDiscoveryImpl.java
```diff
@@ -207,9 +207,15 @@ protected void     internalRegisterService(ServiceInstance<T> service) throws Ex
         {
             try
             {
-                CreateMode      mode = (service.getServiceType() == ServiceType.DYNAMIC) ? CreateMode.EPHEMERAL : CreateMode.PERSISTENT;
-                client.create().creatingParentsIfNeeded().withMode(mode).forPath(path, bytes);
-                isDone = true;
+                synchronized(service) {
+                    // Check that the service didn't get deleted while we were reconnecting
+                    if (services.containsKey(service.getId()))
```
Randgalt: Why is this needed? The test passes without this and logically it seems unnecessary.

curator-x-discovery/src/main/java/org/apache/curator/x/discovery/details/ServiceDiscoveryImpl.java
```diff
@@ -228,15 +234,18 @@ protected void     internalRegisterService(ServiceInstance<T> service) throws Ex
...
     {
         String          path = pathForInstance(service.getName(), service.getId());
-        try
-        {
-            client.delete().forPath(path);
-        }
-        catch ( KeeperException.NoNodeException ignore )
-        {
-            // ignore
+        synchronized(service) {
```
Randgalt: I don't think the sync adds anything. The services are usually copies.

madrob: Imagine the following sequence:

T1: reconnect()
T2: unregister(X)
T1: for each service in cache:
T1:   register(X)
T2: services.remove(X)
T2: client.delete(X)
T1: client.create(X)

Randgalt:
I have an alternate fix that avoids syncs. Please review: https://github.com/apache/curator/pull/73

PR #74, opened 2015-04-28
LeaderSelector mutex resurrection

wrmsr:
I am not alone in experiencing situations in which my LeaderSelectors will wind up in an indefinite state of having no leader (relevant JIRA issues are listed below). This problem had been occurring nearly daily in our AWS environments (which regularly experience transient network issues). I believe I have solved this issue. This may not be the most elegant approach but, in the least, our deployments have behaved correctly since its activation.

This branch allows an InterProcessMutex to optionally reuse an existing acquisition. This of course breaks the contract of re-entrance as stated by the InterProcessLock interface but it is not done by default and only used specifically by the LeaderSelector (which is the only thing I am interested in using it for). I have a test reliably (though hackily) reproducing this issue but it is written in terms of an internal project and as I am unfamiliar with your test code I haven't ported it yet. All existing tests pass. The term resurrect probably isn't the best but hey, it works :p

Issues possibly fixed by this branch:
https://issues.apache.org/jira/browse/CURATOR-3
https://issues.apache.org/jira/browse/CURATOR-171
https://issues.apache.org/jira/browse/CURATOR-188
https://issues.apache.org/jira/browse/CURATOR-202
https://issues.apache.org/jira/browse/CURATOR-205

### DIFF
--- a/curator-recipes/src/main/java/org/apache/curator/framework/recipes/leader/LeaderSelector.java
+++ b/curator-recipes/src/main/java/org/apache/curator/framework/recipes/leader/LeaderSelector.java
@@ -383,7 +383,7 @@ public class LeaderSelector implements Closeable
         hasLeadership = false;
         try
         {
-            mutex.acquire();
+            mutex.acquire(true);
 
             hasLeadership = true;
             try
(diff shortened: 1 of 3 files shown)


jolynch:
Any update on this? It seems like it might fix a lot of really critical bugs?

Randgalt:
We believe that Curator 2.8.0 fixes the problems with Leader Selector. Have you tested with it?

Randgalt:
Have you tried Curator 2.8.0?

tisonkun:

PR #77, opened 2015-05-09
[CURATOR-215] Support async multi/transaction APIs

Randgalt:
I took this opportunity to re-write the transaction APIs. The old ones were far too clever and hard to use (though pretty!). At the same time, added support for ZK 3.5's async multi() APIs. Old Curator transaction APIs have been deprecated. This uses the old transactions internal code for consistency. However, this makes the implementation a bit cumbersome. In the future, this can all be cleaned up

### DIFF
--- a/curator-framework/src/main/java/org/apache/curator/framework/api/transaction/CuratorMultiTransactionMain.java
+++ b/curator-framework/src/main/java/org/apache/curator/framework/api/transaction/CuratorMultiTransactionMain.java
@@ -1,0 +1,30 @@
+/**
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package org.apache.curator.framework.api.transaction;
+
+import org.apache.zookeeper.Op;
+import org.apache.zookeeper.OpResult;
+import java.util.List;
+
+public interface CuratorMultiTransactionMain
+{
+    List<OpResult> forOperations(Op... operations) throws Exception;
+
+    List<OpResult> forOperations(List<Op> operations) throws Exception;
+}
--- a/curator-framework/src/main/java/org/apache/curator/framework/api/transaction/CuratorMultiTransactionMain.java
+++ b/curator-framework/src/main/java/org/apache/curator/framework/api/transaction/CuratorMultiTransactionMain.java
@@ -28,7 +28,7 @@ public interface CuratorMultiTransactionMain
      * operation instances via {@link CuratorFramework#transactionOp()}
      *
      * @param operations operations that make up the transaction.
-     * @return result details
+     * @return result details for foreground operations or <code>null</code> for background operations
      * @throws Exception errors
      */
     List<CuratorTransactionResult> forOperations(CuratorOp... operations) throws Exception;
@@ -38,7 +38,7 @@ public interface CuratorMultiTransactionMain
      * operation instances via {@link CuratorFramework#transactionOp()}
      *
      * @param operations operations that make up the transaction.
-     * @return result details
+     * @return result details for foreground operations or <code>null</code> for background operations
      * @throws Exception errors
      */
     List<CuratorTransactionResult> forOperations(List<CuratorOp> operations) throws Exception;
--- a/curator-framework/src/main/java/org/apache/curator/framework/api/transaction/CuratorMultiTransactionMain.java
+++ b/curator-framework/src/main/java/org/apache/curator/framework/api/transaction/CuratorMultiTransactionMain.java
@@ -18,11 +18,28 @@
  */
 package org.apache.curator.framework.api.transaction;
 
+import org.apache.curator.framework.CuratorFramework;
 import java.util.List;
 
 public interface CuratorMultiTransactionMain
 {
+    /**
+     * Commit the given operations as a single transaction. Create the
+     * operation instances via {@link CuratorFramework#transactionOp()}
+     *
+     * @param operations operations that make up the transaction.
+     * @return result details
+     * @throws Exception errors
+     */
     List<CuratorTransactionResult> forOperations(CuratorOp... operations) throws Exception;
 
+    /**
+     * Commit the given operations as a single transaction. Create the
+     * operation instances via {@link CuratorFramework#transactionOp()}
+     *
+     * @param operations operations that make up the transaction.
+     * @return result details
+     * @throws Exception errors
+     */
     List<CuratorTransactionResult> forOperations(List<CuratorOp> operations) throws Exception;
 }
--- a/curator-framework/src/main/java/org/apache/curator/framework/api/transaction/CuratorMultiTransactionMain.java
+++ b/curator-framework/src/main/java/org/apache/curator/framework/api/transaction/CuratorMultiTransactionMain.java
@@ -18,13 +18,11 @@
  */
 package org.apache.curator.framework.api.transaction;
 
-import org.apache.zookeeper.Op;
-import org.apache.zookeeper.OpResult;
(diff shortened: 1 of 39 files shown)


curator-framework/src/main/java/org/apache/curator/framework/api/transaction/CuratorMultiTransactionMain.java
```diff
@@ -0,0 +1,45 @@
...
+
+    /**
+     * Commit the given operations as a single transaction. Create the
+     * operation instances via {@link CuratorFramework#transactionOp()}
+     *
+     * @param operations operations that make up the transaction.
+     * @return result details
+     * @throws Exception errors
+     */
+    List<CuratorTransactionResult> forOperations(List<CuratorOp> operations) throws Exception;
```
madrob: Be more flexible, accept an Iterable or Collection instead of a List, maybe?

Randgalt: I think order matters here. Thus, the list.

curator-framework/src/main/java/org/apache/curator/framework/imps/CuratorMultiTransactionImpl.java
```diff
@@ -0,0 +1,166 @@
...
+import org.apache.curator.framework.api.transaction.CuratorTransactionResult;
+import org.apache.zookeeper.AsyncCallback;
+import org.apache.zookeeper.OpResult;
+import java.util.Arrays;
+import java.util.List;
+import java.util.concurrent.Callable;
+import java.util.concurrent.Executor;
+
+/**
+ * @deprecated Use {@link CuratorFramework#transaction()}
```
madrob: Why are we adding a class that is immediately deprecated? Or is this a result of a move that isn't clear on the web view of the PR?

curator-framework/src/main/java/org/apache/curator/framework/api/transaction/CuratorMultiTransactionMain.java
madrob: If users believe that order matters, then they can use a list. If they don't care, then it might make sense for them to use a Set, or in Java8 a Stream. Well, a finite Stream. Also, would be more flexible to use `<? extends CuratorOp>`

curator-framework/src/main/java/org/apache/curator/framework/imps/CuratorTransactionImpl.java
```diff
@@ -188,19 +169,45 @@ private CuratorTransactionResult makeCuratorResult(OpResult opResult, CuratorMul
...
-        return new CuratorTransactionResult(metadata.type, metadata.forPath, resultPath, resultStat);
+        return new CuratorTransactionResult(metadata.getType(), metadata.getForPath(), resultPath, resultStat);
+    }
+
+    private List<OpResult> doOperation(AtomicBoolean firstTime) throws Exception
+    {
+        boolean localFirstTime = firstTime.getAndSet(false);
+        if ( !localFirstTime )
+        {
+            // TODO
```
madrob: Do what?

curator-framework/src/main/java/org/apache/curator/framework/imps/ExtractingCuratorOp.java
```diff
@@ -0,0 +1,59 @@
...
+class ExtractingCuratorOp implements CuratorOp
+{
+    private final CuratorMultiTransactionRecord record = new CuratorMultiTransactionRecord();
+
+    CuratorMultiTransactionRecord getRecord()
+    {
+        return record;
+    }
+
+    CuratorOp asCuratorOp()
```
madrob: This should not be necessary.

curator-framework/src/main/java/org/apache/curator/framework/imps/SetDataBuilderImpl.java
```diff
@@ -246,7 +243,7 @@ public Stat forPath(String path, byte[] data) throws Exception
         Stat        resultStat = null;
         if ( backgrounding.inBackground()  )
         {
-            client.processBackgroundOperation(new OperationAndData<PathAndBytes>(this, new PathAndBytes(path, data), backgrounding.getCallback(), null, backgrounding.getContext()), null);
+            client.processBackgroundOperation(new OperationAndData<>(this, new PathAndBytes(path, data), backgrounding.getCallback(), null, backgrounding.getContext()), null);
```
madrob: This is not a valid JDK6 construct.

curator-framework/src/site/confluence/index.confluence
```diff
@@ -42,11 +42,16 @@ h3. Methods
 |getData()|Begins an operation to get a ZNode's data. Call additional methods (watch, background or get stat) and finalize the operation by calling forPath()|
 |setData()|Begins an operation to set a ZNode's data. Call additional methods (version or background) and finalize the operation by calling forPath()|
 |getChildren()|Begins an operation to get a ZNode's list of children ZNodes. Call additional methods (watch, background or get stat) and finalize the operation by calling forPath()|
+<<<<<<< HEAD
```
madrob: Merge conflict needs to be resolved.

curator-framework/src/main/java/org/apache/curator/framework/api/transaction/CuratorMultiTransactionMain.java
Randgalt: I realize now that the results are returned in the order of the operations. It's the only way to correlate. So, that suggests leaving it as a list.

curator-framework/src/main/java/org/apache/curator/framework/imps/CuratorMultiTransactionImpl.java
Randgalt: The deprecated is a mistake here. I'll remove it.

curator-framework/src/main/java/org/apache/curator/framework/imps/CuratorTransactionImpl.java
Randgalt: That's been in there a long time. I should just remove it.

curator-framework/src/main/java/org/apache/curator/framework/imps/SetDataBuilderImpl.java
Randgalt: Curator 3.0 will be Java 7 as ZK 3.5 is.

curator-framework/src/main/java/org/apache/curator/framework/imps/ExtractingCuratorOp.java
Randgalt: It avoids a cast in some code (see TransactionOpImpl)

madrob: Two notes - in Java 8, I believe the compiler can do that inferencing for you so no cast required, and today I think you can add a type witness: `new CreateBuilderImpl(client).<CuratorOp>asTransactionCreateBuilder(op, op.getRecord())`

Randgalt: Yeah, in Java 8 this is relaxed. I'll look at the second thing.

curator-framework/src/main/java/org/apache/curator/framework/api/transaction/CuratorMultiTransactionMain.java
```diff
@@ -0,0 +1,45 @@
...
+     * @throws Exception errors
+     */
+    List<CuratorTransactionResult> forOperations(CuratorOp... operations) throws Exception;
+
+    /**
+     * Commit the given operations as a single transaction. Create the
+     * operation instances via {@link CuratorFramework#transactionOp()}
+     *
+     * @param operations operations that make up the transaction.
+     * @return result details
```
madrob: Expand the javadoc here, please. For example, I was surprised to read the code and discover that this call returns null if the operation is in the background.

madrob:
Can you add a class to curator-examples that demonstrates how to use this new API?

Randgalt:
That's a good idea - will do

madrob:
TestCompressionInTransaction uses the old APIs. A test with the new APIs would be good to check for regressions.

Also, I think something weird happened during the merge in CreateBuilderImpl. It's missing `debugForceFindProtectedNode` which is causing test failures.

madrob:
Another thing that is unclear is whether the `TransactionOp` are reusable or if they are one-shot.

Randgalt:
Anything else before I merge?

madrob:
I don't think you addressed the floating TODO, and `forOperation` should take `List<? extends CuratorOp>` instead of `List<CuratorOp>`. Other than those minor issues, the rest of this looks good. Thanks, Jordan!

Randgalt:
I'm going to leave the TODO for future investigation. I want to know why it's there. Thanks for the great review.

PR #78, opened 2015-05-12
CURATOR-219 Catch Exceptions, not Throwables

madrob:


curator-client/src/main/java/org/apache/curator/ensemble/exhibitor/ExhibitorEnsembleProvider.java
```diff
@@ -301,7 +301,7 @@ else if ( port != thePort )
                     values.putAll(decodeExhibitorList(encoded));
                     done = true;
                 }
-                catch ( Throwable e )
+                catch ( Exception e )
```
Randgalt: I don't understand how this is an improvement. Exceptions such as NPE or OOM would be lost with this change. By catching Throwable, those exceptions are logged.

curator-framework/src/main/java/org/apache/curator/framework/imps/CuratorFrameworkImpl.java
```diff
@@ -812,31 +812,28 @@ private void performBackgroundOperation(OperationAndData<?> operationAndData)
                 queueOperation(operationAndData);
             }
         }
-        catch ( Throwable e )
+        catch ( CuratorConnectionLossException e )
```
Randgalt: Similar to above. If Throwable is not caught, the exception will be lost. By catching Throwable, the exception will get logged.

curator-client/src/main/java/org/apache/curator/ensemble/exhibitor/ExhibitorEnsembleProvider.java
madrob: NPE still gets caught. If OOM is thrown, does it really make sense to retry? That's usually a pretty drastic condition and very often the JVM is not actually in a recoverable state. The OOM can be thrown because of a different thread going wonky, and we have no way to clean up the resources it was using.

Randgalt: Probably not. But, I'd view that as a separate issue. It's technically up to the retry policy passed to ExhibitorEnsembleProvider.

SamG1000: You also need to \worry about ThreadDeath error:

https://docs.oracle.com/javase/7/docs/api/java/lang/ThreadDeath.html
```
An application should catch instances of this class only if it must clean up after being terminated asynchronously. If ThreadDeath is caught by a method, it is important that it be rethrown so that the thread actually dies.
```

So ThreadDeath Error can be logged, but should never retry to avoid destabilizing JVM.

tisonkun (changes requested):
-1

This patch includes various topics. You cannot simply "replace Throwable with Exceptions"; instead, you may make each change with a concrete reason.

You're welcome to improve error handling logics. But "replace Throwable with Exceptions" is not a reason.

PR #79, opened 2015-05-19
[CURATOR-222] Container support

Randgalt:
Assuming that ZOOKEEPER-2163 is accepted, container node support has been added to Curator. Also, all recipes that create parent nodes now create container nodes


madrob:
I think it would be helpful to users for `@deprecated` annotations to say "since 3.0.0"

Looks good so far, I'll give it a more thorough look tomorrow.

Randgalt:
Good point. In fact, we should use @since more broadly. But, I'll add this.

cammckenzie:
I've had a brief look and it looks OK to me, though the deprecated tags still don't have "since 3.0.0" from what I can see?

Randgalt:
Oh right. Those are added now.

Randgalt:
Any last comments before I merge?

Randgalt:
I ended up adding some more behavior and refining things. Please give a once-over if possible.

cammckenzie:
Will try and have a look tomorrow, real work is pretty flat out at the moment.

cammckenzie:
Unfortunately, I'm too flat out at the moment to look at this. Later next week I'll get a chance, but the real world is just too busy at the moment sorry.

curator-recipes/src/test/java/org/apache/curator/framework/recipes/cache/TestPathChildrenCache.java
```diff
@@ -100,16 +100,16 @@ public void childEvent(CuratorFramework client, PathChildrenCacheEvent event) th
         }
     }
 
-    @Test
+    @Test(enabled = false)  // I didn't write this test and it's not working. I'm not even sure what it tests. Check into it later -JZ
```
cammckenzie: What's the status of this? Looks like it was added as part of CURATOR-141, and I guess it's trying to get the client to shutdown during the build process. It looks pretty flaky though? Maybe we should just remove it

cammckenzie:
@Randgalt, looks good to me. A couple of minor things for discussion.
1.) Can the internal EnsurePath stuff in NamespaceImpl be replaced with the new container stuff?
2.) When executing creatingContainerParentsIfNeeded(), I was wondering if some sort of warning log should be generated if the parents already exist and are not container nodes? I'm just thinking of the case where you try and create a hierachy of nodes and some of them already exist as persistent nodes. Maybe it's just up to the user to remove any existing parents if this is important?

Randgalt:
> Can the internal EnsurePath stuff in NamespaceImpl be replaced with the new container stuff?

Possibly, but I'm deprecating EnsurePath so I don't see the need to add to it

> When executing creatingContainerParentsIfNeeded(), I was wondering if some sort of warning log should be generated if the parents already exist and are not container nodes? 

Unfortunately, there's no way to know if a node is a container or not right now. I had to shoehorn container support in with a hack. Maybe in a future version of ZK.

curator-recipes/src/test/java/org/apache/curator/framework/recipes/cache/TestPathChildrenCache.java
Randgalt: Yeah - I think it should be removed. I'll remove it.

cammckenzie:
My point with the NamespaceImpl stuff is that it's using EnsurePath (which is now deprecated) internally, Presumably if it's deprectated then it should be replaced by the container stuff.

Randgalt:
OK - I see the issue with NamespaceImpl. Fixed.

cammckenzie:
Ok, all looks good as far as I'm concerned. Nice work, this looks like really good feature.

Randgalt:
Thank you

PR #83, opened 2015-06-16
[CURATOR-223] Add executorService methods to ServiceCacheBuilder

tdyas:
Add .executorService variants to ServiceCacheBuilder to allow the caller to specify an ExecutorService or a CloseableExecutorService to be used by the PathChildrenCache embedded in ServiceCacheImpl.

Extracts ExecuteCalledWatchingExecutorService (and DelegatingExecutorService) into the curator-test module for use by TestServiceCache.

Relates to CURATOR-223

### DIFF
--- a/curator-recipes/src/test/java/org/apache/curator/framework/recipes/cache/TestPathChildrenCache.java
+++ b/curator-recipes/src/test/java/org/apache/curator/framework/recipes/cache/TestPathChildrenCache.java
@@ -31,6 +31,7 @@ import org.apache.curator.framework.api.UnhandledErrorListener;
 import org.apache.curator.framework.imps.CuratorFrameworkImpl;
 import org.apache.curator.retry.RetryOneTime;
 import org.apache.curator.test.BaseClassForTests;
+import org.apache.curator.test.ExecuteCalledWatchingExecutorService;
 import org.apache.curator.test.KillSession;
 import org.apache.curator.test.Timing;
 import org.apache.curator.utils.CloseableUtils;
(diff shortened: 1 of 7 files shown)


curator-recipes/src/main/java/org/apache/curator/framework/recipes/cache/PathChildrenCache.java
```diff
@@ -184,6 +184,28 @@ public PathChildrenCache(CuratorFramework client, String path, boolean cacheData
     }
 
     /**
+     * @param client        the client
```
Randgalt: Why are these new ctors need? PCC already has a bazillion

curator-x-discovery/src/main/java/org/apache/curator/x/discovery/details/ServiceCacheImpl.java
```diff
@@ -66,6 +68,30 @@
         cache.getListenable().addListener(this);
     }
 
+    ServiceCacheImpl(ServiceDiscoveryImpl<T> discovery, String name, ExecutorService executorService)
```
Randgalt: I'd prefer to see 1 ctor. It can take a CloseableExecutorService. The builder can wrap the executorService. I think there are other examples of this in the Curator code.

tdyas: Makes sense. Next iteration will have one ctor.

curator-recipes/src/main/java/org/apache/curator/framework/recipes/cache/PathChildrenCache.java
tdyas: I was trying to avoid using the ctor's with the `dataIsCompressed` argument since `ServiceCacheImpl` did not use those ctor's. I suppose it is fine to just invoke those ctor's and pass in `false` for `dataIsCompressed`.

tdyas:
I made the requested updates. The proposed commit no longer modifies `PathChildrenCache`, and `ServiceCacheImpl` has one additional ctor.

curator-x-discovery/src/main/java/org/apache/curator/x/discovery/details/ServiceCacheImpl.java
```diff
@@ -66,6 +68,18 @@
         cache.getListenable().addListener(this);
     }
 
+    ServiceCacheImpl(ServiceDiscoveryImpl<T> discovery, String name, CloseableExecutorService executorService)
```
Randgalt: Please have the other ctor call this one. You may need to introduce a third private ctor. i.e. only 1 ctor should have code. The others should forward to the 1 with code.

curator-recipes/src/test/java/org/apache/curator/framework/recipes/cache/TestPathChildrenCache.java
```diff
@@ -1039,127 +1040,4 @@ public void invoke() throws Exception
             CloseableUtils.closeQuietly(client);
         }
     }
-
-    public static class ExecuteCalledWatchingExecutorService extends DelegatingExecutorService
```
Randgalt: Why were these moved to separate files?

tdyas: To allow reuse in `TestServiceCache`.

Randgalt:
Please rename this PR so that the Jira Issue number is in the subject (see the other PRs)

curator-x-discovery/src/main/java/org/apache/curator/x/discovery/details/ServiceCacheImpl.java
tdyas: Will do. I had to add a conversion from `ThreadFactory` to `CloseableExecutorService` in a static method since Java doesn't permit ctor's to have code before the call to the common ctor.

PR #84, opened 2015-06-22
Curator-224: Fixed the requeuing problem with DistributedIdQueue.

zzcoder:
Made the one-line fix in DistributedQueue.java and added a test-case in TestDistributedIdQueue.java.


cammckenzie:
Looks good, thanks for the fix.

PR #85, opened 2015-06-24
CURATOR-225: Added new ErrorMode value KEEP.

zzcoder:
Added ErrorMode.KEEP. Created a new test-case testKeepMode() to make sure it works.

If the new mode is not used, the code should work as is so the impact of this change is very limited.

### DIFF
--- a/curator-recipes/src/main/java/org/apache/curator/framework/recipes/queue/DistributedQueue.java
+++ b/curator-recipes/src/main/java/org/apache/curator/framework/recipes/queue/DistributedQueue.java
@@ -650,7 +650,8 @@ public class DistributedQueue<T> implements QueueBase<T>
     private enum ProcessMessageBytesCode
     {
         NORMAL,
-        REQUEUE
+        REQUEUE,
+        KEEP
     }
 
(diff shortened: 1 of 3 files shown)


madrob:
Hi zzcoder! Thank you for your contribution and especially for including a test. I'm a little confused about how this behaviour is supposed to differ from `REQUEUE`. Can you please clarify the intent for me? Thanks!

zzcoder:
@madrob With REQUEUE, the entry is removed from the queue and appended to the back of the queue,  it violates the FIFO principal of the queue and it matters to our application. With the added option, the item will be kept in the queue intact.

Randgalt:
Maybe we should have REQUEUE attempt to put it back where it was instead of adding a new mode.

zzcoder:
@Randgalt I thought about that and it's problematic. The queue uses SEQUENTIAL flag. The sequence number determines the order in the queue. It seems that you have to tamper with ZooKeeper to achieve that.

Randgalt:
For the requeue, the sequence number has already been used so a simple create with the same sequence number would work.

Randgalt:
@zzcoder any updates on this?

zzcoder:
This PULL request is not needed anymore.

PR #87, opened 2015-07-07
CURATOR-235 LeaderSelector.internalRequeue should be private

bonifaido:
I think this was left public unintentionally.


madrob:
@bonifaido - This looks like a good catch! Can you create a JIRA issue for this at https://issues.apache.org/jira/browse/CURATOR and then reference that from here? That improves our ability to trace the work back later. Thanks!

bonifaido:
I have closed and reopened it to make it appear in JIRA.

madrob:
Looks like it's been accepted - can one of the committers add a commit with "closes #87" to close the loop on this?

tisonkun:
This is on the master https://github.com/apache/curator/commit/0bec8a066f8a4b5d26ef3621800543a54b46ac3e.

But it's from another author. Sorry that we don't handle this PR in time >_<

PR #88, opened 2015-07-08
CURATOR-218 Reorder ConnectionState process event

madrob:
Address a race condition in ConnectionState.process where it will
trigger watchers first before updating its own state. This can lead to
inconsistencies when blocking until connected.

### DIFF
--- a/curator-framework/src/test/java/org/apache/curator/framework/imps/TestBlockUntilConnected.java
+++ b/curator-framework/src/test/java/org/apache/curator/framework/imps/TestBlockUntilConnected.java
@@ -232,4 +232,28 @@ public class TestBlockUntilConnected extends BaseClassForTests
             CloseableUtils.closeQuietly(client);
         }
     }
+
+    /**
+     * Test that we are actually connected every time that we block until connection is established in a tight loop.
+     */
+    @Test
+    public void testBlockUntilConnectedTightLoop() throws InterruptedException
+    {
+        CuratorFramework client;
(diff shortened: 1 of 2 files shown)


curator-framework/src/test/java/org/apache/curator/framework/imps/TestBlockUntilConnected.java
```diff
@@ -232,4 +232,30 @@ public void run()
             CloseableUtils.closeQuietly(client);
         }
     }
+
+    /**
+     * Test that we are actually connected every time that we block until connection is established in a tight loop.
+     */
+    @Test
```
dragonsinth: qq: how long does this test take to run? not sure it matters, but some projects use @SlowTest or annotations like that

dragonsinth:
Remove the TreeCache changes?  Seems unrelated to the underlying bug in retrospect.

curator-framework/src/test/java/org/apache/curator/framework/imps/TestBlockUntilConnected.java
madrob: ~10 seconds on my laptop.

Randgalt:
I'm looking at this very carefully. It makes sense but this is old old code and deep deep in the heart of things. We should proceed with caution.

madrob:
Agree that this is old code that should be stable. Do you have a specific scenario/test case that you think this would cause issue with, or is it "gut feeling?"

madrob:
Merged.

PR #89, opened 2015-07-08
CURATOR-232 Consolidate Timing and Client

madrob:
Pull references to Timing and CuratorFramework client into BeforeMethod
and AfterMethod invocations so that we can manage them more efficiently.

### DIFF
--- a/curator-test/src/main/java/org/apache/curator/test/BaseClassForTests.java
+++ b/curator-test/src/main/java/org/apache/curator/test/BaseClassForTests.java
@@ -102,7 +102,7 @@ public class BaseClassForTests
     @AfterMethod
     public void teardown() throws Exception
     {
-        for (Closeable closeable : toClose)
+        for ( Closeable closeable : toClose )
         {
             // Don't need closeQuietly because we don't care that it advertises a throw
             Closeables.close(closeable, true);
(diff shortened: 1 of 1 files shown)


curator-framework/src/test/java/org/apache/curator/framework/FrameworkTest.java
```diff
@@ -0,0 +1,83 @@
+/*
```
Randgalt: I don't see enough value in this class to maintain it. In order for it to be used in the recipes module we'd have to build a test jar and include and/or create another test module. Create new clients isn't always consistent in the tests and it's only 1 line of code anyway. The closeable stuff could be pushed into BaseClassForTests by having a List<Closeable> that can be added to and auto closed on @After.

Randgalt:
I'm concerned about the number of changes in this PR. I like the idea of putting some common functionality in BaseClassForTests but I think it's too risky to touch all the tests. I suggest that we add the things to BaseClassForTests but leave the tests as is an updated them as needed over time.

curator-test/src/main/java/org/apache/curator/test/BaseClassForTests.java
```diff
@@ -55,6 +63,8 @@
         INTERNAL_RETRY_FAILED_TESTS = retryFailedTests;
     }
 
+    private List<Closeable> toClose;
```
Randgalt: Is this needed anymore given that Java 7 has Autocloseable?

tisonkun:

PR #90, opened 2015-07-13
CURATOR-234 Add slf4j-log4j12 test dependency

madrob:


Randgalt:
What problem is this solving?

madrob:
When you run tests, since there is currently no binding you wouldn't get any logging.

```
mdrob@mdrob-W530:~/workspace/curator$ cat curator-client/target/surefire-reports/org.apache.curator.TestSessionFailRetryLoop-output.txt
Configuring TestNG with: TestNG652Configurator
SLF4J: Failed to load class "org.slf4j.impl.StaticLoggerBinder".
SLF4J: Defaulting to no-operation (NOP) logger implementation
SLF4J: See http://www.slf4j.org/codes.html#StaticLoggerBinder for further details.
```

cammckenzie:
I'm a complete newbie to Maven, but is it possible to add test only dependencies? It would be nice to be able to just run unit tests with the slf4j-log4j12 dependency, but not enforce this as a dependency of the project itself (as it's not).  Alternatively, would it be better to use java.util.logging (via slf4j), as this presumably doesn't require an additional sfl4j bridge library?

madrob:
In the dependency section `<scope>test</scope>` indicates that it is test only, and not propagated as a transitive dependency. I don't know if there is a j.u.logging binding since I've never used it, but if it exists then that's fine too. Mostly, I just want output so that I can debug failing tests, and I don't care where it comes from.

cammckenzie:
Thanks for the clarification on the test scope stuff. I'm easy as to whether log4j or java.util.Logging is used, just figured that if we use java.util.Logging we don't need an additional library.

Randgalt:
My only concern is making sure it doesn't add a transitive dep to users. I don't think it will but we should test.

madrob:
According to the table at https://maven.apache.org/guides/introduction/introduction-to-dependency-mechanism.html#Dependency_Scope we will not inflict transitive dependencies to users. And if somehow we do, I will be more than happy to file the maven issue personally. :)

PR #92, opened 2015-07-22
TreeCache should inform listeners when its fully reconnected.

dragonsinth:
@Randgalt @madrob @cammckenzie

### DIFF
--- a/curator-recipes/src/main/java/org/apache/curator/framework/recipes/cache/TreeCache.java
+++ b/curator-recipes/src/main/java/org/apache/curator/framework/recipes/cache/TreeCache.java
@@ -779,6 +779,7 @@ public class TreeCache implements Closeable
             break;
 
         case LOST:
+            isInitialized.set(false);
             publishEvent(TreeCacheEvent.Type.CONNECTION_LOST);
             break;
 
--- a/curator-recipes/src/main/java/org/apache/curator/framework/recipes/cache/TreeCacheEvent.java
+++ b/curator-recipes/src/main/java/org/apache/curator/framework/recipes/cache/TreeCacheEvent.java
@@ -69,6 +69,9 @@ public class TreeCacheEvent
          * While the connection is down, the TreeCache will continue to have its state from before it lost
          * the connection and after the connection is restored, the TreeCache will emit normal child events
          * for all of the adds, deletes and updates that happened during the time that it was disconnected.
+         * </p><p>
+         * After reconnection, the cache will resynchronize its internal state with the server, then fire a
+         * {@link #INITIALIZED} event.
          * </p>
          */
         CONNECTION_RECONNECTED,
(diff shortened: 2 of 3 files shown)


dragonsinth:
Another idea that came out of this is TreeCache.isConnected() to reflect the state.

curator-recipes/src/main/java/org/apache/curator/framework/recipes/cache/TreeCache.java
```diff
@@ -758,6 +758,7 @@ private void handleStateChange(ConnectionState newState)
             break;
 
         case LOST:
```
Randgalt: LOST doesn't mean session lost. It only means that the retry policy has given up. Is this your intent?

curator-recipes/src/main/java/org/apache/curator/framework/recipes/cache/TreeCacheEvent.java
```diff
@@ -69,6 +69,9 @@
          * While the connection is down, the TreeCache will continue to have its state from before it lost
          * the connection and after the connection is restored, the TreeCache will emit normal child events
          * for all of the adds, deletes and updates that happened during the time that it was disconnected.
+         * </p><p>
+         * After reconnection, the cache will resynchronize its internal state with the server, then fire a
+         * {@link #INITIALIZED} event.
          * </p>
          */
         CONNECTION_RECONNECTED,
```
Randgalt: Shouldn't isInitialized.set(false) occur for SUSPENDED (i.e. any connection loss)?

curator-recipes/src/main/java/org/apache/curator/framework/recipes/cache/TreeCache.java
dragonsinth: Let me ramble a bit on my thinking, and hopefully you can help me clarify my thinking.

My intent is to follow the semantics around leaders/locks/leases/ephemeral nodes.  My intuition is that I should be seeing a LOST event right about the same time the server considers my session to be timed out.  So I want to inform TreeCache listeners here that they can no longer trust the internal tree state as being reasonably up to date.  Once the connection is re-established, I then re-walk the tree to synchronize to the server, after which I re-publish INITIALIZED, to inform listeners that they can now reasonably trust the tree state.

The use case I want to support is when someone is using TreeCache to watch for tree changes while using the same client to (say) put ephemeral nodes into the Tree.  The semantic I want to convey is that you can only trust that those nodes are really there if the TreeCache is currently in an initialized state.

curator-recipes/src/main/java/org/apache/curator/framework/recipes/cache/TreeCacheEvent.java
dragonsinth: I don't *think* so since that can be an extremely transient failure, and wouldn't cause ephemeral nodes to disappear.  Based on my testing this with ssh tunnels that I'd drop and then recreate, I'd see SUSPENDED immediately but only see LOST much later.

dragonsinth:
Rebased this, and wanted to ping for more feedback on this.  A friend of mine was just asking about this change.

cammckenzie:
This change probably doesn't make sense unless it's in the CURATOR 3.0 branch does it? With the state management in 2.X, the LOST event doesn't meant that the session is lost, just that Curator has given up trying to reconnect based on its retry policy. This could mean that the actual session has expired or it could mean that the session is still alive.

On the Curator 3.0 branch @Randgalt reimplemented the way that state management occurs for the LOST event. The default is now that LOST will be published when Curator thinks that the session has been LOST (based on a timer while disconnected), and this is now the default behaviour.

So, perhaps this can only be merged into 3.0? I don't think that the logic will work based on the 2.x version of the codebase.

dragonsinth:
I actually think it would be okay to land it in master.  Even if the event semantics are slightly fuzzier pre 3.0, the publishing of events actually does match TreeCache's internal operations-- internally TreeCache really does operate based on LOST and RECONNECTED, so the events will still be useful for anyone listening.

cammckenzie:
Yeah, I guess you're right. In that case, I think that it looks ok to me.

PR #94, opened 2015-07-23
CURATOR-239 - RetryPolicy that retries forever

leandronunes85:


curator-client/src/main/java/org/apache/curator/retry/RetryForever.java
```diff
@@ -0,0 +1,58 @@
...
+    @Override
+    public boolean allowRetry(int retryCount, long elapsedTimeMs, RetrySleeper sleeper)
+    {
+        try
+        {
+            sleeper.sleepFor(retryIntervalMs, TimeUnit.MILLISECONDS);
+        }
+        catch (InterruptedException e)
+        {
+            log.warn("Error occurred while sleeping", e);
```
Randgalt: When catching InterrptedException, always reset the state via:

{code}
Thread.currentThread().interrupt();
{code}

leandronunes85: Thks for the feedback. Already updated it

PR #95, opened 2015-08-02
CURATOR-241: Write updated data on reconnect

alexbrasetvik:
PersistentEphemeralNode can be initialised with certain data, then later updated. If the client reconnects, the replacing ephemeral should write the updated data.

### DIFF
--- a/curator-recipes/src/main/java/org/apache/curator/framework/recipes/nodes/PersistentEphemeralNode.java
+++ b/curator-recipes/src/main/java/org/apache/curator/framework/recipes/nodes/PersistentEphemeralNode.java
@@ -338,7 +338,7 @@ public class PersistentEphemeralNode implements Closeable
         this.data.set(Arrays.copyOf(data, data.length));
         if ( isActive() )
         {
-            client.setData().inBackground().forPath(getActualPath(), this.data.get());
+            client.setData().inBackground().forPath(getActualPath(), getData());
         }
     }
 
--- a/curator-recipes/src/main/java/org/apache/curator/framework/recipes/nodes/PersistentEphemeralNode.java
+++ b/curator-recipes/src/main/java/org/apache/curator/framework/recipes/nodes/PersistentEphemeralNode.java
@@ -338,11 +338,11 @@ public class PersistentEphemeralNode implements Closeable
         this.data.set(Arrays.copyOf(data, data.length));
         if ( isActive() )
         {
-            client.setData().inBackground().forPath(getActualPath(), getData());
+            client.setData().inBackground().forPath(getActualPath(), this.data.get());
         }
     }
 
-    byte[] getData() {
+    private byte[] getData() {
         return this.data.get();
     }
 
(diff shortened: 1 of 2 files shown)


curator-recipes/src/main/java/org/apache/curator/framework/recipes/nodes/PersistentEphemeralNode.java
```diff
@@ -338,10 +338,14 @@ public void setData(byte[] data) throws Exception
         this.data.set(Arrays.copyOf(data, data.length));
         if ( isActive() )
         {
-            client.setData().inBackground().forPath(getActualPath(), this.data.get());
+            client.setData().inBackground().forPath(getActualPath(), getData());
         }
     }
 
+    byte[] getData() {
```
cammckenzie: Is there a reason that this is package protected rather than private?

curator-recipes/src/main/java/org/apache/curator/framework/recipes/nodes/PersistentEphemeralNode.java
```diff
@@ -338,10 +338,14 @@ public void setData(byte[] data) throws Exception
         this.data.set(Arrays.copyOf(data, data.length));
         if ( isActive() )
         {
-            client.setData().inBackground().forPath(getActualPath(), this.data.get());
+            client.setData().inBackground().forPath(getActualPath(), getData());
```
cammckenzie: I think that this change isn't required as the data is being accessed directly from within this method.

curator-recipes/src/main/java/org/apache/curator/framework/recipes/nodes/PersistentEphemeralNode.java
alexbrasetvik: The idea was to make it available for future tests, but can make it private.

curator-recipes/src/main/java/org/apache/curator/framework/recipes/nodes/PersistentEphemeralNode.java
alexbrasetvik: Changed it so it was the same both places actually getting the data, but can revert this change. :)

curator-recipes/src/main/java/org/apache/curator/framework/recipes/nodes/PersistentEphemeralNode.java
cammckenzie: I'd mark it as private for now and it can be exposed later if needed. If it were to be for testing we mark these as @VisibleForTesting to be explicit as to why it's package protected.

curator-recipes/src/main/java/org/apache/curator/framework/recipes/nodes/PersistentEphemeralNode.java
cammckenzie: Yeah, fair enough. Leave it as is, it's probably nicer to have the data accessed in the same way.

alexbrasetvik:
Updated based on feedback. Want me to squash the commits together? :)

cammckenzie:
Don't worry about squashing the commits. Thanks for the patch.

+1 from me.

PR #96, opened 2015-08-22
CURATOR-214 - Implemented new ZK API to allow Stat objects to be retu…

cammckenzie:
…rned as part of the creation of a new ZNode.

### DIFF
--- a/curator-framework/src/main/java/org/apache/curator/framework/imps/CreateBuilderImpl.java
+++ b/curator-framework/src/main/java/org/apache/curator/framework/imps/CreateBuilderImpl.java
@@ -34,6 +34,8 @@ import org.apache.zookeeper.CreateMode;
 import org.apache.zookeeper.KeeperException;
 import org.apache.zookeeper.Op;
 import org.apache.zookeeper.data.ACL;
+import org.apache.zookeeper.data.Stat;
+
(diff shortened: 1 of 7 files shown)


curator-framework/src/main/java/org/apache/curator/framework/imps/CreateBuilderImpl.java
```diff
@@ -520,12 +539,127 @@ public void processResult(int rc, String path, Object ctx, String name)
...
+                acling.getAclList(operationAndData.getData().getPath()),
+                createMode,
+                new AsyncCallback.Create2Callback() {
+                    
+                    @Override
+                    public void processResult(int rc, String path, Object ctx, String name, Stat stat) {
+                        trace.commit();
+                        
+                        //Need to manually populate the provided stat here.
+                        storingStat.setAversion(stat.getAversion());
```
Randgalt: is this the only place this is done? Seems like a candidate for a utility.

cammckenzie: Yeah, this is the only place that it's done which is why I left it inline.

Randgalt:
+1

PR #97, opened 2015-08-23
[CURATOR-247] Extend Curator's connection state to support SESSION_LOST

Randgalt:
This is a significant change. Please review carefully and let's have a lot of discussion. Because the new behavior is so much better (and more consistent with expectations) <strong>I've enabled it by default</strong>.

Major differences from the older behavior are:
* Session/connection timeouts are no longer managed by the low-level client. They are managed by the CuratorFramework instance. There should be no noticeable differences.
* Prior to 3.0.0, each iteration of the retry policy would allow the connection timeout to elapse if the connection hadn't yet succeeded. This meant that the true connection timeout was the configured value times the maximum retries in the retry policy. This longstanding issue has been address. Now, the connection timeout can elapse only once for a single API call.
* MOST IMPORTANTLY! Prior to 3.0.0, ConnectionState.LOST did not imply a lost session (much to the confusion of users). Now, Curator will set the LOST state only when it believes that the ZooKeeper session has expired. ZooKeeper connections have a session. When the session expires, clients must take appropriate action. In Curator, this is complicated by the fact that Curator internally manages the ZooKeeper connection. Now, Curator will set the LOST state when any of the following occurs: a) ZooKeeper returns a Watcher.Event.KeeperState.Expired or KeeperException.Code.SESSIONEXPIRED; b) Curator closes the internally managed ZooKeeper instance; c) The configured session timeout elapses during a network partition.

Something important to consider. Given the significance of this change it really should be part of 3.0.0 (i.e. a major version bump) but if we merge it into 3.0.0 now it will be harder to maintain master and 3.0.0 as separate branches. Some git expertise is needed here on how to manage this.


Randgalt:
Also, FYI, each test is now run twice. Once with the old connection handling and once with the new handling.

dragonsinth:
Very cool.  Before I dive in, couple of high level questions.

1) Are we able to precisely articulate the relationship between session lost and existing watchers?  For example, an acceptable answer might be "When a session is lost, all outstanding watchers will fire a lost event.  Afterwards, all watchers will be removed.  Client code should create new watchers if the session is re-established.  If a connection can be re-established before the session is lost, all watchers will remain active."

2) Assume a timing discrepancy between the curator client and ZK server, such that the client assumes a session is lost by timeout, but in a (racy) way, the connection is successfully re-established to the server and the server does not drop the session.  Can we be sure to resolve this race, such that if session lost events have already fired or started firing, the client forcibly creates a new session instead of using the old one?

cammckenzie:
Nice work Jordan,
I just had a quick preliminary look through stuff and I had one observation. For the timing stuff that checks if we've been in a SUSPENDED state for long enough for the server to have timed out the connection, it's using the session timeout specified in the builder.

Isn't there some sort of negotiation when connecting to Zookeeper in regards to session timeout? I.e. It's possible for Zookeeper to specify that a different session timeout will be used rather than the one that the client wants?

Randgalt:
> Isn't there some sort of negotiation when connecting to Zookeeper in regards to session timeout? 
Oh right. Good point. I don't think that value is accessible on the client side. But, it could be calculated if we had the server configuration values. I think we can get those through the new config api right? I'll look into this.

cammckenzie:
Looks like you can call getSessionTimeout() on the ZooKeeper object and it returns the negotiated session timeout. You would need to call it every time a connection / reconnection occurred.

Randgalt:
> Looks like you can call getSessionTimeout() on the ZooKeeper object and it returns the negotiated session timeout. You would need to call it every time a connection / reconnection occurred.

Excellent. I'll do that.

Randgalt:
> Are we able to precisely articulate the relationship between session lost and existing watchers

Excellent point. I missed that. I just did a test and, unfortunately, when the client is closed the watchers are not notified. I'll need to emulate a session expiration event of some kind. I'll look into this.

Randgalt:
> Assume a timing discrepancy between the curator client and ZK server, such that the client assumes a session is lost by timeout, but in a (racy) way, the connection is successfully re-established to the server and the server does not drop the session. Can we be sure to resolve this race, such that if session lost events have already fired or started firing, the client forcibly creates a new session instead of using the old one?

This isn't an issue because when the LOST state is generated the current client is also closed so the session will expire in any event.

Randgalt:
This is hilarious. In July 2013, I submitted a patch to add some testable hooks to ZooKeeper. I forgot all about it was never touched (https://issues.apache.org/jira/browse/ZOOKEEPER-1730). Well, it's now merged in to ZK 3.5 and, guess what, it has a SESSION expiration injection method! This is Twilight Zone level stuff. I was on my computer writing a hack to inject the session event and digging into ZooKeeper source when I saw what looked like my code. Imagine my surprise...

Randgalt:
Any more thoughts on this? Can it be merged?

cammckenzie:
I haven't given it more than a cursory look. If Scott's had a proper look and is OK with it then I'm happy, otherwise I can have a look, but it's probably not going to be for a few days.

Randgalt:
Let's wait until you and Mike have a look if you both want.

curator-client/src/main/java/org/apache/curator/connection/ConnectionHandlingPolicy.java
```diff
@@ -0,0 +1,72 @@
...
+    }
+
+    /**
+     * Check timeouts. NOTE: this method is only called when an attempt to access to the ZooKeeper instances
+     * is made and the connection has not completed.
+     *
+     * @param hasNewConnectionString proc to call to check if there is a new connection string. Important: the internal state is cleared after
+     *                               this is called so you MUST handle the new connection string if <tt>true</tt> is returned
+     * @param connectionStartMs the epoch/ms time that the connection was first initiated
+     * @param sessionTimeoutMs the configured session timeout in milliseconds
```
cammckenzie: Should this be the negotiated session timeout rather than the configured session timeout? Looking at the code that calls into this interface, it would seem that the configured timeout from the build is being used. Is that correct though? I would have thought that the negotiated timeout was what was important as it's the source of truth.

curator-framework/src/main/java/org/apache/curator/framework/imps/CuratorFrameworkImpl.java
```diff
@@ -103,23 +106,34 @@
...
+                        CuratorEvent event = new CuratorEventImpl(CuratorFrameworkImpl.this, CuratorEventType.WATCHED, watchedEvent.getState().getIntValue(), unfixForNamespace(watchedEvent.getPath()), null, null, null, null, null, watchedEvent, null, null);
+                        processEvent(event);
+                    }
+                },
+                builder.getRetryPolicy(),
+                builder.canBeReadOnly(),
+                builder.getConnectionHandlingPolicy()
+            );
 
+        internalConnectionHandler = builder.getConnectionHandlingPolicy().isEmulatingClassicHandling() ? new ClassicInternalConnectionHandler() : new StandardInternalConnectionHandler();
```
cammckenzie: Is there a reason that this needs to be a separate instance rather than just a reference to the connection handler from the builder?

curator-framework/src/main/java/org/apache/curator/framework/state/ConnectionStateManager.java
```diff
@@ -240,27 +248,37 @@ private void processEvents()
     {
         try
         {
-            while ( !Thread.currentThread().isInterrupted() )
+            while ( !Thread.currentThread().isInterrupted() && (state.get() == State.STARTED) )
             {
-                final ConnectionState newState = eventQueue.take();
-
-                if ( listeners.size() == 0 )
+                int pollMaxMs = (sessionTimeoutMs * 2) / 3; // 2/3 of session timeout
```
cammckenzie: This is the configured session timeout MS. Will this cause issues if the negotiated timeout is different?

I guess it just means that it will potentially take 4/3 * configured session timeout to determine if the session has timed out during network partition?

Randgalt: Good point. I should use the negotiated timeout.

cammckenzie:
Other than those minor comments, I think this looks good to me.

PR #98, opened 2015-08-24
[CURATOR-248] Introduce pluggable error handlers

Randgalt:


Randgalt:
This is now feature complete.

cammckenzie:
Looks good to me Jordan. My only nit would be that the name ErrorPolicy doesn't really convey what it's an error policy. Would ConnectionStateErrorPolicy or something like that be more descriptive?

Randgalt:
I'm OK with renaming it.

PR #99, opened 2015-08-24
CURATOR-233 Fix DistributedDoubleBarrier::leave

madrob:
DDB::leave would not respect the wait time parameters correctly, and
would sometimes return true, even though not all clients had left the
barrier.


tisonkun:

PR #100, opened 2015-08-28
[CURATOR-253] Use new injectSessionExpiration() in the test module

Randgalt:


curator-recipes/src/main/java/org/apache/curator/framework/recipes/nodes/PersistentEphemeralNode.java
```diff
@@ -111,11 +112,26 @@ public void stateChanged(CuratorFramework client, ConnectionState newState)
...
+                        e.printStackTrace();
+                    }
+                }
                 createNode();
             }
         }
     };
 
+    @VisibleForTesting
+    volatile CountDownLatch debugReconnectLatch = null;
```
madrob: Why does this need to be volatile?

curator-recipes/src/test/java/org/apache/curator/framework/recipes/nodes/TestPersistentEphemeralNode.java
```diff
@@ -634,11 +640,6 @@ private CuratorFramework newCurator() throws IOException
         return client;
     }
 
-    public void killSession(CuratorFramework curator) throws Exception
-    {
-        KillSession.kill(curator.getZookeeperClient().getZooKeeper(), curator.getZookeeperClient().getCurrentConnectionString());
```
madrob: Why not leave the method and have it do the new thing instead of repeating the same snippet in a bunch of places?

curator-recipes/src/main/java/org/apache/curator/framework/recipes/nodes/PersistentEphemeralNode.java
Randgalt: Can't predict what thread it will come from. In the current use case it doesn't need to be, but who knows in the future.

curator-test/src/main/java/org/apache/curator/test/KillSession.java
```diff
@@ -18,24 +18,12 @@
...
- *
- * <blockquote>
- *     In the case of testing we want to cause a problem, so to explicitly expire a session an
- *     application connects to ZooKeeper, saves the session id and password, creates another
- *     ZooKeeper handle with that id and password, and then closes the new handle. Since both
- *     handles reference the same session, the close on second handle will invalidate the session
- *     causing a SESSION_EXPIRED on the first handle.
- * </blockquote>
  */
 public class KillSession
```
madrob: New method suggestion: KillSession::kill(CuratorFramework) and that can extract the zookeeper bits on it's own. More robust for future improvement, maybe?

Randgalt: Yeah, probably a good thing. I'll add it.

Randgalt: Hmm - the only problem is where to put it. We don't currently build test Jars. So, I'd have to have a version for each module. That's ugly.

curator-recipes/src/test/java/org/apache/curator/framework/recipes/cache/TestTreeCache.java
```diff
@@ -377,7 +377,6 @@ public void testKilledSession() throws Exception
         assertEvent(TreeCacheEvent.Type.NODE_ADDED, "/test/me");
 
         KillSession.kill(client.getZookeeperClient().getZooKeeper(), server.getConnectString());
-        assertEvent(TreeCacheEvent.Type.CONNECTION_SUSPENDED);
```
cammckenzie: Why does the CONNECTION_SUSPENDED event no longer occur? Is this because of the new lifecycle stuff for session loss?

Randgalt: Killing the session goes immediately to LOST. It actually was always a possibility but now it does it so fast that it's always the case.

curator-recipes/src/test/java/org/apache/curator/framework/recipes/nodes/TestPersistentEphemeralNode.java
cammckenzie: I was wondering this too

Randgalt: Because no other code uses it. We should have a general solution.

cammckenzie: Maybe it should go into the BaseClassForTests?

curator-recipes/src/test/java/org/apache/curator/framework/recipes/cache/TestTreeCache.java
cammckenzie: Ok

PR #101, opened 2015-08-31
CURATOR-254 - Modified the test assertions so that they ensure that t…

cammckenzie:
…he theoretical maximum queue size is not exceeded. The previous limit was exceedable depending on how quick consumers were pulling data off the queue.


Randgalt:
LGTM

PR #103, opened 2015-09-06
[CURATOR-259] Added Locker which uses Java 7's try-with-resource feature to make locking more reliable

Randgalt:


cammckenzie:
Looks OK to me. Do we need to provide an option to acquire a lock with no timeout that would just call into lock.acquire()?

Randgalt:
I was thinking about that but I've come to the opinion that that that's inherently unsafe. But, if you think it's good for completeness I'll add it.

cammckenzie:
While I agree that it's inherently unsafe and probably not really going to be used in practice. I think that if we're going to expose that interface on the lock itself then we should probably expose it on this utility class as well.

I don't have a strong opinion either way though really.

cammckenzie:
Looks good to me.

pvorb:
This class is no longer available. Is this correct?

It is still mentioned on http://curator.apache.org/utilities.html, which is why I found this PR.

cammckenzie:
Curator 2.x only supports Java 1.6 so can't support the auto close features
in Java 1.7

Curator 3.x supports Java 1.7. The Locker class is present in this version.

cheers

pvorb:
@cammckenzie Thanks for the clarification. I didn't notice that I wasn't using the latest version of Curator since I use it through spring-integration-zookeeper.

PR #104, opened 2015-09-10
[CURATOR-261] Removed usages of Javaassist

Randgalt:
Removed usages of Javaassist and simplified how the in-memory ZK server is created and run. Tests should be more reliable and the testing server should respond faster


Randgalt:
This has turned into more of a refine, tests task

PR #105, opened 2015-09-15
Better session expiration management

Randgalt:
Added an option for session expiration management to be a fraction of the negotiated session timeout. This is meant to account for timing/network differences between the client and server.


cammckenzie:
Looks ok to me.

PR #106, opened 2015-09-22
[CURATOR-264] Duplicate ephemeral nodes with same owner id

Randgalt:
CURATOR-45 added findAndDeleteProtectedNodeInBackground to handle cases where a protected node can get lost. However, the code wasn't correctly handling namespaces


cammckenzie:
So, the problem was that if a namespace was being used that it was being applied to the path twice? Once in the forPath() and again in findAndDeleteProtectedNodeInBackground()?

Randgalt:
It was being un-namespaced inside of the CreateBuilder. Then, that resolved path was passed via findAndDeleteProtectedNodeInBackground to the Curator routines which would resolve it again. So, that's twice, right? Yes.

PR #107, opened 2015-09-26
[CURATOR-265] Dynamic Reconfig APIs needs some work

Randgalt:
* Completely re-wrote the tests
* The APIs weren't correct for all permutations
* adding() API was not correct. In ZK, it's called "newMembers" and is mutually exclusive with join/leave


Randgalt:
Oops - wrong base

PR #110, opened 2015-09-27
[CURATOR-267] First pass implementation of group membership

Randgalt:


curator-recipes/src/main/java/org/apache/curator/framework/recipes/nodes/GroupMember.java
```diff
@@ -0,0 +1,154 @@
...
+        return builder.build();
+    }
+
+    /**
+     * Given a full ZNode path, return the member ID
+     *
+     * @param path full ZNode path
+     * @return id
+     */
+    public String idFromPath(String path)
```
cammckenzie: Any reason this is public?

cammckenzie:
Other than my minor comment, this looks good to me.

PR #111, opened 2015-10-02
Support delete().quietly() and create().orSetData()

Randgalt:


cammckenzie:
Looks ok to me, though when I ran the tests an unrelated test is failling.

BasicTests:testFactory is failing.

This appears to be because the constructor for CuratorZookeeperClient does not use the ZookeeperFactory reference passed to it, and just instantiates a DefaultZookeeperFactory. I presume this change was made by some other fix? If I change the constructor to use the passed in factory, the test passes ok.

PR #115, opened 2015-10-22
[CURATOR-274] v2.9.0 broken EnsurePath behavior

Randgalt:
Previous change to create containers broke some old behavior. Old EnsurePath class only created parents once. So, introduce new EnsureContainers class to do similar but for containers. Also, make usage in a protected method so that it can be turned off by users

### DIFF
--- a/curator-framework/src/main/java/org/apache/curator/framework/EnsureContainers.java
+++ b/curator-framework/src/main/java/org/apache/curator/framework/EnsureContainers.java
@@ -18,7 +18,6 @@
  */
 package org.apache.curator.framework;
 
-import org.apache.curator.utils.EnsurePath;
 import java.util.concurrent.atomic.AtomicBoolean;
 
 /**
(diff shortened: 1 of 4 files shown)


JasonRosenberg:
should there be tests to verify this behavior?

Randgalt:
There's already a test for ensuring path in TestPathChildrenCache

Randgalt:
I changed my mind and added a few tests

curator-framework/src/main/java/org/apache/curator/framework/EnsureContainers.java
```diff
@@ -0,0 +1,65 @@
...
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package org.apache.curator.framework;
+
+import org.apache.curator.utils.EnsurePath;
```
cammckenzie: Unused import?

cammckenzie:
Other than that minor import, looks good to me.

PR #116, opened 2015-11-02
[CURATOR-275] Allow service instances to be disabled temporarily

joelittlejohn:
Service providers will omit any instance that is marked 'enabled: false'.
The low level API provided by ServiceDiscovery (e.g. queryForInstance,
queryForInstances) will continue to return all instances, including
those that are marked 'enabled: false'.

This change allows instances to be temporarily removed from service
discovery via an update to the instance node data.

### DIFF
--- a/curator-x-discovery/src/main/java/org/apache/curator/x/discovery/ServiceInstance.java
+++ b/curator-x-discovery/src/main/java/org/apache/curator/x/discovery/ServiceInstance.java
@@ -42,6 +42,7 @@ public class ServiceInstance<T>
     private final long          registrationTimeUTC;
     private final ServiceType   serviceType;
     private final UriSpec       uriSpec;
+    private final boolean       enabled;
 
     /**
      * Return a new builder. The {@link #address} is set to the ip of the first
@@ -75,7 +76,7 @@ public class ServiceInstance<T>
      * @param serviceType type of the service
      * @param uriSpec the uri spec or null
      */
-    public ServiceInstance(String name, String id, String address, Integer port, Integer sslPort, T payload, long registrationTimeUTC, ServiceType serviceType, UriSpec uriSpec)
+    public ServiceInstance(String name, String id, String address, Integer port, Integer sslPort, T payload, long registrationTimeUTC, ServiceType serviceType, UriSpec uriSpec, boolean enabled)
     {
         name = Preconditions.checkNotNull(name, "name cannot be null");
         id = Preconditions.checkNotNull(id, "id cannot be null");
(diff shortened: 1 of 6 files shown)


serranom:
why not just unregister and then re-register the service?

joelittlejohn:
There's some [more discussion about this on the mailing list](http://mail-archives.apache.org/mod_mbox/curator-user/201510.mbox/%3CCABcULGHWQo_u0r6-KMiVQWa0z1Y%2Bt8paH%2BryOKbg6LUJ8HBTtQ%40mail.gmail.com%3E).

The goal here is to allow a single instance to be removed temporarily from discovery by updating only Zookeeper - without having to rely on all clients implementing some method of forcing unregister/reregister. A few users have implemented their own solution to this, but this feels like a very useful core feature that ideally shouldn't require further customization of all clients.

Imagine this as similar to taking an instance out of a load balancer (e.g. when troubleshooting some problem affecting one instance). It's very useful for Curator to be able to avoid sending instances to a node without having that node unregister, since it requires no action or implementation from the client.

serranom:
Makes sense now.  Thanks.  I do wonder how you would envision this being
used at the API level.  Changing data just at the ZK level implies internal
knowledge of the znode layout which doesn't seem to be intended to be
exposed by the service-discovery framework.

joelittlejohn:
It's a very good point @serranom. 

There's nothing stopping someone from doing this via the API (I don't think) since you can simply create a new ServiceInstance (based on the field values of the instance you want to disable) but pass `false` as the enabled value, then use [ServiceDiscovery#updateService(ServiceInstance)](https://github.com/joelittlejohn/curator/blob/4776a2daacb3a776e9dd0bdc214ed4987fef96c9/curator-x-discovery/src/main/java/org/apache/curator/x/discovery/ServiceDiscovery.java#L48) to apply the change. This is the kind of thing a dashboard/tool (that has no knowledge of the znode layout) could do. So in this way, all ServiceInstance fields are always updatable via the API.

I saw this as an administrative task that (as the least worst option) allows an update to the znode to disable an instance. I thought of this as a useful administrative hook but not necessarily an operation that needs to part of the core API. If we did want _improved_ support for this in the API it would be easy to add this e.g. 

`ServiceDiscovery#enable(ServiceInstance)` and `ServiceDiscovery#disable(ServiceInstance)`

or

`ServiceInstance#enable()` and `ServiceInstance#disable()`
(followed by `ServiceDiscovery#updateService(ServiceInstance)`)

joelittlejohn:
Thanks to whoever merged this! :+1:

Randgalt:
You're welcome. I'm sorry it took so long. We need more Curator committers.

PR #119, opened 2015-11-13
CURATOR-281: Usage of TestingServer fails when upgrading to 3.0

granthenke:
Replace testng assertion with java assertion


madrob:
Doesn't that effectively remove the assertion, though? Java assertions are not enabled by default, you need to specify the `-ea` flag for that.

granthenke:
Good point, I could instead throw an IllegalStateException if the result is not true

granthenke:
@madrob Does this follow up patch look better?

madrob:
Yea, it looks good, I'll get that in shortly.

granthenke:
Closing as this was committed, but not auto-closed since it was in the 3.0 branch and not the current default branch.

PR #120, opened 2015-12-28
[CURATOR-209] Better handling of background errors

Randgalt:
1. Don't queue background operation if the client is closed
2. Moved findAndDeleteProtectedNodeInBackground code into separate operation that is processed through the standard Curator background code. This way, retries are applied (with sleep), etc. In the previous implementation, errors caused the background check to be run immediately and infinitely.


cammckenzie:
Looks good to me

PR #123, opened 2016-01-12
[CURATOR-287] Deprecated PersistentEphemeralNode in favor of PersistentNode which is the same code but now accepts any createmode

Randgalt:


Randgalt:
PersistentNode is a straight copy of PersistentEphemeralNode. The only differences:

* Now accepts any CreateMode
* Logs a message if the node that exists doesn't match the mode requested (i.e. a user reported that an error caused another process to create the node as persistent when ephemeral was wanted, etc.)

curator-recipes/src/main/java/org/apache/curator/framework/recipes/nodes/PersistentNode.java
```diff
@@ -0,0 +1,382 @@
...
+                    path = event.getPath();
+                    nodeExists = true;
+                }
+                else if ( event.getResultCode() == KeeperException.Code.OK.intValue() )
+                {
+                    path = event.getName();
+                }
+                else if ( event.getResultCode() == KeeperException.Code.NOAUTH.intValue() )
+                {
+                    log.warn("Client does not have authorisation to write ephemeral node at path {}", event.getPath());
```
cammckenzie: This log message refers to ephemeral which is no longer necessarily the case.

curator-recipes/src/main/java/org/apache/curator/framework/recipes/nodes/PersistentNode.java
```diff
@@ -0,0 +1,382 @@
...
+     *
+     * @return node path or null
+     */
+    public String getActualPath()
+    {
+        return nodePath.get();
+    }
+
+    /**
+     * Set data that ephemeral node should set in ZK also writes the data to the node
```
cammckenzie: Javadoc should not refer to ephemeral

curator-recipes/src/test/java/org/apache/curator/framework/recipes/nodes/TestPersistentNode.java
```diff
@@ -0,0 +1,62 @@
...
+import org.apache.curator.utils.CloseableUtils;
+import org.apache.zookeeper.CreateMode;
+import org.testng.Assert;
+import org.testng.annotations.Test;
+import java.util.concurrent.TimeUnit;
+
+public class TestPersistentNode extends BaseClassForTests
+{
+    @Test
+    public void testBasic() throws Exception
```
cammckenzie: Maybe additional tests for the sequential case?

curator-recipes/src/main/java/org/apache/curator/framework/recipes/nodes/PersistentNode.java
```diff
@@ -0,0 +1,382 @@
...
+        if ( !state.compareAndSet(State.STARTED, State.CLOSED) )
+        {
+            return;
+        }
+
+        client.getConnectionStateListenable().removeListener(connectionStateListener);
+
+        try
+        {
+            deleteNode();
```
cammckenzie: Does it make sense to delete the node on close if it's persistent?

cammckenzie:
In general, looks ok to me. Is there a particular use case that you envisage for a persistent PersistentNode?

Randgalt:
particular use case that you envisage for a persistent PersistentNode

Yes - at my new job :)

curator-recipes/src/main/java/org/apache/curator/framework/recipes/nodes/PersistentNode.java
Randgalt: I think it makes sense still. The idea is that the node persists until the instance is closed.

cammckenzie: OK

PR #125, opened 2016-01-13
[CURATOR-208] General fix for catch-alls throughout the code and interrupts

Randgalt:
General fix for catch-alls throughout the code. This is overkill but it strikes me as the safest way to address the problem. Everywhere there is a catch-all add a check for InterruptedException and reset the thread's interrupted state

### DIFF
--- a/curator-framework/src/main/java/org/apache/curator/framework/imps/CuratorFrameworkImpl.java
+++ b/curator-framework/src/main/java/org/apache/curator/framework/imps/CuratorFrameworkImpl.java
@@ -271,6 +271,7 @@ public class CuratorFrameworkImpl implements CuratorFramework
         }
         catch ( Exception e )
         {
+            ThreadUtils.checkInterrupted(e);
             handleBackgroundOperationException(null, e);
         }
     }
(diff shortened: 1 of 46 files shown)


curator-framework/src/main/java/org/apache/curator/framework/imps/CuratorFrameworkImpl.java
```diff
@@ -257,8 +258,7 @@ public void stateChanged(CuratorFramework client, ConnectionState newState)
 
             client.start();
 
-            executorService = Executors.newFixedThreadPool(2, threadFactory);  // 1 for listeners, 1 for background ops
-
+            executorService = Executors.newSingleThreadScheduledExecutor(threadFactory);
```
cammckenzie: Is there a reason that this has gone from a 2 thread pool to a single thread pool? Looking at the old code, it would seem that only 1 is used (for the background loop), but just double checking.

Randgalt: The comment was wrong. This must have changed at some point. I noticed only 1 was being used and corrected it.

cammckenzie:
Looks ok to me I think. I agree that it's kind of brute force, but I can't think of a cleaner way of doing it without a complete audit of the code.

PR #126, opened 2016-01-18
[CURATOR-290] Make sure NamespaceWatcherMap is cleared

Randgalt:
Make sure NamespaceWatcherMap is cleared when the corresponding watcher is removed via new APIs. Added tests to ensure this.


cammckenzie:
Looks good to me.

PR #127, opened 2016-01-28
CURATOR-294

dragonsinth:
@Randgalt


njhill:
changes look great, exactly what I had in mind

PR #128, opened 2016-01-30
CURATOR-294: Optimize TreeCache, fix possible concurrency issue

dragonsinth:


curator-recipes/src/main/java/org/apache/curator/framework/recipes/cache/ChildData.java
```diff
@@ -25,15 +25,15 @@
```
cammckenzie: Unused import of AtomicReference

cammckenzie:
Other than the unused import, looks good to me.

curator-recipes/src/main/java/org/apache/curator/framework/recipes/cache/ChildData.java
```diff
@@ -92,7 +92,7 @@ public int hashCode()
     {
         int result = path != null ? path.hashCode() : 0;
         result = 31 * result + (stat != null ? stat.hashCode() : 0);
-        result = 31 * result + (data != null ? Arrays.hashCode(data.get()) : 0);
+        result = 31 * result + Arrays.hashCode(data);
```
Randgalt: data can legally be null. null checks need to be added throughout this class.

curator-recipes/src/main/java/org/apache/curator/framework/recipes/cache/ChildData.java
```diff
@@ -126,12 +126,7 @@ public Stat getStat()
      */
     public byte[] getData()
     {
-        return data.get();
-    }
-
-    void clearData()
-    {
-        data.set(null);
+        return data;
```
Randgalt: clearData() needs to remain. PathChildrenCache uses it to release memory. The data is needed temporarily and is cleared afterwards.

Randgalt:
Data is an AtomicReference because PathChildrenCache needs to be able to clear it. As I recall, data is needed temporarily and is cleared if the object is marked not to cache data.

curator-recipes/src/main/java/org/apache/curator/framework/recipes/cache/ChildData.java
dragonsinth: Arrays.hashCode handles null internally

curator-recipes/src/main/java/org/apache/curator/framework/recipes/cache/PathChildrenCache.java
```diff
@@ -448,7 +448,10 @@ public boolean clearDataBytes(String fullPath, int ifVersion)
         {
             if ( (ifVersion < 0) || (ifVersion == data.getStat().getVersion()) )
             {
-                data.clearData();
+                if ( data.getData() != null )
+                {
+                    currentData.replace(fullPath, data, new ChildData(data.getPath(), data.getStat(), null));
+                }
```
dragonsinth: @Randgalt this is the replacement for `clearData()`.  Instead of the ChildData object being mutable, we just replace the ChildData entirely with a copy that has no data.

curator-recipes/src/main/java/org/apache/curator/framework/recipes/cache/ChildData.java
dragonsinth: removed that call site, see below

curator-recipes/src/main/java/org/apache/curator/framework/recipes/cache/ChildData.java
dragonsinth: done

curator-recipes/src/main/java/org/apache/curator/framework/recipes/cache/PathChildrenCache.java
Randgalt: OK - I see

Randgalt:
LGTM

PR #130, opened 2016-02-02
[CURATOR-297] Handle quick close after start for PersistentNode

Randgalt:
Because start() creates the node async, if close is called immediately, it might occur before the node creation. Handle this edge case by having the background response delete (async) the node if the instance has been closed


cammckenzie:
LGTM

PR #131, opened 2016-02-05
Major re-work of Watcher wrappers

Randgalt:
Major re-work of Watcher wrappers. It really isn't necessary to keep a map of watchers. It was originally done so that watcher identity was maintained, but this can be achieved much easier using special-purpose hashCode() and equals() on NamespaceWatcher.


curator-framework/src/main/java/org/apache/curator/framework/imps/CuratorFrameworkImpl.java
```diff
@@ -240,11 +238,7 @@ public void createContainers(String path) throws Exception
     @Override
     public void clearWatcherReferences(Watcher watcher)
     {
-        NamespaceWatcher namespaceWatcher = namespaceWatcherMap.remove(watcher);
-        if ( namespaceWatcher != null )
-        {
-            namespaceWatcher.close();
-        }
```
dragonsinth: Just a question here.  The super class doc reads `@deprecated As of ZooKeeper 3.5 Curators recipes will handle removing watcher references when they are no longer used.`

Is this still the most accurate comment?  My (loose) understanding is that even the recipes themselves don't need to worry about this as the framework will handle it for them.

Randgalt: Recipes have to do a tiny bit of work. They need to get a specially wrapped instance by calling CuratorFramework.newWatcherRemoveCuratorFramework() and then calling removeWatchers() when done.

curator-framework/src/main/java/org/apache/curator/framework/imps/NamespaceWatcher.java
```diff
@@ -75,4 +89,53 @@ else if ( curatorWatcher != null )
...
+        return false;
+    }
+
+    @Override
+    public int hashCode()
+    {
+        int result = actualWatcher != null ? actualWatcher.hashCode() : 0;
+        result = 31 * result + unfixedPath.hashCode();
+        result = 31 * result + (curatorWatcher != null ? curatorWatcher.hashCode() : 0);
+        return result;
```
dragonsinth: shouldn't this just be `return Objects.hashCode(actualWatcher)`?

If the goal is to make this hashCode() and equals() same as underlying watcher, why add in the other components to the hash?

Randgalt: It needs to match the watcher and the path - that's how ZooKeeper stores it.

curator-framework/src/main/java/org/apache/curator/framework/imps/NamespaceWatcher.java
```diff
@@ -75,4 +89,53 @@ else if ( curatorWatcher != null )
...
+            return curatorWatcher != null ? curatorWatcher.equals(watcher.curatorWatcher) : watcher.curatorWatcher == null;
+        }
+
+        //noinspection SimplifiableIfStatement
+        if ( Watcher.class.isAssignableFrom(o.getClass()) )
+        {
+            return actualWatcher == o;
+        }
+
+        return false;
```
dragonsinth: I think you can simplify all this

```
        if ( this == o )
        {
            return true;
        }

        if ( o instanceof NamespaceWatcher )
        {
            o = ((NamespaceWatcher)o).actualWatcher;
        }

        return Objects.equal(this.actualWatcher, o);
```

curator-framework/src/main/java/org/apache/curator/framework/imps/Watching.java
```diff
@@ -48,15 +53,30 @@
...
+            if ( client.getWatcherRemovalManager() != null )
+            {
+                client.getWatcherRemovalManager().add(namespaceWatcher);
+            }
+        }
+
+        return namespaceWatcher;
     }
 
     boolean isWatched()
```
dragonsinth: unrelated to PR, but a name like `needsImplicitWatch` might better explain the purpose

curator-framework/src/main/java/org/apache/curator/framework/imps/CuratorFrameworkImpl.java
dragonsinth: Cool, maybe update the comment on super decl to reflect this?

curator-framework/src/main/java/org/apache/curator/framework/imps/NamespaceWatcher.java
dragonsinth: Ah, I see what's going on now.

curator-framework/src/main/java/org/apache/curator/framework/imps/NamespaceWatcher.java
dragonsinth: Actually, disregard this comment, missed the `curatorWatcher` field

curator-framework/src/main/java/org/apache/curator/framework/imps/NamespaceWatcher.java
```diff
@@ -75,4 +89,53 @@ else if ( curatorWatcher != null )
...
+            }
+
+            //noinspection SimplifiableIfStatement
+            if ( actualWatcher != null ? !actualWatcher.equals(watcher.actualWatcher) : watcher.actualWatcher != null )
+            {
+                return false;
+            }
+            return curatorWatcher != null ? curatorWatcher.equals(watcher.curatorWatcher) : watcher.curatorWatcher == null;
+        }
+
```
dragonsinth: Okay, I think there's a missing case here?

```
        if ( o instanceof CuratorWatcher )
        {
            return curatorWatcher == o;
        }
```

curator-framework/src/main/java/org/apache/curator/framework/imps/NamespaceWatcher.java
```diff
@@ -75,4 +89,53 @@ else if ( curatorWatcher != null )
...
+        //noinspection SimplifiableIfStatement
+        if ( Watcher.class.isAssignableFrom(o.getClass()) )
```
dragonsinth: why not just `o instanceof Watcher`?

curator-framework/src/main/java/org/apache/curator/framework/imps/NamespaceWatcher.java
```diff
@@ -75,4 +89,53 @@ else if ( curatorWatcher != null )
...
+        if ( this == o )
+        {
+            return true;
+        }
+        if ( o == null )
+        {
+            return false;
+        }
+
+        if ( getClass() == o.getClass() )
```
dragonsinth: Why not just `o instanceof NamespaceWatcher`?

Randgalt: This is IntelliJ's auto-gen again the entire equals()/hashCode()/toString() methods are auto generated.

curator-framework/src/main/java/org/apache/curator/framework/imps/NamespaceWatcher.java
Randgalt: I don't think it can actually happen as my main concern is inside ZK. But, for completeness it should be added.

PR #134, opened 2016-02-10
Don't allow NamespaceWatcher to compare equal to a raw Watcher

Randgalt:
Don't allow NamespaceWatcher to compare equal to a raw Watcher or CuratorWatcher. It can't be done reciprocally in a raw Watcher and is actually unnecessary. NOTE: a Curator Tech Note should be written to warn that Watchers are wrapped internally


dragonsinth:
There will be a test failure, see https://github.com/apache/curator/pull/135

Randgalt:
Closing in favor of #135

PR #136, opened 2016-02-26
CURATOR-262: Fixed integer overflow issue with sleepMs

ddyankov:
https://issues.apache.org/jira/browse/CURATOR-262


cammckenzie:
Looks OK to me. It does change a public interface, but I doubt that many people are using custom SleepingRetry.

PR #137, opened 2016-03-02
Support EPHEMERAL_SEQUENTIAL nodes for ServiceDiscovery instances

nirdothan:
Added a new DYNAMIC_SEQUENTIAL  ServiceType for ServiceDiscovery instances, which creates EPHEMERAL_SEQUENTIAL nodes in Zookeeper. Before this patch only EPHEMERAL or PERSISTENT nodes were supported by Service Discovery.  
[Jira issue](https://issues.apache.org/jira/browse/CURATOR-304)

### DIFF
--- a/curator-x-discovery-server/src/main/java/org/apache/curator/x/discovery/server/rest/DiscoveryResource.java
+++ b/curator-x-discovery-server/src/main/java/org/apache/curator/x/discovery/server/rest/DiscoveryResource.java
@@ -90,7 +90,7 @@ public abstract class DiscoveryResource<T>
             return Response.status(Response.Status.BAD_REQUEST).build();
         }
         
-        if ( instance.getServiceType() == ServiceType.DYNAMIC )
+        if ( instance.getServiceType() == ServiceType.DYNAMIC || instance.getServiceType() == ServiceType.DYNAMIC_SEQUENTIAL )
         {
             log.info("Service type cannot be dynamic");
             return Response.status(Response.Status.BAD_REQUEST).build();
--- a/curator-x-discovery-server/src/main/java/org/apache/curator/x/discovery/server/rest/DiscoveryResource.java
+++ b/curator-x-discovery-server/src/main/java/org/apache/curator/x/discovery/server/rest/DiscoveryResource.java
@@ -90,7 +90,7 @@ public abstract class DiscoveryResource<T>
             return Response.status(Response.Status.BAD_REQUEST).build();
         }
         
-        if ( instance.getServiceType() == ServiceType.DYNAMIC || instance.getServiceType() == ServiceType.DYNAMIC_SEQUENTIAL )
+        if ( instance.getServiceType().isDynamic() )
         {
             log.info("Service type cannot be dynamic");
             return Response.status(Response.Status.BAD_REQUEST).build();
--- a/curator-x-discovery/src/main/java/org/apache/curator/x/discovery/details/ServiceDiscoveryImpl.java
+++ b/curator-x-discovery/src/main/java/org/apache/curator/x/discovery/details/ServiceDiscoveryImpl.java
@@ -220,7 +220,17 @@ public class ServiceDiscoveryImpl<T> implements ServiceDiscovery<T>
         {
             try
             {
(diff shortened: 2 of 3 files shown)


Randgalt:
What's up with "Copyright 2016, Amdocs Development Limited"?

nirdothan:
Legal department insisted. Please let me know if there's another way to indicate this.

Randgalt:
You should review the Apache docs. You must assign rights to Apache when submitting.

nirdothan:
Will resolve that. Otherwise are we OK?

Randgalt:
I'll keep reviewing.

nirdothan:
Copyright statement removed

curator-x-discovery/src/main/java/org/apache/curator/x/discovery/details/ServiceDiscoveryImpl.java
```diff
@@ -220,7 +220,17 @@ protected void internalRegisterService(ServiceInstance<T> service) throws Except
...
+            	CreateMode mode;
+            	switch (service.getServiceType()){
+            	case DYNAMIC:
+            		mode =  CreateMode.EPHEMERAL;
+            		break; 
+            	case  DYNAMIC_SEQUENTIAL:
+            		mode =  CreateMode.EPHEMERAL_SEQUENTIAL;
+            		break; 
+            	default:
+            		mode = CreateMode.PERSISTENT;
```
Randgalt: needs a <code>break;</code>

Randgalt: missing <code>break</code>

Randgalt:
Please update docs at:

* curator-x-discovery-server/README.txt
* curator-x-discovery-server/src/site/confluence/index.confluence

Randgalt:
DiscoveryResource.java has a check for <code>ServiceType.DYNAMIC</code>. I believe the new type must be checked as well, right?

nirdothan:
- curator-x-discovery-server/README.txt

`  Either "STATIC" or "PERMANENT". STATIC will get purged
                                            after the defined threshold has elapsed. PERMANENT must be
                                            manually purged.`

and

- curator-x-discovery-server/src/site/confluence/index.confluence

`*Description:* {name} is the service name, {id} is the instance id. The request entity is a _ServiceInstance_. This
method registers a service instance. If the ServiceType is STATIC, the instance is registered only for the pre\-defined period
(defined in the DiscoveryContext). STATIC services must call putService at least once per period. PERMANENT services are registered
until they are manually deleted.`

Only mention STATIC and PERMANENT, as those are the only allowed service types. DYNAMIC isn't mentioned, so I guess there's no reason to mention DYNAMIC_SEQUENTIAL.

The break and the validation check have been added

curator-x-discovery-server/src/main/java/org/apache/curator/x/discovery/server/rest/DiscoveryResource.java
```diff
@@ -90,7 +90,7 @@ public Response     putService(ServiceInstance<T> instance, @PathParam("name") S
             return Response.status(Response.Status.BAD_REQUEST).build();
         }
         
-        if ( instance.getServiceType() == ServiceType.DYNAMIC )
+        if ( instance.getServiceType() == ServiceType.DYNAMIC || instance.getServiceType() == ServiceType.DYNAMIC_SEQUENTIAL )
```
madrob: Maybe we can add an `.isDynamic()` method to `ServiceType` and call that instead of checking against these two specific cases?

curator-x-discovery/src/main/java/org/apache/curator/x/discovery/details/ServiceDiscoveryImpl.java
```diff
@@ -220,7 +220,18 @@ protected void internalRegisterService(ServiceInstance<T> service) throws Except
         {
             try
             {
-                CreateMode mode = (service.getServiceType() == ServiceType.DYNAMIC) ? CreateMode.EPHEMERAL : CreateMode.PERSISTENT;
+            	CreateMode mode;
+            	switch (service.getServiceType()){
```
madrob: nit: whitespace

PR #138, opened 2016-03-02
[CURATOR-305] Async multi calls not returning error code

Randgalt:
CuratorTransactionImpl.makeCuratorResult() needs to check for ErrorResult and return it in CuratorTransactionResult (new field probably needed).


cammckenzie:
Looks good to me.

PR #139, opened 2016-03-07
[CURATOR-306] Background operations would benefit from an optional error handler

Randgalt:


cammckenzie:
Looks good to me. It does make me appreciate how much duplicate code there is in all of the builder classes though. We should probably look at rationalising that at some point.

Randgalt:
Yeah - I agree. Maybe we can start a discussion on Curator 4.0 and open it up to major refactoring.

PR #140, opened 2016-03-29
[CURATOR-310] Race in PersistentNode startup

Randgalt:
This is a semi-punt but a reasonable one. The simple solution is for the client to call waitForInitialCreate() before calling setData(). So, now, setData() throws an exception if the initial create has not occurred.

### DIFF
--- a/curator-recipes/src/main/java/org/apache/curator/framework/recipes/nodes/PersistentNode.java
+++ b/curator-recipes/src/main/java/org/apache/curator/framework/recipes/nodes/PersistentNode.java
@@ -309,7 +309,10 @@ public class PersistentNode implements Closeable
     }
 
     /**
-     * Set data that node should set in ZK also writes the data to the node
+     * Set data that node should set in ZK also writes the data to the node. NOTE: it
+     * is an error to call this method after {@link #start()} but before the initial create
+     * has completed. Use {@link #waitForInitialCreate(long, TimeUnit)} to ensure initial
+     * creation.
      *
      * @param data new data value
      * @throws Exception errors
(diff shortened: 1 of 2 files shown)


curator-recipes/src/main/java/org/apache/curator/framework/recipes/nodes/PersistentNode.java
```diff
@@ -317,6 +317,7 @@ public String getActualPath()
     public void setData(byte[] data) throws Exception
     {
         data = Preconditions.checkNotNull(data, "data cannot be null");
+        Preconditions.checkState(nodePath.get() != null, "initial create has not been processed. Call waitForInitialCreate() to ensure.");
         this.data.set(Arrays.copyOf(data, data.length));
```
cammckenzie: Isn't this potentially going to break legacy clients depending on how they handle exceptions thrown from this method?

Randgalt: The current behavior is undefined. I can't imagine anyone is dependent on an NPE. Right?

cammckenzie: Yes, good point

cammckenzie:
Looks like an OK short fix to me. We probably need to document the new contract (call waitForInitialCreate before calling setData) somewhere?

Randgalt:
I updated the method doc.

cammckenzie:
Looks OK to me.

fiksn:
Wouldn't it also make sense to use waitForInitialCreate() in GroupMember's setThisData()? I seem to be getting the same issue and can't directly access the underlying PersistentEphemeralNode.

Randgalt:
@fiksn please create a test and open a new issue given that this one is so old

PR #141, opened 2016-04-05
[CURATOR-314] Listenable support for GroupMember recipe

shivramk:
I've added support for listenable events to the GroupMember recipe. Its built on the event support provided by the underlying PathChildenCache object being used by GroupMember

### DIFF
--- a/curator-recipes/src/main/java/org/apache/curator/framework/recipes/nodes/GroupData.java
+++ b/curator-recipes/src/main/java/org/apache/curator/framework/recipes/nodes/GroupData.java
@@ -1,0 +1,94 @@
+/**
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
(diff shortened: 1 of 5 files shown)


curator-recipes/src/main/java/org/apache/curator/framework/recipes/nodes/GroupData.java
```diff
@@ -0,0 +1,94 @@
+/**
```
Randgalt: Can't ChildData be re-used here? The only difference is the stat object which doesn't hurt for GroupMember.

curator-recipes/src/main/java/org/apache/curator/framework/recipes/nodes/GroupMember.java
```diff
@@ -75,6 +119,7 @@ public void start()
         try
         {
             cache.start();
+            cache.getListenable().addListener(childrenCacheListener);
```
Randgalt: Why not add a proxy listener getListenable()? I need to think about this. Does this single listener correctly proxy all behaviors from getListenable()?

curator-recipes/src/main/java/org/apache/curator/framework/recipes/nodes/GroupMemberEvent.java
```diff
@@ -0,0 +1,75 @@
...
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package org.apache.curator.framework.recipes.nodes;
+
+/**
+ * POJO that abstracts a change to a group membership
+ */
+public class GroupMemberEvent
```
Randgalt: PathChildrenCacheEvent or TreeCacheEvent can be used, right? I'm leery expanded the number of event objects.

tisonkun:

PR #145, opened 2016-04-21
[CURATOR-299] WatcherRemovalManager must be updated when there's an exception, retry, etc.

Randgalt:
If there's an exeption, retries exceeded, etc., the watcher was still being stored in WatcherRemovalManager thus causing it to be incorrectly removed. This fix removes the watch from the WatcherRemovalmanager on exceptions, retries exceeded, etc.

### DIFF
--- a/curator-framework/src/main/java/org/apache/curator/framework/imps/Backgrounding.java
+++ b/curator-framework/src/main/java/org/apache/curator/framework/imps/Backgrounding.java
@@ -112,10 +112,15 @@ class Backgrounding
         return callback;
     }
 
-    void checkError(Throwable e) throws Exception
+    void checkError(Throwable e, Watching watching) throws Exception
     {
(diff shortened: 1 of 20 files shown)


cammckenzie:
Looks OK to me. A couple of thoughts though. Wouldn't it have been cleaner to overload the constructor for OperationAndData with a constructor that didn't take a Watching reference? There seems to have been a lot of changes just updating the constructor reference and passing in a null Watching reference. Same for Backgrounding.checkError.

Randgalt:
I struggled with this PR. It ended up way more complicated than I wanted but it's difficult to simplify. 

> Wouldn't it have been cleaner to overload the constructor for OperationAndData with a constructor that didn't take a Watching reference

I don't generally like disjointed ctors. It's harder to use correctly.

PR #148, opened 2016-05-19
[CURATOR-324] Test for event ordering in both caches

Randgalt:
@dragonsinth Please pull then modify this branch...

This test randomly creates, updates, deletes nodes and records events as they come through both PathChildrenCache and TreeCache. On every run that I try, TreeCache will occasionally send TreeCacheEvent.Type.NODE_ADDED for the same node. It appears to always happen after that node has been deleted.

### DIFF
--- a/curator-recipes/src/main/java/org/apache/curator/framework/recipes/cache/TreeCache.java
+++ b/curator-recipes/src/main/java/org/apache/curator/framework/recipes/cache/TreeCache.java
@@ -447,11 +447,11 @@ public class TreeCache implements Closeable
                     }
 
                     NodeState oldState = nodeState.getAndSet(NodeState.LIVE);
-                    if ( oldState != NodeState.LIVE )
+                    if ( oldState == NodeState.PENDING || (oldState == NodeState.DEAD && parent == null))
                     {
                         publishEvent(TreeCacheEvent.Type.NODE_ADDED, toPublish);
                     }
-                    else
+                    else if ( oldState == NodeState.LIVE )
                     {
                         if ( oldChildData == null || oldChildData.getStat().getMzxid() != newStat.getMzxid() )
                         {
--- a/curator-recipes/src/test/java/org/apache/curator/framework/recipes/cache/TestEventOrdering.java
+++ b/curator-recipes/src/test/java/org/apache/curator/framework/recipes/cache/TestEventOrdering.java
@@ -12,7 +12,6 @@ import org.apache.zookeeper.KeeperException;
 import org.testng.Assert;
 import org.testng.annotations.Test;
 import java.io.Closeable;
-import java.util.Date;
 import java.util.List;
 import java.util.Random;
 import java.util.concurrent.BlockingQueue;
--- a/curator-recipes/src/test/java/org/apache/curator/framework/recipes/cache/TestEventOrdering.java
+++ b/curator-recipes/src/test/java/org/apache/curator/framework/recipes/cache/TestEventOrdering.java
@@ -1,0 +1,160 @@
+package org.apache.curator.framework.recipes.cache;
+
+import com.google.common.collect.Lists;
+import com.google.common.collect.Queues;
+import org.apache.curator.framework.CuratorFramework;
+import org.apache.curator.framework.CuratorFrameworkFactory;
+import org.apache.curator.retry.RetryOneTime;
+import org.apache.curator.test.BaseClassForTests;
+import org.apache.curator.test.Timing;
+import org.apache.curator.utils.CloseableUtils;
+import org.apache.zookeeper.KeeperException;
+import org.testng.Assert;
+import org.testng.annotations.Test;
+import java.io.Closeable;
+import java.util.Date;
+import java.util.List;
+import java.util.Random;
+import java.util.concurrent.BlockingQueue;
+import java.util.concurrent.Callable;
+import java.util.concurrent.CountDownLatch;
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.Executors;
+import java.util.concurrent.TimeUnit;
+
+public abstract class TestEventOrdering<T extends Closeable> extends BaseClassForTests
+{
+    private final Timing timing = new Timing();
+    private final long start = System.currentTimeMillis();
+    private static final int THREAD_QTY = 100;
+    private static final int ITERATIONS = 100;
+    private static final int NODE_QTY = 10;
+
+    public enum EventType
+    {
+        ADDED,
+        DELETED
+    }
+
+    public static class Event
+    {
+        public final EventType eventType;
+        public final String path;
+        public final long time = System.currentTimeMillis();
+
+        public Event(EventType eventType, String path)
+        {
+            this.eventType = eventType;
+            this.path = path;
+        }
+    }
+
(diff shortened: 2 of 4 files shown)


Randgalt:
I did some further debugging and the problem has something to do with the same node being deleted/updated/recreated in quick succession. I'm concerned that the node states are handle by individual TreeNode instances. Look at this log (filtered for node "/root/0"):

```
2133 processResult: @Node1 CuratorEventImpl{type=GET_DATA, resultCode=0, path='/root/0', name='null', children=null, context=null, stat=5135,5153,1463694250431,1463694250434,2,0,0,0,3,0,5135, data=[110, 101, 119], watchedEvent=null, aclList=null} [main-EventThread]
2133 publishEvent: @Node1 /root/0 NODE_ADDED [main-EventThread]
2134 processResult: @Node1 CuratorEventImpl{type=CHILDREN, resultCode=0, path='/root/0', name='null', children=[], context=null, stat=5135,5153,1463694250431,1463694250434,2,0,0,0,3,0,5135
2135 processResult: @root CuratorEventImpl{type=CHILDREN, resultCode=0, path='/root', name='null', children=[3, 0, 7, 9, 8], context=null, stat=2,2,1463694248754,1463694248754,0,1729,0,0,9,5,5154
2136 process: @Node1 WatchedEvent state:SyncConnected type:NodeDataChanged path:/root/0 [main-EventThread]
2136 process: @Node1 WatchedEvent state:SyncConnected type:NodeDeleted path:/root/0 [main-EventThread]
2136 publishEvent: @Node1 /root/0 NODE_REMOVED [main-EventThread]
2154 processResult: @root CuratorEventImpl{type=CHILDREN, resultCode=0, path='/root', name='null', children=[0, 6, 5, 4, 8], context=null, stat=2,2,1463694248754,1463694248754,0,1765,0,0,9,5,5270
2155 processResult: @Node1 CuratorEventImpl{type=GET_DATA, resultCode=0, path='/root/0', name='null', children=null, context=null, stat=5268,5268,1463694250454,1463694250454,0,0,0,0,9,0,5268
2155 publishEvent: @Node1 /root/0 NODE_ADDED [main-EventThread]
2167 processResult: @Node2 CuratorEventImpl{type=GET_DATA, resultCode=-101, path='/root/0', name='null', children=null, context=null, stat=null, data=null, watchedEvent=null, aclList=null} [main-EventThread]
2167 wasDeleted @Node2 /root/0 PENDING [main-EventThread]
2167 processResult: @Node2 CuratorEventImpl{type=CHILDREN, resultCode=-101, path='/root/0', name='null', children=[], context=null, stat=null, data=null, watchedEvent=null, aclList=null} [main-EventThread]
2167 wasDeleted @Node2 /root/0 DEAD [main-EventThread]
2183 processResult: @root CuratorEventImpl{type=CHILDREN, resultCode=0, path='/root', name='null', children=[3, 7, 5, 9, 8], context=null, stat=2,2,1463694248754,1463694248754,0,1805,0,0,9,5,5380
2233 processResult: @root CuratorEventImpl{type=CHILDREN, resultCode=0, path='/root', name='null', children=[0, 7, 5, 4], context=null, stat=2,2,1463694248754,1463694248754,0,1842,0,0,9,4,5484
2259 processResult: @Node3 CuratorEventImpl{type=GET_DATA, resultCode=-101, path='/root/0', name='null', children=null, context=null, stat=null, data=null, watchedEvent=null, aclList=null} [main-EventThread]
2259 wasDeleted @Node3 /root/0 PENDING [main-EventThread]
2259 processResult: @Node3 CuratorEventImpl{type=CHILDREN, resultCode=-101, path='/root/0', name='null', children=[], context=null, stat=null, data=null, watchedEvent=null, aclList=null} [main-EventThread]
2260 wasDeleted @Node3 /root/0 DEAD [main-EventThread]
2261 processResult: @root CuratorEventImpl{type=CHILDREN, resultCode=0, path='/root', name='null', children=[7, 6, 9], context=null, stat=2,2,1463694248754,1463694248754,0,1875,0,0,9,3,5604
2273 processResult: @root CuratorEventImpl{type=CHILDREN, resultCode=0, path='/root', name='null', children=[1, 0, 5, 9], context=null, stat=2,2,1463694248754,1463694248754,0,1906,0,0,9,4,5706
2282 processResult: @Node4 CuratorEventImpl{type=GET_DATA, resultCode=0, path='/root/0', name='null', children=null, context=null, stat=5797,5811,1463694250589,1463694250591,1,0,0,0,3,0,5797
2282 publishEvent:@Node4 /root/0 NODE_ADDED [main-EventThread]
```

* The first column is time
* The second column is the operation
* The third is the TreeNode doing the operation (I made the name human readable) 
* The last bit is the event - wasDeleted means that wasDeleted() did not publish NODE_REMOVED because the node state was not LIVE

The problem here is the two NODE_ADDED publishes at 2155 and 2282. I believe there is a missing NODE_REMOVED event at 2260.

Note, at time 2136 we get watcher process() NodeDataChanged immediately followed by a NodeDeleted on @Node1's watcher. 
@Node1 then publishes NODE_REMOVED (correctly) at 2136.

At 2154, @root gets the result of calling getChildren. I assume that "/root/0" was recreated and @root's watcher got notified. Note that
"0" is in the children list. So, this will cause a new node to get created and its wasCreated() called. At 2155, the old @Node1 gets the
result of GET_DATA. This causes a false NODE_ADDED to get generated. @Node1 is officially dead and should not be reporting this.

I believe the sequence of events is: Just prior to 2136, 1 test code thread changes "/root/0" and another deletes it and another recreates it almost immediately.
@Node1's watcher is triggered for NodeDataChanged and begins an async call to getData(), it then receives NodeDeleted BEFORE the async getData() occurs on the server.
The node then gets recreated. The getData() finally executes on the server and the result is returned such that @Node1 misses the node deletion.

Anyway, I have a potential fix for now. In processResults() GET_DATA case, add:

```
if ( nodeState.get() == NodeState.DEAD )
{
    break;
}
```

This fixes the problem. Longer term, we should consider moving node management into the main TreeCache instance and not the individual node objects.

dragonsinth:
Thanks @Randgalt I'll take a look

dragonsinth:
Looks like the naive fix will make testDeleteThenCreateRoot() fail.  I need to dig into this a bit more.

dragonsinth:
I think I may have a fix...

dragonsinth:
Pushed a fix that should pass all the tests.

Randgalt:
For some reason your change isn't sync'ing here. Can you explain this line of code?

```
if ( oldState == NodeState.PENDING || (oldState == NodeState.DEAD && parent == null))
```

Maybe a comment to explain it. I find it confusing.

dragonsinth:
Weird, I don't know what's up with the branch syncing.

Basically, the root node transitions between LIVE <-> DEAD depending on whether it exists or not.  Child nodes can only transition PENDING -> LIVE -> DEAD; a child node isn't allowed to go from DEAD -> LIVE; instead it's replaced with a new child node at the same spot.  That's the source of the double event, I think, although I didn't dig deep enough to fully understand it.  So this test is saying, if you're going PENDING -> LIVE then fire an ADD event; if you're the root then DEAD -> LIVE is allowed to fire an ADD event too.

PR #149, opened 2016-05-20
[CURATOR-308] Parent container deletions could cause a hang

Randgalt:
Parent container deletions could cause a hang as the watcher would never fire. If parent containers get deleted now, reset and recreate them


cammckenzie:
Looks good to me.

PR #150, opened 2016-05-20
[CURATOR-327] -  NamespaceImpl does not report full error message to caller

rmauge:


Randgalt:
Is there a Jira issue for this? Please name is PR "[CURATOR-NNN] - Fixes incomplete error message sent to caller"

rmauge:
My apologies, I should have read the commit protocol beforehand. I have created the JIRA issue and updated the title.

curator-framework/src/main/java/org/apache/curator/framework/imps/NamespaceImpl.java
```diff
@@ -45,7 +45,7 @@
             }
             catch ( IllegalArgumentException e )
             {
-                throw new IllegalArgumentException("Invalid namespace: " + namespace);
+                throw new IllegalArgumentException("Invalid namespace: " + namespace + ", " + e.getMessage());
```
Randgalt: getMessage() can return `null`. Is that OK?

rmauge: Hmm, it general yes of course. But validatePath did not have such a path. I can correct this if needed?

Randgalt: It's not a big deal, it would just print "null".

PR #151, opened 2016-05-21
[CURATOR-326] exists() with createParentContainersIfNeeded was ignoring too many KeeperExceptions

Randgalt:
exists() with createParentContainersIfNeeded was ignoring too many KeeperExceptions and thus hiding connection problems. Should only ignore node exists/not-exists


cammckenzie:
Looks good to me.

PR #153, opened 2016-05-24
CURATOR-329 - SharedValue does not update to current value if zookeeper node version is 0

ARascius:
Fixed SharedValue class to correctly update its value on start() call to value stored in Zookeeper node if node version is zero.


cammckenzie:
Thanks for the patch!

PR #154, opened 2016-05-26
[CURATOR-332] Internal watcher tracking for new WatcherRemoveCuratorFramework feature is not correct

Randgalt:
Reworked WatcherRemovalManager. It now stores watchers only on successful operations. This is more like how ZK does it.

Also, exists watcher must be stored when there is a NoNode result.

### DIFF
--- a/curator-framework/src/main/java/org/apache/curator/framework/imps/EnsembleTracker.java
+++ b/curator-framework/src/main/java/org/apache/curator/framework/imps/EnsembleTracker.java
@@ -43,6 +43,7 @@ import java.io.ByteArrayInputStream;
 import java.io.Closeable;
 import java.util.Arrays;
 import java.util.Properties;
+import java.util.concurrent.atomic.AtomicInteger;
 import java.util.concurrent.atomic.AtomicReference;
 
 @VisibleForTesting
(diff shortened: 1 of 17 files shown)


curator-framework/src/main/java/org/apache/curator/framework/imps/EnsembleTracker.java
```diff
@@ -121,22 +123,38 @@ public QuorumVerifier getCurrentConfig()
...
+            outstanding.incrementAndGet();
+            try
+            {
+                client.getConfig().usingWatcher(this).inBackground(backgroundCallback).forEnsemble();
+            }
+            catch ( Exception e )
+            {
+                outstanding.decrementAndGet();
+                throw e;
+            }
```
dragonsinth: you could also formulate this as:

```
           outstanding.incrementAndGet();
            try
            {
                client.getConfig().usingWatcher(this).inBackground(backgroundCallback).forEnsemble();
                outstanding.incrementAndGet();
            }
            finally
            {
                outstanding.decrementAndGet();
            }
```

dragonsinth:
not familiar enough with this code to really vet, but the idea seems right

curator-framework/src/main/java/org/apache/curator/framework/imps/EnsembleTracker.java
Randgalt: Yeah, that's safer.

PR #157, opened 2016-06-14
Long session timeout issue

cammckenzie:
Fix for issue where LOST events were created at 4/3 of the session timeout rather than at the end of session timeout.

### DIFF
--- a/curator-framework/src/main/java/org/apache/curator/framework/state/ConnectionStateManager.java
+++ b/curator-framework/src/main/java/org/apache/curator/framework/state/ConnectionStateManager.java
@@ -252,7 +252,9 @@ public class ConnectionStateManager implements Closeable
             {
                 int lastNegotiatedSessionTimeoutMs = client.getZookeeperClient().getLastNegotiatedSessionTimeoutMs();
                 int useSessionTimeoutMs = (lastNegotiatedSessionTimeoutMs > 0) ? lastNegotiatedSessionTimeoutMs : sessionTimeoutMs;
-                int pollMaxMs = (useSessionTimeoutMs * 2) / 3; // 2/3 of session timeout
+                long elapsedMs = startOfSuspendedEpoch == 0 ? 0 : System.currentTimeMillis() - startOfSuspendedEpoch;
+                long pollMaxMs = useSessionTimeoutMs - elapsedMs;
+
                 final ConnectionState newState = eventQueue.poll(pollMaxMs, TimeUnit.MILLISECONDS);
                 if ( newState != null )
                 {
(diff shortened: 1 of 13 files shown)


curator-framework/src/main/java/org/apache/curator/framework/state/ConnectionStateManager.java
```diff
@@ -252,7 +252,9 @@ private void processEvents()
             {
                 int lastNegotiatedSessionTimeoutMs = client.getZookeeperClient().getLastNegotiatedSessionTimeoutMs();
                 int useSessionTimeoutMs = (lastNegotiatedSessionTimeoutMs > 0) ? lastNegotiatedSessionTimeoutMs : sessionTimeoutMs;
-                int pollMaxMs = (useSessionTimeoutMs * 2) / 3; // 2/3 of session timeout
+                long elapsedMs = startOfSuspendedEpoch == 0 ? 0 : System.currentTimeMillis() - startOfSuspendedEpoch;
+                long pollMaxMs = useSessionTimeoutMs - elapsedMs;
```
Randgalt: We should probably add a 50% factor here or something. It seems suspicious to wait the entire session time if there isn't a suspended state.

curator-recipes/src/test/java/org/apache/curator/framework/recipes/locks/TestInterProcessSemaphore.java
```diff
@@ -148,10 +148,9 @@ else if ( newState == ConnectionState.RECONNECTED  )
                     }
                 }
             });
-
-            timing.sleepABit();
+      
             server.stop();
-            Assert.assertTrue(timing.awaitLatch(lostLatch));
+            Assert.assertTrue(timing.multiple(2).awaitLatch(lostLatch));
```
Randgalt: Maybe timing.multiple(1.25) or something to see how close we can get.

curator-framework/src/main/java/org/apache/curator/framework/state/ConnectionStateManager.java
cammckenzie: What do you mean if there isn't a suspended state? Shouldn't we go into a suspended state as soon as the connection is dropped? Then, we will wait up to session timeout for an event of any type. If we don't get one, then we assume the session is lost and inject a LOST event.

Randgalt: If startOfSuspendedEpoch is 0, the wait is useSessionTimeoutMs. I'm thinking that if startOfSuspendedEpoch the max wait should (useSessionTimeoutMs / 2) or something. It seems safer to me.

cammckenzie: Ok, sounds reasonable.

cammckenzie:
Ok, I've pushed some additional changes to fix your comments. Are you happy for me to merge?

Randgalt:
LGTM

PR #162, opened 2016-09-06
[CURATOR-348] validateCreate() should be taking a path argument

Randgalt:
validateCreate() should be taking a path argument. The class field "path" was hiding this so I renamed that as well.


cammckenzie:
Looks ok to me

PR #167, opened 2016-10-03
[CURATOR-352] SchemaViolation errors do not contain enough information

Randgalt:
Better exceptions messaging when there is a Schema violation. This is much more useful for debugging purposes


cammckenzie:
Looks ok to me.

PR #168, opened 2016-10-06
CURATOR-354: native memory leak in GzipCompressionProvider

epollan:
explicitly close GZIP input/output streams in GzipCompressionProvider


curator-framework/src/main/java/org/apache/curator/framework/imps/GzipCompressionProvider.java
```diff
@@ -31,8 +31,12 @@
     {
         ByteArrayOutputStream       bytes = new ByteArrayOutputStream();
         GZIPOutputStream            out = new GZIPOutputStream(bytes);
-        out.write(data);
-        out.finish();
+        try {
```
epollan: Looks like contribution guidelines cite java 6 compat, so I didn't use try-with-resources.

cammckenzie:
Thanks for the PR!

PR #171, opened 2016-10-28
[CURATOR-351] Support for TTL Nodes

Randgalt:
Early PR for TTL nodes support so others can test.


curator-framework/src/test/java/org/apache/curator/framework/imps/TestTtlNodes.java
```diff
@@ -0,0 +1,88 @@
...
+
+    @Test
+    public void testBasic() throws Exception
+    {
+        try ( CuratorFramework client = CuratorFrameworkFactory.newClient(server.getConnectString(), new RetryOneTime(1)) )
+        {
+            client.start();
+
+            client.create().withTtl(10).creatingParentsIfNeeded().withMode(CreateMode.PERSISTENT_WITH_TTL).forPath("/a/b/c");
+            Thread.sleep(20);
```
cammckenzie: I assume that client should get a 'removed' event when the node times out and gets removed by ZK? If so, wouldn't it be less error prone to wait on a latch and checking that the node was deleted within 10 +- some fudge factor rather than just sleeping?

curator-framework/src/main/java/org/apache/curator/framework/imps/CreateBuilderImpl.java
```diff
@@ -83,6 +85,13 @@ public CreateBuilderMain orSetData()
         return this;
     }
 
+    @Override
+    public CreateBuilderMain withTtl(long ttl)
```
cammckenzie: I prefer withTTL, but that's just personal preference. I'm not sure what the standard for capitalisation of acronyms is.

curator-framework/src/test/java/org/apache/curator/framework/imps/TestTtlNodes.java
Randgalt: I'll take a look at that.

Randgalt:
FYI - TTL nodes will be in the next release of ZK (3.5.3-beta).

curator-framework/src/test/java/org/apache/curator/framework/imps/TestTtlNodes.java
Randgalt: On reflection, I want to verify that the node is deleted within the TTL so the sleep seems appropriate to me. Right?

Randgalt:
ZooKeeper 3.5.3-beta has been released with TTL support so this PR is now ready.

curator-framework/src/main/java/org/apache/curator/framework/imps/CreateBuilderImpl.java
Randgalt: I've started using strict camel case for everything. It's easier to remember. But, it's not a big deal to me.

curator-framework/src/test/java/org/apache/curator/framework/imps/TestTtlNodes.java
cammckenzie: You're probably right, my concern is that I don't know how quick ZK responds to these TTL expiry events. You've only got a 10ms window here for ZK to remove the node. Maybe that's enough.

Randgalt: I know exactly how long because I wrote it :D Note at the top of the test class: `System.setProperty("znode.container.checkIntervalMs", "1");`

curator-recipes/src/test/java/org/apache/curator/framework/recipes/nodes/TestPersistentTtlNode.java
```diff
@@ -0,0 +1,58 @@
...
+    public void testBasic() throws Exception
+    {
+        try (CuratorFramework client = CuratorFrameworkFactory.newClient(server.getConnectString(), new RetryOneTime(1)))
+        {
+            client.start();
+
+            try (PersistentTtlNode node = new PersistentTtlNode(client, "/test", 10, new byte[0]))
+            {
+                node.start();
+                node.waitForInitialCreate(timing.session(), TimeUnit.MILLISECONDS);
```
cammckenzie: This should probably assert that the return is true?

curator-recipes/src/test/java/org/apache/curator/framework/recipes/nodes/TestPersistentTtlNode.java
```diff
@@ -44,25 +45,49 @@ public void testBasic() throws Exception
         {
             client.start();
 
-            try (PersistentTtlNode node = new PersistentTtlNode(client, "/test", 10, new byte[0]))
+            try (PersistentTtlNode node = new PersistentTtlNode(client, "/test", 100, new byte[0]))
             {
                 node.start();
                 node.waitForInitialCreate(timing.session(), TimeUnit.MILLISECONDS);
```
cammckenzie: This should probably assert that the return is true?

curator-recipes/src/main/java/org/apache/curator/framework/recipes/nodes/PersistentTtlNode.java
```diff
@@ -0,0 +1,208 @@
...
+     * @param initData data for the node
+     */
+    public PersistentTtlNode(CuratorFramework client, String path, long ttlMs, byte[] initData)
+    {
+        this(client, Executors.newSingleThreadScheduledExecutor(ThreadUtils.newThreadFactory("PersistentTtlNode")), path, ttlMs, initData, DEFAULT_CHILD_NODE_NAME, DEFAULT_TOUCH_SCHEDULE_FACTOR);
+    }
+
+    /**
+     * @param client the client
+     * @param path path for the parent ZNode
```
madrob: missing javadoc for executor service. probably worthwhile to caution people that if the executor service is overloaded then it may impact the ability to do ttl updates. we should also give users a way to realize that this is the case when their nodes start expiring. probably not a very common use case though.

curator-framework/src/main/java/org/apache/curator/framework/api/CreateBuilder.java
```diff
@@ -18,8 +18,19 @@
  */
 package org.apache.curator.framework.api;
 
-public interface CreateBuilder extends
-    CreateBuilderMain
+public interface CreateBuilder extends CreateBuilderMain
```
Randgalt: It's because of `orSetData()`

madrob: Right. That's not a problem.

`public interface CreateBuilder extends CreateBuilder2 { CreateBuilder2 orSetData() }`

PR #172, opened 2016-11-01
[CURATOR-349] Trace the session id even when it is disconnected

lvfangmin:
The default session id inside ZooKeeper is 0, we should keep consistent with it in tracker.

### DIFF
--- a/curator-client/src/main/java/org/apache/curator/ConnectionState.java
+++ b/curator-client/src/main/java/org/apache/curator/ConnectionState.java
@@ -163,7 +163,7 @@ class ConnectionState implements Watcher, Closeable
         for ( Watcher parentWatcher : parentWatchers )
         {
 
-            OperationTrace trace = new OperationTrace("connection-state-parent-process", tracer.get());
+            OperationTrace trace = new OperationTrace("connection-state-parent-process", tracer.get(), getSessionId());
             parentWatcher.process(event);
             trace.commit();
         }
@@ -213,16 +213,14 @@ class ConnectionState implements Watcher, Closeable
      * Return the current session id
      */
     public long getSessionId() {
-        long sessionId = -1;
-        if (isConnected()) {
-            try {
-                ZooKeeper zk = zooKeeper.getZooKeeper();
-                if (zk != null) {
-                    sessionId = zk.getSessionId();
-                }
-            } catch (Exception e) {
-                // Ignore the exception
+        long sessionId = 0;
+        try {
+            ZooKeeper zk = zooKeeper.getZooKeeper();
+            if (zk != null) {
+                sessionId = zk.getSessionId();
             }
+        } catch (Exception e) {
+            // Ignore the exception
         }
         return sessionId;
     }


curator-client/src/main/java/org/apache/curator/ConnectionState.java
```diff
@@ -213,16 +213,14 @@ private synchronized void checkTimeouts() throws Exception
...
-        long sessionId = -1;
-        if (isConnected()) {
-            try {
-                ZooKeeper zk = zooKeeper.getZooKeeper();
-                if (zk != null) {
-                    sessionId = zk.getSessionId();
-                }
-            } catch (Exception e) {
-                // Ignore the exception
+        long sessionId = 0;
```
Randgalt: Why did we lose the `isConnected()` check?

lvfangmin: We also want to track the sessionId when the connection timed-out happened, also it would be useful if we can know what's the disconnected session id when the create/setData/getData is called. 

Any concern of removing the isConnected() check?

Randgalt: My concern is that `zooKeeper.getZooKeeper()` can cause a new connection to get started. I'm not sure that that kind of side-effect is expected. If there is no current connection or the connection is in error somehow, `zooKeeper.getZooKeeper()` will call `HandleHolder.closeAndReset()`.

lvfangmin: @Randgalt, here we're calling the HandleHolder.getZookeeper() directly, so it won't call closeAndReset(), please correct me if I missed anything.

lvfangmin:
(Not related with this diff) 
@Randgalt , do we have the schedule of Curator release, are we still blocked by ZOOKEEPER-1525?

The ZOOKEEPER-1525 looks good to me except the new additional parameters concern, if we can refactor that a bit, I'll add a +1 and ask Ben (one of the ZooKeeper committer, also my team member) to accelerate the review process and accept it.

lvfangmin:
Btw, I also like to get this in the next release :)

curator-client/src/main/java/org/apache/curator/ConnectionState.java
Randgalt: Yes, I think you're right. I think I can accept this.

Randgalt:
I'll start working on a release

lvfangmin:
Thanks @Randgalt, please let me know when the new release is published 👍

lvfangmin:
Hi @Randgalt, hope you have a great weekend :) 
Sorry about this, but could you please give me some updates about the release, so I can plan this weeks' work about whether I can start migrating the internal projects to Curator or not.

Randgalt:
@lvfangmin do you subscribe to dev@curator? We released this morning and an email was sent to that list. So, the bits are available now.

lvfangmin:
Ah, just saw it, thanks @Randgalt!

PR #173, opened 2016-11-20
CURATOR-358 - Fixed race condition with getLeader()

cammckenzie:
-If leadership changes between the getParticipantNodes() call and the getLeader() internal call the NoNodeException is now handled and the next child in the list is evaluated.

Another option would be to just return the default empty Participant object and not iterate over the whole list of participants.

### DIFF
--- a/curator-recipes/src/main/java/org/apache/curator/framework/recipes/leader/LeaderSelector.java
+++ b/curator-recipes/src/main/java/org/apache/curator/framework/recipes/leader/LeaderSelector.java
@@ -303,17 +303,14 @@ public class LeaderSelector implements Closeable
         boolean isLeader = true;
         for ( String path : participantNodes )
         {
-            try
+            Participant participant = participantForPath(client, path, isLeader);
+            
+            if( participant != null )
             {
-                Participant participant = participantForPath(client, path, isLeader);
                 builder.add(participant);
-            }
-            catch ( KeeperException.NoNodeException ignore )
-            {
-                // ignore
-            }
 
-            isLeader = false;   // by definition the first node is the leader
+                isLeader = false;   // by definition the first node is the leader
+            }
         }
 
         return builder.build();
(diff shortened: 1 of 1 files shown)


curator-recipes/src/main/java/org/apache/curator/framework/recipes/leader/LeaderSelector.java
```diff
@@ -341,11 +342,41 @@ public Participant getLeader() throws Exception
...
         if ( participantNodes.size() > 0 )
         {
-            return participantForPath(client, participantNodes.iterator().next(), true);
+            Iterator<String> iter = participantNodes.iterator();
+            while ( iter.hasNext() )
+            {
+                
+                try
+                {
+                    result = participantForPath(client, iter.next(), true);
```
dragonsinth: can probably just return here right?

Randgalt: Why not have `participantForPath()` handle the exception and return `null`?

cammckenzie: It's called from other places too, so I'll need to refactor those as well. Probably a better approach though.

dragonsinth: Hehe, I wasn't worried about efficiency, it was a code clarity thought.

cammckenzie: Any other comments?

Randgalt:
LGTM

curator-recipes/src/main/java/org/apache/curator/framework/recipes/leader/LeaderSelector.java
dragonsinth: Nah, looks good.  What I was talking about totally depends on which style you find more readable.  I'm not from the school of single-exit; to me multiple return points is easier to follow than mentally tracking variable state.  I would probably have written this:

```java
static Participant getLeader(CuratorFramework client, Collection<String> participantNodes) throws Exception {
    for (String node : participantNodes) {
        Participant result = participantForPath(client, node, true);
        if ( result != null )  {
            return result;
        }
    }
    return new Participant();
}
```

But what you got works just as well.

cammckenzie: Yep, fair call, it's probably more Java like to just return. I've probably been writing too much C code lately!

PR #174, opened 2016-11-29
Introduced hostname to InstanceSpec class for improved network interface compatibility

IMS94:
Currently, when we create a TestingCluster, all the TestingServers will be bound to the local interface (127.0.0.1).  This becomes a constraint if we want to run a TestingCluster which can be used for testing over a network (say LAN). 

To provide more flexibility over network interfaces, I have added the concept of **hostname** into the **InstanceSpec** class in this PR.


Randgalt:
Please create a corresponding Jira issue for this. Please see: https://cwiki.apache.org/confluence/display/CURATOR/Submitting+Pull+Requests

PR #175, opened 2016-11-30
[CURATOR-360] Allow TestingCluster to listen on other network interfaces

IMS94:
Currently, when we create a TestingCluster, all the TestingServers will be bound to the local interface (127.0.0.1). This becomes a constraint if we want to run a TestingCluster which can be used for testing over a network (say LAN).

To provide more flexibility over network interfaces, I have added the concept of **hostname** into the **InstanceSpec** class in this PR.

### DIFF
--- a/curator-test/src/main/java/org/apache/curator/test/InstanceSpec.java
+++ b/curator-test/src/main/java/org/apache/curator/test/InstanceSpec.java
@@ -72,6 +72,7 @@ public class InstanceSpec
     private final int tickTime;
     private final int maxClientCnxns;
     private final Map<String,Object> customProperties;
+    private final String hostname;
 
     public static InstanceSpec newInstanceSpec()
     {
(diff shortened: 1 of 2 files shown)


cammckenzie:
Looks good to me. I assume there's no easy way of providing unit tests for the new functionality?

IMS94:
Yes. If we are to provide unit tests, we must be aware of the network interfaces in the host machine.I manually tested whether all are fine when I wanted to run a zookeeper cluster to test my clustering implementation through a LAN and it worked fine.

```
        List<InstanceSpec> specs = new ArrayList<>();
        List<Integer> ports = new ArrayList<>();

        int port = 30155, electionPort = 31155, quorumPort = 32155;

        for (int i = 0; i < 3; i++) {
            InstanceSpec spec = new InstanceSpec(null, port, electionPort, quorumPort,
                    true, i, 10000, 100, null, "0.0.0.0");
            logger.info("Zookeeper-{} : port : {}, election port : {}, quorum port : {}", i, port, electionPort, quorumPort);

            specs.add(spec);
            ports.add(port);

            port++;
            electionPort++;
            quorumPort++;
        }

        logger.info("Connect string ports : {}", ports);
        TestingCluster cluster = new TestingCluster(specs);

        try {
            cluster.start();
            Runtime.getRuntime().addShutdownHook(new Thread(() -> {
                try {
                    logger.debug("Stopping cluster");
                    cluster.stop();
                } catch (IOException e) {
                    logger.error("Error when stopping cluster", e);
                }
            }));
        } catch (Exception e) {
            logger.error("Error when starting zookeeper cluster", e);
            throw e;
        }
        logger.info("Cluster started with ports : {}", ports);
```
Here, I had used "0.0.0.0" to open ports on all interfaces. I tried with my IPv4 address assigned by my LAN and confirmed that I can access these ports through network

PR #177, opened 2016-12-15
[CURATOR-364] Persistent Node Constructor Requires a Started ZK Connection

IMS94:
I have moved the 

```
createMode= useProtection ? client.create().creatingParentContainersIfNeeded().withProtection() : client.create().creatingParentContainersIfNeeded();
```
code segment to the **createNode()** method. Then, an exception will be raised only when you call the `start()` method of the PersistentNode without starting the client. This should be the expected behavior I guess?


ilyabiryukov-r7:
Hey guys, are we ok to merge this?

Randgalt:
Note: after merging I noticed that the original PR breaks nodes with protection as the protection ID is stored in the create builder. https://github.com/apache/curator/commit/0b699a368bbb5f3c72ab609fa590bf225c2a2d7a fixes this

PR #179, opened 2016-12-23
[CURATOR-369] Improve log for new configuration events

swallez:
Logging the configuration data only after it has been parsed into a `Properties` object. This allows logging readable text instead of an array of bytes.

Also guarded the log statement with a test on the log level, as `Properties.toString()` is a costly operation.

### DIFF
--- a/curator-framework/src/main/java/org/apache/curator/framework/imps/EnsembleTracker.java
+++ b/curator-framework/src/main/java/org/apache/curator/framework/imps/EnsembleTracker.java
@@ -181,13 +181,10 @@ public class EnsembleTracker implements Closeable, CuratorWatcher
     {
         Properties properties = new Properties();
         properties.load(new ByteArrayInputStream(data));
+        log.info("New config event received: {}", properties);
+
         QuorumMaj newConfig = new QuorumMaj(properties);
         currentConfig.set(newConfig);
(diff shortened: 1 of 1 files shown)


curator-framework/src/main/java/org/apache/curator/framework/imps/EnsembleTracker.java
```diff
@@ -179,13 +179,16 @@ public static String configToConnectionString(QuorumVerifier data) throws Except
...
     private void processConfigData(byte[] data) throws Exception
     {
-        log.info("New config event received: " + Arrays.toString(data));
-
         Properties properties = new Properties();
         properties.load(new ByteArrayInputStream(data));
         QuorumMaj newConfig = new QuorumMaj(properties);
         currentConfig.set(newConfig);
 
+        if ( log.isInfoEnabled() )
```
Randgalt: Instead of this check you can do:

```java
log.info("New config event received: {}", properties.toString());
```

or

```java
log.info(String.format("New config event received: %s", properties.toString()));
```

swallez:
Doh. Thanks @Randgalt! I've been using ifXxxxEnabled for so many years that it's still engraved in my fingers. Fixed!

PR #182, opened 2017-01-02
CURATOR-373 Add option to PersistentNode to not overwrite/delete existing node

lakospeter91:
https://issues.apache.org/jira/browse/CURATOR-373


Randgalt:
If I understand this PR correctly, overwriteExisting only applies on the first use, right? If so, I'd like to see this renamed to something that better reflects it's use. `overwriteInitialExisting` or `overwriteStartingExisting` or something.

tisonkun:

PR #184, opened 2017-01-04
[CURATOR-374] TreeCache: reduce memory usage by using AtomicReferenceFieldUpdaters

swallez:
Reduce memory usage by using `AtomicFieldReferenceUpdaters` rather than instanciating 2 or 3 `AtomicReference` in each tree node.

See [CURATOR-374](https://issues.apache.org/jira/browse/CURATOR-374)


dragonsinth:
Merged

PR #186, opened 2017-01-05
[CURATOR-375] - fix thread interruption being reported twice

thecoop:
In some classes, thread interruption was being reported twice - by setting the thread interrupted flag and re-throwing the `InterruptedException` - this made it look like the thread was actually interrupted twice, and generally causes spurious interruption - even if the `InterruptedException` is handled, the thread interruption status is still set, meaning the next async operation that happens will also fail.

Affected classes:
- `ConnectionState` - `closeAndClear` can never throw `InterruptedException` canyway, so remove the check completely
- `PersistentNode` - as mentioned in the docs for `AutoCloseable`, close methods shouldn't throw InterruptedException anyway. The delete operation is guaranteed, so even if `deleteNode` throws an exception, the node will still (probably) be deleted
- `CreateBuilderImpl`/`DeleteBuilderImpl` - the exception is being re-thrown, so don't need to re-set the interrupt status

### DIFF
--- a/curator-framework/src/main/java/org/apache/curator/framework/imps/CreateBuilderImpl.java
+++ b/curator-framework/src/main/java/org/apache/curator/framework/imps/CreateBuilderImpl.java
@@ -485,7 +485,6 @@ class CreateBuilderImpl implements CreateBuilder, BackgroundOperation<PathAndByt
         }
         catch ( Exception e)
         {
-            ThreadUtils.checkInterrupted(e);
             if ( ( e instanceof KeeperException.ConnectionLossException ||
                 !( e instanceof KeeperException )) && protectedId != null )
             {
--- a/curator-recipes/src/test/java/org/apache/curator/framework/recipes/nodes/TestPersistentNode.java
+++ b/curator-recipes/src/test/java/org/apache/curator/framework/recipes/nodes/TestPersistentNode.java
@@ -27,6 +27,8 @@ import org.apache.curator.utils.CloseableUtils;
 import org.apache.zookeeper.CreateMode;
 import org.testng.Assert;
 import org.testng.annotations.Test;
+
+import java.io.IOException;
 import java.util.concurrent.TimeUnit;
 
 public class TestPersistentNode extends BaseClassForTests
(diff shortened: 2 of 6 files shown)


curator-framework/src/main/java/org/apache/curator/framework/imps/CreateBuilderImpl.java
```diff
@@ -485,7 +485,6 @@ private String protectedPathInForeground(String adjustedPath, byte[] data) throw
         }
         catch ( Exception e)
         {
-            ThreadUtils.checkInterrupted(e);
```
cammckenzie: Won't the removal of this call mean that the thread will not be marked as interrupted unless the caller catches the exception and resets the value themselves? This code is called from many recipes and can also be called explicitly by users Curator code. I would think that there's great potential to break code here, unless I'm missing something.

cammckenzie:
Is this fix specifically for the PersistentNode class? It seems like it will potentially break a lot of stuff.

curator-framework/src/main/java/org/apache/curator/framework/imps/CreateBuilderImpl.java
thecoop: No, because the `InterruptedException` is rethrown in every case. Thread interruption status is represented by _either_ the thread interrupt status _or_ an `InterruptedException`, not both. This is mentioned in the Java tutorials at http://docs.oracle.com/javase/tutorial/essential/concurrency/interrupt.html:

> By convention, any method that exits by throwing an InterruptedException clears interrupt status when it does so.

thecoop:
I first came across this issue in the context of `PersistentNode`, but it applies to the other classes in the PR

curator-framework/src/main/java/org/apache/curator/framework/imps/CreateBuilderImpl.java
Randgalt: re-throwing InterruptedException does not reset the Thread interrupted status of the thread.

thecoop: The caller knows precisely _because_ InterruptedException is being thrown - that is how the information is conveyed to the caller. Check http://docs.oracle.com/javase/tutorial/essential/concurrency/interrupt.html:
> By convention, any method that exits by throwing an InterruptedException clears interrupt status when it does so.

Randgalt: Most code does not check for `InterruptedException` and this is the problem. Again, please explain what the issue being fixed is here. Is this merely an academic issue or is there a real problem?

thecoop: I've just pushed a basic test case showing the issue. The thread is interrupted once, but it is exposed to the caller twice - once through close() throwing an exception, and once through whatever next checks the interrupt status (here, I've used Thread.sleep)

thecoop: As a correction to the previous comment, this fixes the issue by exposing the interruption using the interrupt status only, and not throwing an exception, rather than doing both as it is at the moment.

curator-recipes/src/test/java/org/apache/curator/framework/recipes/nodes/TestPersistentNode.java
```diff
@@ -136,4 +138,52 @@ public void testQuickCloseNodeExists() throws Exception
             CloseableUtils.closeQuietly(client);
         }
     }
+
+    @Test
+    public void testInterruption() throws Exception
```
Randgalt: I don't understand what this is testing. What does it matter how many times a thread is interrupted? I feel like this is an academic exercise. Is there a real problem here?

thecoop: This issue is a root cause of some significant problems we've encountered. I'll have another look at this and come up with a more pertinent test case.

Randgalt: We'll need that otherwise I don't see the issue. The thread interrupt status is boolean. No matter how many times you set it it's still the same value.

PR #190, opened 2017-01-08
[CURATOR-200] Shade Guava into Curator

Randgalt:
Shaded Guava into Curator Client (and Curator Test) so that it's no longer an issue for users. This should have been done a long time ago. Unfortunately, `com.google.common.base.Function`, `com.google.common.base.Predicate` and `com.google.common.reflect.TypeToken` are in Curator's public API so those 3 classes are NOT shaded. However, this won't be a problem because there's no reason to think that Guava will change or remove these basic interfaces.

NOTE: I've tested this by creating a temp project that uses Guava 21.0-rc2. When used with Curator 2.11.1 I get a `NoSuchMethodError`. When used with this new shaded version of Curator there are no errors.

### DIFF
--- a/curator-client/pom.xml
+++ b/curator-client/pom.xml
@@ -1,5 +1,4 @@
-<?xml version="1.0" encoding="UTF-8"?>
-<!--
+<?xml version="1.0" encoding="UTF-8"?><!--
     Licensed to the Apache Software Foundation (ASF) under one
     or more contributor license agreements.  See the NOTICE file
     distributed with this work for additional information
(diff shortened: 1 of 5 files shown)


oza:
Thanks for taking this big issue. It must be helpful for Curator's users. Do you think that this change must be done with source-level compatibility? If the answer is positive, it looks good to me.

> However, this won't be a problem because there's no reason to think that Guava will change or remove these basic interfaces.

It's realistic decision, especially for Function and Predicate interfaces. About [TypeToken](https://google.github.io/guava/releases/snapshot/api/docs/com/google/common/reflect/TypeToken.html), however, it is marked with [Beta annotation](https://google.github.io/guava/releases/snapshot/api/docs/com/google/common/annotations/Beta.html). Hence, I think the interface can be changed.

Randgalt:
@oza breaking public APIs is very disruptive. I've tested this with Guava 21 and it works. If Guava ever changes we can address it then.

oza:
@Randgalt sounds good, thanks for the clarification.

Randgalt:
I don't know OSGI so I can't speak to that. But, Guava is relocated in this PR - that's the point of it. Notice the `relocation` tags in the pom file.

kscarr73:
Ahh, I looked for them, but didn't see them before my comment.  That works then.  Not sure how much bigger the jar is, yet, but I see about doing a build later.

Randgalt:
FYI - the JAR grows by about 1.9M

oza:
LGTM from license's point of view. Guava doesn't have its own NOTICE file, so we don't need to add the file to our jar.

PR #193, opened 2017-01-10
[CURATOR-311] SharedValue could hold stall data when quourm membership changes

oza:
Adding error handling to refresh stale data when stateChanged is called.

### DIFF
--- a/curator-recipes/src/main/java/org/apache/curator/framework/recipes/shared/SharedValue.java
+++ b/curator-recipes/src/main/java/org/apache/curator/framework/recipes/shared/SharedValue.java
@@ -19,6 +19,7 @@
 
 package org.apache.curator.framework.recipes.shared;
 
+import com.google.common.annotations.VisibleForTesting;
 import com.google.common.base.Function;
 import com.google.common.base.Preconditions;
 import org.apache.curator.framework.CuratorFramework;
@@ -56,8 +57,9 @@ public class SharedValue implements Closeable, SharedValueReader
     private final byte[] seedValue;
     private final AtomicReference<State> state = new AtomicReference<State>(State.LATENT);
     private final AtomicReference<VersionedValue<byte[]>> currentValue;
+    private final CuratorWatcher watcher;
 
-    private final CuratorWatcher watcher = new CuratorWatcher()
+    private class SharedValueCuratorWatcher implements CuratorWatcher
     {
         @Override
         public void process(WatchedEvent event) throws Exception
@@ -76,6 +78,17 @@ public class SharedValue implements Closeable, SharedValueReader
         public void stateChanged(CuratorFramework client, ConnectionState newState)
         {
             notifyListenerOfStateChanged(newState);
+            if ( newState == ConnectionState.RECONNECTED )
+            {
+                try
+                {
+                    readValueAndNotifyListenersInBackground();
+                }
+                catch ( Exception e )
+                {
+                    log.error("Could not read value after reconnect", e);
+                }
+            }
         }
     };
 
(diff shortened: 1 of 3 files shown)


curator-recipes/src/main/java/org/apache/curator/framework/recipes/shared/SharedValue.java
```diff
@@ -75,14 +75,41 @@ public void process(WatchedEvent event) throws Exception
...
+
+            case RECONNECTED:
+                // update currentValue
+                try {
+                    state.compareAndSet(State.STALE, State.STARTING);
+                    startInternal();
+                    state.compareAndSet(State.STARTING, State.STARTED);
+                } catch (Exception e) {
+                    ThreadUtils.checkInterrupted(e);
+                    log.error("Could not re-start instances after reconnection", e);
```
oza: I would like to know whether this error handling is correct. Please let me know if you know a better way.

Randgalt:
I'm -1 on this PR overall. The purpose of `state` internally is not to hold the connection state. It's to hold the started state of the instanced. Mixing the two is bound to cause problems. 

A simpler solution would be:

```java
    private final ConnectionStateListener connectionStateListener = new ConnectionStateListener()
    {
        @Override
        public void stateChanged(CuratorFramework client, ConnectionState newState)
        {
            notifyListenerOfStateChanged(newState);
            if ( newState == ConnectionState.RECONNECTED )
            {
                try
                {
                    readValueAndNotifyListenersInBackground();
                }
                catch ( Exception e )
                {
                    log.error("Could not read value after reconnect", e);
                }
            }
        }
    };
```

curator-recipes/src/test/java/org/apache/curator/framework/recipes/shared/TestSharedCount.java
```diff
@@ -389,4 +390,93 @@ public void processResult(CuratorFramework client, CuratorEvent event) throws Ex
             CloseableUtils.closeQuietly(curatorFramework);
         }
     }
+
+    @Test
+    public void testDisconnectReconnectEventDoesNotFireValueWatcherWithMultipleClients() throws Exception
```
Randgalt: I don't think this tests the problem. I tried this test with an unchanged version SharedValue and it passes.

oza:
Thanks Jordan for taking a look. I will double check my test case and resend fixed version later.

oza:
Based on the report in jira, I will create a test case, TestSharedValueWithCluster.java, with TestingCluster to reproduce the problem. Please let me know if you know a better way.

oza:
@Randgalt I could create a test case to reproduce the problem by injecting emulated faulty watcher instead of using pseudo cluster. Could you check it?

curator-recipes/src/test/java/org/apache/curator/framework/recipes/shared/TestSharedCount.java
```diff
@@ -381,12 +385,118 @@ public void processResult(CuratorFramework client, CuratorEvent event) throws Ex
             }).forPath("/count");
             flushDone.await(5, TimeUnit.SECONDS);
 
-            Assert.assertEquals(2, numChangeEvents.get());
+            Assert.assertEquals(numChangeEvents.get(), 3);
```
oza: This means this PR includes semantic change of countHasChanged callback: after reconnecting, countHasChanged() is called.

Randgalt: This commit is no longer available. Can you post it again?

oza: Sure.

oza:
@Randgalt can you see changes?

Randgalt:
No - the commit is still missing.

oza:
Hmm, I don't know what is happening. I would like to close this and try to create new PR.

PR #194, opened 2017-01-11
[CURATOR-379] unfixForNamespace corrupts child path when it contains namespace

computerlove:
When a child path starts with the namespace unfixForNamespace removes the namespace string, thus creating a invalid path.
With namespace "foo" the result of unfixForNamespace("/foobar") is "bar".


tisonkun:

PR #195, opened 2017-01-12
[CURATOR-171] LeaderLatch isn't aware if it's own ephemeral node goes away

oza:
The root cause of the problem reported on CURATOR-171 is that LeaderLatch is not aware of losing self znode after acquire lock. This PR fixes that LeaderLatch is aware of my znode losing and marking itself "not leader".


Randgalt:
This has come up before. There's no reason why a leader latch's node should get deleted. The LeaderLatch code already handles connection loss/reconnection so I don't see why this is needed.

oza:
Make sense. Thanks for your clarification.
I would like to close CURATOR-171 as not a problem issue after jira maintenance is finished.

jacky1193610322:
when I want to change leader manually, I can delete the leader node manually, let the others become the leader.

PR #197, opened 2017-01-26
[CURATOR-367] Delay reconnect on session expired

Zoltan-Szekeres:


curator-client/src/main/java/org/apache/curator/ConnectionState.java
```diff
@@ -160,13 +162,33 @@ public void process(WatchedEvent event)
             }
         }
 
+        // only wait during tests
+        assert waitOnExpiredEvent(event.getState());
```
Zoltan-Szekeres: Yes, assuming assertions are enabled in testing. I've added the delay here wait for the reconnect to happen, but since now the reset is called later in the method it might no longer be necessary.

curator-client/src/main/java/org/apache/curator/ConnectionState.java
```diff
@@ -160,13 +162,33 @@ public void process(WatchedEvent event)
...
             OperationTrace trace = new OperationTrace("connection-state-parent-process", tracer.get(), getSessionId());
             parentWatcher.process(event);
             trace.commit();
         }
+
+        if (eventTypeNone) handleState(event.getState());
+    }
+
+    // only for testing
+    private boolean waitOnExpiredEvent(Event.KeeperState currentState)
```
Zoltan-Szekeres: Yes - only so that it can be called form assert.

curator-client/src/main/java/org/apache/curator/ConnectionState.java
cammckenzie: I don't think that it's reasonable to assume that assertions will only be turned on during testing. If you look at something like LeaderSelector, it has specific code in there to support unit testing (the debugLeadershipLatch variable).

If you can cause the problem to occur without this code though, then it should be removed. I had a quick play with it and I couldn't seem to reproduce it without this code though.

Zoltan-Szekeres: I changed using a debug flag instead of assert. I added an accessor class to be able to set the flag in the package private ConnectionState class.

I cannot cause the problem to occur without this code.

cammckenzie:
Thanks for the update, looks OK now I think. I'll give it a bit more of a test early next week and then merge.

Zoltan-Szekeres:
Thanks for the review. I've also added the mockito dependency to the pom.xml, which I've left out of my previous commit. 

Since we're on Curator 2.x, we'd like to push this change to the 2.x version as well. Do I need to create another PR for that?

cammckenzie:
I've run through some more tests and it all looks good.

My only concern is that we're introducing the Mockito just for this test case. Could you move the test case into the curator-client sub package under org.apache.curator. Then no reflection would be required to set the debug flag and we could remove the mockito dependency?

Do you have an opinion on introducing the Mockito dependency @Randgalt?

In regards to the Curator 2.x and 3.x, don't worry about a separate PR, it should be ok to just merge from the Curator 2.x branch into 3.x.

Randgalt:
I'm -1 on introducing a new dependency. Can the PR be re-worked to not use Mockito?

Zoltan-Szekeres:
Thanks again for going through the change.

I've put the test into curator-framework, because the org.apache.curator.framework.state.ConnectionStateManager got into an invalid state, that's why the CuratorFramework.blockUntilConnected didn't return and that's what I want to test.

I'll remove the mockito dependency.

Randgalt:
This is a significant change. I would like amble time to look at it. @cammckenzie please do not merge.

Randgalt:
We need to see a version of this for the CURATOR-3.0 branch. The code has changed a lot there.

Zoltan-Szekeres:
I couldn't reproduce the issue with the CURATOR-3.0 branch. I think because ConnectionStateManager injects an Expired event after the session timeout has elapsed after the suspended event.

cammckenzie:
Any thoughts on this @Randgalt , I'd like to merge before doing a build.

curator-client/src/main/java/org/apache/curator/ConnectionState.java
```diff
@@ -283,12 +300,19 @@ private boolean checkState(Event.KeeperState state, boolean wasConnected)
...
 
-        if ( checkNewConnectionString && zooKeeper.hasNewConnectionString() )
+        return isConnected;
+    }
+
+    private void handleState(Event.KeeperState state)
+    {
+        if (state == Event.KeeperState.Expired)
+        {
+            handleExpiredSession();
```
Randgalt: We've lost functionality here right? If there's a new connection string and an expired session we don't get the new connection string.

cammckenzie: It gets handled in the handleState() method, which is called from process()

Randgalt: It's `handleState()` that I'm concerned about. When `state == Event.KeeperState.Expired` the `zooKeeper.hasNewConnectionString()` isn't called.

cammckenzie: Isn't that true of the existing implementation though? If the state is expired then the checkNewConnectionString flag gets set to false so the handleNewConnectionString() method won't get called.

Randgalt: Oh - I think you're right. The old version set `checkNewConnectionString=false`

PR #198, opened 2017-02-10
[CURATOR-386] Allow listener to be passed in to PersistentNode

akira:
This provides ability to hook into events from PersistentNode when a
node gets created.  PersistentNode would then notify any registered listeners that the 
node was created, along with the path created (useful when protection is used).

### DIFF
--- a/curator-recipes/src/main/java/org/apache/curator/framework/recipes/nodes/PersistentNode.java
+++ b/curator-recipes/src/main/java/org/apache/curator/framework/recipes/nodes/PersistentNode.java
@@ -19,6 +19,7 @@
 
 package org.apache.curator.framework.recipes.nodes;
 
+import com.google.common.base.Function;
 import com.google.common.annotations.VisibleForTesting;
 import com.google.common.base.Preconditions;
 import org.apache.curator.framework.CuratorFramework;
@@ -28,6 +29,7 @@ import org.apache.curator.framework.api.CreateBuilder;
 import org.apache.curator.framework.api.CreateModable;
 import org.apache.curator.framework.api.CuratorEvent;
 import org.apache.curator.framework.api.CuratorWatcher;
+import org.apache.curator.framework.listen.ListenerContainer;
 import org.apache.curator.framework.state.ConnectionState;
 import org.apache.curator.framework.state.ConnectionStateListener;
 import org.apache.curator.utils.PathUtils;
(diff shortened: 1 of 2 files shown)


akira:
Ran tests locally:

```
[INFO] ------------------------------------------------------------------------
[INFO] Reactor Summary:
[INFO]
[INFO] Apache Curator ..................................... SUCCESS [  3.114 s]
[INFO] Curator Testing .................................... SUCCESS [  8.040 s]
[INFO] Curator Client ..................................... SUCCESS [ 37.217 s]
[INFO] Curator Framework .................................. SUCCESS [09:36 min]
[INFO] Curator Recipes .................................... SUCCESS [29:43 min]
[INFO] Curator Service Discovery .......................... SUCCESS [01:26 min]
[INFO] Curator Examples ................................... SUCCESS [  2.376 s]
[INFO] Curator Service Discovery Server ................... SUCCESS [ 10.766 s]
[INFO] Curator RPC Proxy .................................. SUCCESS [ 13.581 s]
[INFO] ------------------------------------------------------------------------
[INFO] BUILD SUCCESS
[INFO] ------------------------------------------------------------------------
```

curator-recipes/src/main/java/org/apache/curator/framework/recipes/nodes/PersistentNode.java
```diff
@@ -117,6 +120,7 @@ public void processResult(CuratorFramework client, CuratorEvent event)
             {
                 //Update is ok, mark initialisation as complete if required.
                 initialisationComplete();
+                notifyListeners();
```
akira: @lvfangmin I looked at this - and it looks to me like this is the callback where the result of node creation comes back (telling us whether it was successful, node already exists, etc). This is in the `processBackgroundCallback` method, and where we know that the result code comes back as `false` for `KeeperException.Code.NODEEXISTS` and `true` for  `KeeperException.Code.OK`.  This means that we just created the node, and seemingly the place to notify listeners.

The callback for updating the node value seems to be `setDataCallback`, which I did not put any code for notifying listeners.

To me this looked like the place to put this, but definitely let me know if you thought of a better place and I can investigate that.

akira:
@lvfangmin I updated the PR based on your feedback - moved the notifyListeners to the `processBackgroundCallback` callback (I squashed the commit with the previous one).

lvfangmin:
Thanks @akira, the new patch looks good to me.

akira:
@lvfangmin thanks, when do you think it will be merged?

lvfangmin:
@akira sorry for the lately reply, I was waiting for another LGTM, the tests look good on my Mac, I'm going to merge the code now.

PR #200, opened 2017-02-24
[CURATOR-311] SharedValue could hold stall data after reconnecting

oza:
This PR introduces to refresh data when client's state becomes ConnectionState.RECONNECTED.

### DIFF
--- a/curator-recipes/src/test/java/org/apache/curator/framework/recipes/shared/TestSharedCount.java
+++ b/curator-recipes/src/test/java/org/apache/curator/framework/recipes/shared/TestSharedCount.java
@@ -25,6 +25,7 @@ import org.apache.curator.framework.CuratorFramework;
 import org.apache.curator.framework.CuratorFrameworkFactory;
 import org.apache.curator.framework.api.BackgroundCallback;
 import org.apache.curator.framework.api.CuratorEvent;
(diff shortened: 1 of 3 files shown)


curator-recipes/src/test/java/org/apache/curator/framework/recipes/shared/TestSharedCount.java
```diff
@@ -381,12 +385,122 @@ public void processResult(CuratorFramework client, CuratorEvent event) throws Ex
             }).forPath("/count");
             flushDone.await(5, TimeUnit.SECONDS);
 
-            Assert.assertEquals(2, numChangeEvents.get());
+            // CURATOR-311: when a Curator client's state became RECONNECTED, countHasChanged method is called back
+            // because the Curator client calls readValueAndNotifyListenersInBackground in SharedValue#ConnectionStateListener#stateChanged.
+            Assert.assertEquals(numChangeEvents.get(), 3);
```
oza: Minor nits: this line also fixes wrong order of the argument of Assert.assertEquals: assertEquals(actual, expected)  is correct one.

To confirm the reason why I changed numChangeEvents' count, please check the comment in the test code.

oza:
@Randgalt could you take a look when you have a time? I'm sorry if I'm missing something in previous PR.

PR #202, opened 2017-03-09
[CURATOR-391] Cannot use only Stat version to determine node change

Randgalt:
Using only Stat version to note node change (for UPDATE event) is not enough. In a split brain if a node delete plus a recreate is missed, the Stat version can appear to be the same. So, when data is available in the cache, compare the data not the stat version

### DIFF
--- a/curator-recipes/src/main/java/org/apache/curator/framework/recipes/cache/PathChildrenCache.java
+++ b/curator-recipes/src/main/java/org/apache/curator/framework/recipes/cache/PathChildrenCache.java
@@ -45,6 +45,7 @@ import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import java.io.Closeable;
 import java.io.IOException;
+import java.util.Arrays;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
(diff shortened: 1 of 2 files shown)


curator-recipes/src/main/java/org/apache/curator/framework/recipes/cache/PathChildrenCache.java
```diff
@@ -694,14 +695,23 @@ private void applyNewData(String fullPath, int resultCode, Stat stat, byte[] byt
...
             }
             updateInitialSet(ZKPaths.getNodeFromPath(fullPath), data);
         }
     }
 
+    private boolean hasChanged(Stat stat, ChildData previousData, ChildData newData)
+    {
+        if ( cacheData )
+        {
+            return !Arrays.equals(previousData.getData(), newData.getData());
```
lvfangmin: If the node is created with the same data, we will also miss the changes. What's the semantic here for CHILD_UPDATED event? Does it means data in this node was changed? If it's being recreated with the same data, should we fire it or not?

Randgalt: It's not possible to do anything else. The only data we have is the version and the data itself. As a general rule, it's not possible to get every event in ZooKeeper.

lvfangmin: why not compare the mtime in the stat besides the version?

Randgalt: That's true actually. Maybe just compare the entire Stat object every time, even if we have the data. I'll push that soon.

Randgalt: Note: TreeCache uses only Mzxid to determine if there's an update. Is that good enough?

dragonsinth: Yeah, that should actually be deterministic.

oza:
Should we also fix in the checking of version in ```clearDataBytes```?

```
    public boolean clearDataBytes(String fullPath, int ifVersion)
    {
        ChildData data = currentData.get(fullPath);
        if ( data != null )
        {
            if ( (ifVersion < 0) || (ifVersion == data.getStat().getVersion()) ) // should we check mzxid here too?
            {
                if ( data.getData() != null )
                {
                    currentData.replace(fullPath, data, new ChildData(data.getPath(), data.getStat(), null));
                }
                return true;
            }
        }
        return false;
    }
```

oza:
Please ignore my previous comment. Rethinking of the point, because I confirmed that clearDataBytes is for clearing data with a specified version for suppressing memory consumption(not for data management).

lvfangmin:
LGTM +1.

PR #205, opened 2017-03-13
[CURATOR-392] Fixing connection string construction for wildcard client addresses

Vile2539:
The client address in Zookeeper can be a wildcard (0.0.0.0), and if this is used by Curator, it will attempt to create a new connection to 0.0.0.0 instead of the correct server address. It is also possible that no client address or port is set, which should be counted as invalid by Curator, since the appropriate information isn't contained in the data sent.

### DIFF
--- a/curator-framework/src/main/java/org/apache/curator/framework/imps/EnsembleTracker.java
+++ b/curator-framework/src/main/java/org/apache/curator/framework/imps/EnsembleTracker.java
@@ -42,8 +42,6 @@ import org.slf4j.LoggerFactory;
 
 import java.io.ByteArrayInputStream;
 import java.io.Closeable;
-import java.net.InetAddress;
-import java.net.InetSocketAddress;
 import java.util.Properties;
 import java.util.concurrent.atomic.AtomicInteger;
 import java.util.concurrent.atomic.AtomicReference;
@@ -175,9 +173,8 @@ public class EnsembleTracker implements Closeable, CuratorWatcher
             {
                 sb.append(",");
             }
-            InetAddress wildcardAddress = new InetSocketAddress(0).getAddress();
             String hostAddress;
-            if ( wildcardAddress.equals(server.clientAddr.getAddress()) )
+            if ( server.clientAddr.getAddress().isAnyLocalAddress() )
             {
                 hostAddress = server.addr.getAddress().getHostAddress();
             }
--- a/curator-framework/src/main/java/org/apache/curator/framework/imps/EnsembleTracker.java
+++ b/curator-framework/src/main/java/org/apache/curator/framework/imps/EnsembleTracker.java
@@ -20,7 +20,6 @@
 package org.apache.curator.framework.imps;
 
 import com.google.common.annotations.VisibleForTesting;
-import com.google.common.base.Objects;
 import com.google.common.base.Preconditions;
 import com.google.common.collect.Maps;
 import org.apache.curator.ensemble.EnsembleProvider;
@@ -40,10 +39,11 @@ import org.apache.zookeeper.server.quorum.flexible.QuorumMaj;
 import org.apache.zookeeper.server.quorum.flexible.QuorumVerifier;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
+
 import java.io.ByteArrayInputStream;
 import java.io.Closeable;
+import java.net.InetAddress;
 import java.net.InetSocketAddress;
-import java.util.Arrays;
 import java.util.Properties;
 import java.util.concurrent.atomic.AtomicInteger;
 import java.util.concurrent.atomic.AtomicReference;
@@ -166,12 +166,26 @@ public class EnsembleTracker implements Closeable, CuratorWatcher
         StringBuilder sb = new StringBuilder();
         for ( QuorumPeer.QuorumServer server : data.getAllMembers().values() )
         {
+            if ( server.clientAddr == null )
+            {
+                // Invalid client address configuration in zoo.cfg
+                continue;
+            }
             if ( sb.length() != 0 )
             {
                 sb.append(",");
             }
-            InetSocketAddress address = Objects.firstNonNull(server.clientAddr, server.addr);
-            sb.append(address.getAddress().getHostAddress()).append(":").append(address.getPort());
+            InetAddress wildcardAddress = new InetSocketAddress(0).getAddress();
+            String hostAddress;
+            if ( wildcardAddress.equals(server.clientAddr.getAddress()) )
+            {
+                hostAddress = server.addr.getAddress().getHostAddress();
+            }
+            else
+            {
+                hostAddress = server.clientAddr.getAddress().getHostAddress();
+            }
+            sb.append(hostAddress).append(":").append(server.clientAddr.getPort());
         }
 
         return sb.toString();
@@ -183,13 +197,19 @@ public class EnsembleTracker implements Closeable, CuratorWatcher
         properties.load(new ByteArrayInputStream(data));
         log.info("New config event received: {}", properties);
 
-        QuorumMaj newConfig = new QuorumMaj(properties);
-        currentConfig.set(newConfig);
-
-        String connectionString = configToConnectionString(newConfig);
-        if ( connectionString.trim().length() > 0 )
+        if (!properties.isEmpty())
         {
-            ensembleProvider.setConnectionString(connectionString);
+            QuorumMaj newConfig = new QuorumMaj(properties);
+            String connectionString = configToConnectionString(newConfig);
+            if (connectionString.trim().length() > 0)
+            {
+                currentConfig.set(newConfig);
+                ensembleProvider.setConnectionString(connectionString);
+            }
+            else
+            {
+                log.error("Invalid config event received: {}", properties);
+            }
         }
         else
         {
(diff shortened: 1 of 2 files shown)


curator-framework/src/main/java/org/apache/curator/framework/imps/EnsembleTracker.java
```diff
@@ -166,12 +166,26 @@ public static String configToConnectionString(QuorumVerifier data) throws Except
         StringBuilder sb = new StringBuilder();
         for ( QuorumPeer.QuorumServer server : data.getAllMembers().values() )
         {
+            if ( server.clientAddr == null )
+            {
+                // Invalid client address configuration in zoo.cfg
+                continue;
```
hanm: The spec of server address is server.<positive id> = <address1>:<port1>:<port2>[:role];[<client port address>:]<client port>, and if server.clientAddr is null then it means the client port address is not present. The client port address is optional, so the address is still a valid address.

curator-framework/src/main/java/org/apache/curator/framework/imps/EnsembleTracker.java
```diff
@@ -166,12 +166,26 @@ public static String configToConnectionString(QuorumVerifier data) throws Except
...
+            }
             if ( sb.length() != 0 )
             {
                 sb.append(",");
             }
-            InetSocketAddress address = Objects.firstNonNull(server.clientAddr, server.addr);
-            sb.append(address.getAddress().getHostAddress()).append(":").append(address.getPort());
+            InetAddress wildcardAddress = new InetSocketAddress(0).getAddress();
```
hanm: I think you can just use server.addr which is guaranteed not null, and then use getHostAddress to get the hostAddress.

Vile2539:
It's been about 4 months, so I'm a bit fuzzy on the exact details.

> The spec of server address is server. = ::[:role];[:], and if server.clientAddr is null then it means the client port address is not present. The client port address is optional, so the address is still a valid address.

The problem is that EnsembleTracker is getting the connection string for itself to Zookeeper. If it receives a config without a valid client port, then it can't create a new connection string to that client. Since Zookeeper isn't sending the clientPort configured outside of that string (in this case), we can't use any of the connection string.

> I think you can just use server.addr which is guaranteed not null, and then use getHostAddress to get the hostAddress.

I'm not entirely sure about that. We want to use the server.clientAddr ideally - unless it's a wildcard, in which case we use the server.addr. It's not about one of them being null, but the clientAddr should be favoured over the server.addr (for example - we might have clientAddr set to localhost to accept only local connections).

curator-framework/src/main/java/org/apache/curator/framework/imps/EnsembleTracker.java
```diff
@@ -166,12 +166,26 @@ public static String configToConnectionString(QuorumVerifier data) throws Except
...
+            String hostAddress;
+            if ( wildcardAddress.equals(server.clientAddr.getAddress()) )
+            {
+                hostAddress = server.addr.getAddress().getHostAddress();
+            }
+            else
+            {
+                hostAddress = server.clientAddr.getAddress().getHostAddress();
+            }
+            sb.append(hostAddress).append(":").append(server.clientAddr.getPort());
```
Randgalt: Isn't calling `getHostAddress()` really expensive? Is this necessary?

curator-framework/src/main/java/org/apache/curator/framework/imps/EnsembleTracker.java
Randgalt: So, @hanm you're saying the original code is correct?

curator-framework/src/main/java/org/apache/curator/framework/imps/EnsembleTracker.java
Vile2539: I don't believe getHostAddress() involves any DNS lookups, hostname resolution, etc. It just gets the String representation of the byte array on the InetAddressHolder. I may be mistaken about this though.

Randgalt: @Vile2539 you're right

curator-framework/src/main/java/org/apache/curator/framework/imps/EnsembleTracker.java
hanm: I checked spec again, and I agree with what @Vile2539 commented - we should try clientAddr first and then fall back if necessary. So the code change in this pull request here looks good to me.

curator-framework/src/main/java/org/apache/curator/framework/imps/EnsembleTracker.java
hanm: I see what you mean - if the client_port_address is not present ZK will not initialize server.clientAddress, which is where you extract the client port. This is not a problem for ZK itself because it has the knowledge of clientPort of a quorum peer which is not publicly accessible.

I think QuorumSever can provide APIs so you guys can use the APIs to get the client / server address / port rather than relying on the clientAddr etc fields. Before that happens it's fine to skip as there is no way to extract the client port (and I think that only happens when clientAddress is not explicitly present in the config string - cases like server.1= 125.23.63.23:1234:1235;1236)

curator-framework/src/test/java/org/apache/curator/framework/imps/TestReconfiguration.java
```diff
@@ -313,6 +313,38 @@ public void testNewMembers() throws Exception
...
+    {
+        String config = "server.1=10.1.2.3:2888:3888:participant;2181";
+        String configString = EnsembleTracker.configToConnectionString(toQuorumVerifier(config.getBytes()));
+        Assert.assertEquals("10.1.2.3:2181", configString);
+    }
+
+    @Test
+    public void testConfigToConnectionStringWildcardClientAddr() throws Exception
+    {
+        String config = "server.1=10.1.2.3:2888:3888:participant;0.0.0.0:2181";
```
lvfangmin: Please also add a test case for IPv6 wildcard:

server.1=[1010:0001:0002:0003:0004:0005:0006:0007]:2888:3888:participant;[::0]:2181

curator-framework/src/main/java/org/apache/curator/framework/imps/EnsembleTracker.java
```diff
@@ -166,12 +166,26 @@ public static String configToConnectionString(QuorumVerifier data) throws Except
...
+            }
             if ( sb.length() != 0 )
             {
                 sb.append(",");
             }
-            InetSocketAddress address = Objects.firstNonNull(server.clientAddr, server.addr);
-            sb.append(address.getAddress().getHostAddress()).append(":").append(address.getPort());
+            InetAddress wildcardAddress = new InetSocketAddress(0).getAddress();
+            String hostAddress;
+            if ( wildcardAddress.equals(server.clientAddr.getAddress()) )
```
lvfangmin: If the clientAddr is in IPv6 format, will this equal? Or we should also check the Inet6Address, would be great if you can add a test case with IPv6 wildcard.

curator-framework/src/test/java/org/apache/curator/framework/imps/TestReconfiguration.java
Vile2539: It doesn't look like Zookeeper supports IPv6 addresses in 3.5.1-alpha (which is what Curator seems to be using). This appears to be fixed in Zookeeper 3.5.2 and 3.6.0:

https://issues.apache.org/jira/browse/ZOOKEEPER-1460

Randgalt: Curator will be at 3.5.3-beta for the next release. That has change since this PR has started. feel free to increment the ZK version in the main pom.

Vile2539: I seem to be getting a NPE with 3.5.3-beta when attempting to construct the TestingZooKeeperServer:

    java.lang.NullPointerException
	    at org.apache.zookeeper.server.quorum.QuorumPeerMain.runFromConfig(QuorumPeerMain.java:158)
	    at org.apache.curator.test.TestingZooKeeperServer$1.run(TestingZooKeeperServer.java:150)
	    at java.lang.Thread.run(Thread.java:745)

It looks like is a known issue, but I don't understand the fix version of 3.4.0 when it occurred on 3.5.3-beta only - I assume that should be 3.5.4-beta:

https://issues.apache.org/jira/browse/CURATOR-409

I'm not really sure how I should proceed here. Updating the pom would break a lot of the tests, and I can't add IPv6 tests without the updated Zookeeper version supporting IPv6.

Randgalt: @Vile2539 CURATOR-3.0 branch has been discontinued. "master" is now what used to be CURATOR-3.0

Vile2539: @Randgalt Ahh right. I've changed the base branch now and I'll have a look into the tests after work.

Vile2539:
@Randgalt Updated pull request base to master.

@lvfangmin Changed the wildcard check to use the [isAnyLocalAddress\(\)](https://docs.oracle.com/javase/7/docs/api/java/net/InetAddress.html#isAnyLocalAddress()) method of InetAddress, which works for IPv4 and IPv6:

> Utility routine to check if the InetAddress in a wildcard address. *(sic)*

Also added additional tests to make sure that both IPv4 and IPv6 wildcards are recognised, that both work together, and that regular IPv6 addresses work.

Let me know if there's anything else you would like added.

Randgalt:
thanks @Vile2539 - I'll run the tests and, assuming they pass, I'll merge. Anyone have any last objections to merging? @hanm or @lvfangmin ?

PR #207, opened 2017-03-16
Fix for CURATOR-233

jld123:
Fix for CURATOR-233.


Randgalt:
Sorry it's taken so long to address this... Can we get a test that shows the original problem with the fix?

eolivelli:
Very useful

tisonkun:
This seems like a duplicate of [CURATOR-536](https://issues.apache.org/jira/browse/CURATOR-536) and should be fixed by https://github.com/apache/curator/pull/322. Closed as completed. Feel free to file another ticket if it's not the case.

PR #208, opened 2017-03-25
[CURATOR-394] UnrecognizedPropertyException: "enabled" incompatibility

Randgalt:
Here is a potential fix for the issue folks. Let me know what you think...

For users who don't want the new "enabled" field serialized, you can create a JsonInstanceSerializer in compatibility mode:

```java
new JsonInstanceSerializer(payloadClass, true)
```

Internally, the JsonInstanceSerializer will not serialize the enabled field. There are tests to ensure that this will deserialize in old clients.

Going forward, JsonInstanceSerializer now sets ignore fields in the ObjectMapper.

### DIFF
--- a/curator-x-discovery/src/main/java/org/apache/curator/x/discovery/details/JsonInstanceSerializer.java
+++ b/curator-x-discovery/src/main/java/org/apache/curator/x/discovery/details/JsonInstanceSerializer.java
@@ -16,12 +16,14 @@
  * specific language governing permissions and limitations
  * under the License.
  */
+
 package org.apache.curator.x.discovery.details;
 
+import com.google.common.annotations.VisibleForTesting;
 import org.apache.curator.x.discovery.ServiceInstance;
+import org.codehaus.jackson.map.DeserializationConfig;
 import org.codehaus.jackson.map.ObjectMapper;
 import org.codehaus.jackson.type.JavaType;
-import java.io.ByteArrayOutputStream;
 
 /**
  * A serializer that uses Jackson to serialize/deserialize as JSON. IMPORTANT: The instance
@@ -29,17 +31,42 @@ import java.io.ByteArrayOutputStream;
  */
 public class JsonInstanceSerializer<T> implements InstanceSerializer<T>
 {
-    private final ObjectMapper      mapper;
-    private final Class<T>          payloadClass;
-    private final JavaType          type;
+    private final ObjectMapper mapper;
+    private final Class<T> payloadClass;
+    private final boolean compatibleSerializationMode;
+    private final JavaType type;
 
     /**
      * @param payloadClass used to validate payloads when deserializing
      */
     public JsonInstanceSerializer(Class<T> payloadClass)
     {
+        this(payloadClass, false, false);
+    }
+
+    /**
+     * CURATOR-275 introduced a new field into {@link org.apache.curator.x.discovery.ServiceInstance}. This caused a potential
+     * {@link org.codehaus.jackson.map.exc.UnrecognizedPropertyException} in older clients that
+     * read newly serialized ServiceInstances. If you are susceptible to this you should set the
+     * serializer to be an instance of {@link org.apache.curator.x.discovery.details.JsonInstanceSerializer}
+     * with <code>compatibleSerializationMode</code> set to true. IMPORTANT: when this is done, the new <code>enabled</code>
+     * field of ServiceInstance is <strong>not</strong> serialized.
+     *
+     * @param payloadClass used to validate payloads when deserializing
+     * @param compatibleSerializationMode pass true to serialize in a manner that supports clients pre-CURATOR-275
+     */
+    public JsonInstanceSerializer(Class<T> payloadClass, boolean compatibleSerializationMode)
+    {
+        this(payloadClass, compatibleSerializationMode, false);
+    }
+
+    @VisibleForTesting
+    JsonInstanceSerializer(Class<T> payloadClass, boolean compatibleSerializationMode, boolean failOnUnknownProperties)
+    {
         this.payloadClass = payloadClass;
+        this.compatibleSerializationMode = compatibleSerializationMode;
         mapper = new ObjectMapper();
+        mapper.configure(DeserializationConfig.Feature.FAIL_ON_UNKNOWN_PROPERTIES, failOnUnknownProperties);
         type = mapper.getTypeFactory().constructType(ServiceInstance.class);
     }
 
@@ -55,8 +82,11 @@ public class JsonInstanceSerializer<T> implements InstanceSerializer<T>
     @Override
     public byte[] serialize(ServiceInstance<T> instance) throws Exception
     {
-        ByteArrayOutputStream           out = new ByteArrayOutputStream();
-        mapper.writeValue(out, instance);
-        return out.toByteArray();
+        if ( compatibleSerializationMode )
+        {
+            OldServiceInstance<T> compatible = new OldServiceInstance<T>(instance.getName(), instance.getId(), instance.getAddress(), instance.getPort(), instance.getSslPort(), instance.getPayload(), instance.getRegistrationTimeUTC(), instance.getServiceType(), instance.getUriSpec());
+            return mapper.writeValueAsBytes(compatible);
+        }
+        return mapper.writeValueAsBytes(instance);
     }
 }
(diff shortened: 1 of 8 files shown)


curator-x-discovery/src/test/java/org/apache/curator/x/discovery/details/NewServiceInstance.java
```diff
@@ -0,0 +1,148 @@
...
+    private final T payload;
+    private final long registrationTimeUTC;
+    private final ServiceType serviceType;
+    private final UriSpec uriSpec;
+    private final boolean enabled;
+    private final String new1;
+    private final Long new2;
+    private final Date new3;
+    private final URI new4;
+
```
cammckenzie: What are these new1-new4 variables? They seem to be a new addition as part of this fix?

Also, Is there a reason that this class can't be a subclass of the OldServiceInstance?

curator-x-discovery/src/main/java/org/apache/curator/x/discovery/details/JsonInstanceSerializer.java
```diff
@@ -16,30 +16,57 @@
...
+    private final Class<T> payloadClass;
+    private final boolean compatibleSerializationMode;
+    private final JavaType type;
 
     /**
      * @param payloadClass used to validate payloads when deserializing
      */
     public JsonInstanceSerializer(Class<T> payloadClass)
     {
+        this(payloadClass, false, false);
```
cammckenzie: Doesn't this approach imply that moving to Curator 2.12.x requires you to change the code base to not break backwards compatibility? Shouldn't the default be to be in compatibility mode? Then you can just upgrade to 2.12.x and everything will still work, but you need to change the code base to pick up the fix for CURATOR-275?

Randgalt: Hmm - that's a good point.

curator-x-discovery/src/test/java/org/apache/curator/x/discovery/details/NewServiceInstance.java
Randgalt: This is a test. I wanted to prove that if we add new fields in the future the same problem won't happen again.

Randgalt:
@cammckenzie I've turned off default serialization of `enabled` and added copious comments about it. Please review.

cammckenzie:
I think that this is OK.

I'm just wondering what the best way forward is after this release? Ideally we don't want to have to have users opt in to pick up CURATOR-275 do we? I've never used the service discovery stuff, so I'm not sure if that's an issue.

Should the next release after the one that includes this have the serialization of enabled set to true? Then we would have an upgrade path from pre 2.12.x to 2.12.x+1 (with this fix) to 2.12.x + 2?

Randgalt:
@cammckenzie TBH - that's why I didn't make compatibleSerializationMode the default originally. I don't have a good answer. Any idea? Maybe we just comment that we'll reverse this in a future release?

cammckenzie:
I can't think of a better option other than deprecating the 1 arg constructor, and forcing clients to explicitly define their behaviour going forward.

Randgalt:
Me neither. Let's let this sit for a bit and see if anyone else has comments.

curator-x-discovery/src/test/java/org/apache/curator/x/discovery/details/NewServiceInstance.java
lvfangmin: These test fields will be removed when landing the code, right?

lvfangmin:
One concern is that this diff changed the default behavior to not passing 'enable', which will break the users who are already using service discovery server and client with the enable field.

If I understand correctly, there are 4 cases:

1. discovery server: w enable,     client: w enable
2. discovery server: w/o enable, client: w/o enable
3. discovery server: w/o enable, client: w enable
4. discovery server: w enable,    client: w/o enable


The first 2 cases we don't need to worry about, the 3rd one will fail because the server cannot recognize the new enable field, the 4th will fail because the client cannot recognize the new field, sounds like we only need to change the code to be able to ignore unknown field, what's the reason of having OldServiceInstance?

curator-x-discovery/src/test/java/org/apache/curator/x/discovery/details/NewServiceInstance.java
Randgalt: Good point - I'll do that

Randgalt:
@lvfangmin the problem as reported is existing clients in a distributed system. That code cannot be updated and the previous versions of Curator's ServiceDiscovery ObjectMapper wasn't set to ignore unknown fields.

lvfangmin:
I see, do we have the client version in the request, if we have it then maybe we can make the server code backward compatible based on the client version.

betalb:
But what about case when service, uses old version (or compat mode) and client has new code?

After de-serialisation enabled flag will be false, and client will be intrepreting this service as disabled, which also is not good

lvfangmin:
@betalb we cannot fully control the version of Curator being used on client side during gradually roll out, but I suppose we we can control the version of Curator service running on the server side, correct me if I'm wrong.

gianm:
@betalb 

> After de-serialisation enabled flag will be false and client will be intrepreting this service as disabled, which also is not good.

IIRC, the deserialization code assumes that if enabled is missing or null, it's treated as "true". So I think there's no issue here.

gianm:
@lvfangmin 

> I see, do we have the client version in the request, if we have it then maybe we can make the server code backward compatible based on the client version.

I think not, since there isn't even really a "request" from the client to the server. The client is just deserializing json written by the server to a znode.

gianm:
@Randgalt 

This patch looks good to me. NB: I have only read it, not tested it.

@cammckenzie 

> Should the next release after the one that includes this have the serialization of enabled set to true? Then we would have an upgrade path from pre 2.12.x to 2.12.x+1 (with this fix) to 2.12.x + 2?

I think change is good, but I would do it in the next 'major' release like 3.x, since it's a breaking change for older clients.

gianm:
Oh wait, there's already a Curator 3.0. Well, whatever the next 'major' release would be in the Curator numbering scheme.

betalb:
@gianm Can you point out where this check is done. I've looked into `JsonInstanceSerializer` and see that plain Jackson `ObjectMapper` and in `ServiceInstance` class enabled flag is primitive boolean, so it will be initialized with `false` upon deserialisation.

gianm:
@betalb, in https://github.com/apache/curator/blob/e7f55f89056f1447cb2ed73a0cdfd66759e11f91/curator-x-discovery/src/main/java/org/apache/curator/x/discovery/ServiceInstance.java, the default constructor sets enabled to true:

```java
    /**
     * Inits to default values. Only exists for deserialization
     */
    ServiceInstance()
    {
        this("", "", null, null, null, null, 0, ServiceType.DYNAMIC, null, true);
    }
```

And if "enabled" is not present in the JSON object then it should stay that way. I'm not sure if there's a unit test for this behavior though, I didn't check that. @Randgalt do you know if there is one?

Randgalt:
You mean a test that proves that "enable" isn't in the JSON? Not explicitly but it's implied by the old serializer tests.

gianm:
I meant a test that proves that deserializing some JSON without "enable" into `ServiceInstance` (the new one) yields an instance that thinks it is enabled. (i.e. enable defaults to true if not provided in the json)

Randgalt:
OK - I added a test for forward compatibility.

gianm:
@Randgalt what's your feeling about how this patch makes it to a release? For Druid were waiting for this in order to upgrade Curator.

Randgalt:
Folks who are happy with the patch as it is please put a 👍 on this message. If anyone is against this patch, please reply.

cammckenzie:
I think that what we've got is probably the best compromise we're going to get.

Randgalt:
So, any last comments before this is merged?

adamdecaf:
Was this supposed to be backported to the 2.x series? We upgraded some clients to 2.12.0 and ran into this. 

https://github.com/apache/curator/blob/apache-curator-2.12.0/curator-x-discovery/src/main/java/org/apache/curator/x/discovery/details/JsonInstanceSerializer.java

Randgalt:
It was marked to go into 2.13.0 but was never released as we moved to a common binary for ZK 3.4 and 3.5. Have you tried 4.0.0?

adamdecaf:
> Have you tried 4.0.0?

Of curator? No. We're actually moving off curator/zk in favor of another service discovery technology. I'm probably going to use the [JsonInstanceSerializer patch](https://github.com/apache/curator/blob/3ca02ea7f076d40333d664dd609962d0fe2e4c5f/curator-x-discovery/src/main/java/org/apache/curator/x/discovery/details/JsonInstanceSerializer.java) in our upgraded clients.

PR #210, opened 2017-04-01
[CURATOR-400] upgrade clirr to fix the check error on java 8

lvfangmin:
Currently, the `mvn compile` failed on CURATOR-3.0 branch, it turns out clirr maven plugin has bug and can fail on Java 8, the latest 2.8 version fixed those bug, and compiled without error.

Error details:
-------------------
[ERROR] Failed to execute goal org.codehaus.mojo:clirr-maven-plugin:2.6.1:check (default) on project curator-x-async: Execution default of goal org.codehaus.mojo:clirr-maven-plugin:2.6.1:check failed: Invalid byte tag in constant pool: 15


Randgalt:
LGTM

PR #212, opened 2017-04-10
Fix guava version incompatibility with MoreExecutors.sameThreadExecutor()

lucasmo:
Before I went to create this PR, I see similar ones have been closed several times, suggesting that it was addressed in https://github.com/apache/curator/pull/190 as part of https://issues.apache.org/jira/browse/CURATOR-200

This doesn't appear to actually address the issue, as this error occurs:

```
java.lang.NoSuchMethodError: com.google.common.util.concurrent.MoreExecutors.sameThreadExecutor()Lcom/google/common/util/concurrent/ListeningExecutorService;
        at org.apache.curator.framework.listen.ListenerContainer.addListener(ListenerContainer.java:40)
        at org.apache.curator.framework.imps.CuratorFrameworkImpl.start(CuratorFrameworkImpl.java:256)
        at XXX.start(XXX.java:88)
```

I'm not familiar with shading, but it appears it's either not implemented or not working properly in this case (does shading rewrite the curator source files?). This error occurs with both 2.12.0 and 3.3.0.

### DIFF
--- a/curator-framework/src/main/java/org/apache/curator/framework/listen/ListenerContainer.java
+++ b/curator-framework/src/main/java/org/apache/curator/framework/listen/ListenerContainer.java
@@ -38,7 +38,7 @@ public class ListenerContainer<T> implements Listenable<T>
     @Override
     public void addListener(T listener)
     {
-        addListener(listener, MoreExecutors.sameThreadExecutor());
+        addListener(listener, MoreExecutors.directExecutor());
     }
 
     @Override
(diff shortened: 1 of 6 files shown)


Randgalt:
I just created this project that shows that Curator works with Guava 21. 

https://github.com/Randgalt/curator-guava-example

What is different about your project?

Randgalt:
You must be using Curator 3.2.0 and not Curator 3.3.0.

LucasMadar:
Looks like you're correct. I also did the same below with 2.12.0 (the version we're using). Thanks for investigating; sorry for the trouble.

```
$ wget http://central.maven.org/maven2/org/apache/curator/curator-framework/3.3.0/curator-framework-3.3.0.jar
$ jar xvf curator-framework-3.3.0.jar
$ javap -c org/apache/curator/framework/listen/ListenerContainer.class

  public void addListener(T);
    Code:
       0: aload_0
       1: aload_1
       2: invokestatic  #50                 // Method org/apache/curator/shaded/com/google/common/util/concurrent/MoreExecutors.sameThreadExecutor:()Lorg/apache/curator/shaded/com/google/common/util/concurrent/ListeningExecutorService;
       5: invokevirtual #53                 // Method addListener:(Ljava/lang/Object;Ljava/util/concurrent/Executor;)V
       8: return
```

PR #213, opened 2017-04-17
[CURATOR-395] Potential null dereference in PersistentNode

Randgalt:
localCreateMethod could be null in a race condition


cammckenzie:
I think that the fix is fine, but why is the createMethod a lazily instantiated reference? Couldn't it just be a normal reference that gets created in the constructor? We seem to have everything we need there to create it.

Randgalt:
> why is the createMethod a lazily instantiated reference? Couldn't it just be a normal reference that gets created in the constructor

It's due to https://issues.apache.org/jira/browse/CURATOR-364

cammckenzie:
Ah, I see now. It's the Preconditions.checkState call that causes the issue. I thought that we could just hold a reference to the builder and then call forPath() on it when required. But that checkState call will fail if the client is not started.

So, the PR looks OK to me. Merge away.

Randgalt:
TBH - I'd totally forgotten why it was done this way. I had to do a git bisect to find the reason.

PR #215, opened 2017-04-21
 CURATOR-106: Ensure background operations happen in the background thread to avoid stack overflows when retrying guaranteed operations

srdo:
See https://issues.apache.org/jira/browse/CURATOR-106?focusedCommentId=15979112&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-15979112

Also a few tests in TestNamespaceFacade were not passing because they were using blank connect strings. I'd be happy to open a separate issue for that change if necessary, or remove it if I misunderstood the tests.

### DIFF
--- a/curator-framework/pom.xml
+++ b/curator-framework/pom.xml
@@ -67,6 +67,12 @@
             <artifactId>slf4j-log4j12</artifactId>
             <scope>test</scope>
         </dependency>
+        
+        <dependency>
+            <groupId>org.mockito</groupId>
+            <artifactId>mockito-core</artifactId>
+            <scope>test</scope>
+        </dependency>
     </dependencies>
 
 </project>
--- a/curator-framework/src/main/java/org/apache/curator/framework/imps/CuratorFrameworkImpl.java
+++ b/curator-framework/src/main/java/org/apache/curator/framework/imps/CuratorFrameworkImpl.java
@@ -504,7 +504,7 @@ public class CuratorFrameworkImpl implements CuratorFramework
         boolean isInitialExecution = (event == null);
         if ( isInitialExecution )
         {
-            performBackgroundOperation(operationAndData);
+            queueOperation(operationAndData);
             return;
         }
 
--- a/curator-framework/src/main/java/org/apache/curator/framework/imps/CuratorFrameworkImpl.java
+++ b/curator-framework/src/main/java/org/apache/curator/framework/imps/CuratorFrameworkImpl.java
@@ -502,35 +502,22 @@ public class CuratorFrameworkImpl implements CuratorFramework
     <DATA_TYPE> void processBackgroundOperation(OperationAndData<DATA_TYPE> operationAndData, CuratorEvent event)
     {
         boolean isInitialExecution = (event == null);
-        if ( isInitialExecution )
-        {
+        if (isInitialExecution) {
+            queueOperation(operationAndData);
+            return;
+        }
+        
+        if(RetryLoop.shouldRetry(event.getResultCode()) && checkBackgroundRetry(operationAndData, event)) {
             queueOperation(operationAndData);
             return;
         }
 
-        boolean doQueueOperation = false;
-        do
-        {
-            if ( RetryLoop.shouldRetry(event.getResultCode()) )
-            {
-                doQueueOperation = checkBackgroundRetry(operationAndData, event);
-                break;
-            }
-
-            if ( operationAndData.getCallback() != null )
-            {
-                sendToBackgroundCallback(operationAndData, event);
-                break;
-            }
-
-            processEvent(event);
+        if (operationAndData.getCallback() != null) {
+            sendToBackgroundCallback(operationAndData, event);
+            return;
         }
-        while ( false );
 
-        if ( doQueueOperation )
-        {
-            queueOperation(operationAndData);
-        }
+        processEvent(event);
     }
 
     <DATA_TYPE> void queueOperation(OperationAndData<DATA_TYPE> operationAndData)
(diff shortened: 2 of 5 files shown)


Randgalt:
This will break most clients. It's vital that, under normal circumstances, background callbacks execute from ZooKeeper's thread. ZooKeeper uses the same thread for watchers and background callbacks. This change will break that.

srdo:
Hi @Randgalt. Just to make sure, you're talking about calls to BackgroundCallback.processResult, right?

I don't follow how this change causes those to be invoked from a different thread than usual. The changed code is only hit if processBackgroundOperation is called with a null event (isInitialExecution is true). As far as I can tell, the invocations that happen from the Zookeeper callbacks always seem to pass a non-null event, and so don't hit the changed code. 

I took a look at the call sites for BackgroundCallback.processResult, and the only way I can see that this change has changed the calling thread is if CuratorFrameworkImpl.processBackgroundOperation is called with a null parameter, and CuratorFrameworkImpl.performBackgroundOperation then hits a ConnectionLossException, through checkBackgroundRetry -> sendToBackgroundCallback -> BackgroundCallback.processResult. https://github.com/apache/curator/blob/master/curator-framework/src/main/java/org/apache/curator/framework/imps/CuratorFrameworkImpl.java#L857. The other invocations of sendToBackgroundCallback (the only call site for BackgroundCallback.processResult except for the wrapper in Backgrounding) only happen through a call to processBackgroundOperation with a non-null event.

I did a quick spot check using the BackgroundCallback in CuratorFrameworkBackground.testBasic. Printing the thread executing BackgroundCallback.processResult before the change:
The executing thread Thread[main-EventThread,5,main]

after the change:
The executing thread Thread[main-EventThread,5,main]

Randgalt:
OK - I didn't look at the change deeply. I'll look more when I have a chance

srdo:
Thanks :)

Randgalt:
@srdo I've looked at this again more deeply and I stand by my original comment. This code:

```java
        boolean isInitialExecution = (event == null);
        if (isInitialExecution) {
            queueOperation(operationAndData);
            return;
        }
```

Breaks a vital contract. `queueOperation` will add the operation to `backgroundOperations` which are handled by CuratorFrameworkImpl's `executorService`. So, we cannot accept this PR. I'm going to close it unless you can think of a way of solving the original problem differently. I'll look at CURATOR-106 when I can to see if I can understand the issue.

srdo:
@Randgalt Could you elaborate on why this is wrong? If I understand you, the issue is that background callbacks must happen in the Zookeeper thread, not CuratorFrameworkImpl's ExecutorService thread? I still don't see how this is changed by the code you posted. In order for there to be a change in behavior, a Zookeeper thread must call `CuratorFrameworkImpl.processBackgroundOperation` with a null event. If the event is not null, the behavior is unchanged.

I'm having a hard time finding any instances of `CuratorFrameworkImpl.processBackgroundOperation` being called with a null event from a Zookeeper callback. The calls I can find all have a non-null event when called from inside a Zookeeper callback.

It looks to me like all the calls to `processBackgroundOperation` with a null event are coming from `forPath` or equivalent (e.g. `forOperations`). Aren't those running in the user's thread, and not one of Zookeeper's threads?

Do you have an example where the thread executing the background callback has changed, or any calls to `processBackgroundOperation` with a null event from Zookeeper callback code?

Randgalt:
Look at CreateBuilderImpl#pathInBackground()

```java
client.processBackgroundOperation(operationAndData, null);
```

All the DSL builders call `processBackgroundOperation()` with null the first time. Now, you ask, how can a ZooKeeper thread call this? Here's how:

```java
class MyWatcher implements Watcher {
    public void process(WatchedEvent event) {
        curatorClient.create().inBackground().forPath(p);
    }
}
```

It can also happen from a background operation callback.

srdo:
Okay, I think I misunderstood your concern. I thought you were worried about something like this
```
curatorClient.create().inBackground((client, event) -> {
  //I thought your concern was that this code sometimes wouldn't run in the EventThread
}).forPath(p)
```

Isn't your example code already broken? 

```
class MyWatcher implements Watcher {
    public void process(WatchedEvent event) {
        curatorClient.create().inBackground().forPath(p);
        //I'm assuming your concern is that code here may assume that the path is now created?
    }
}
```

If the surrounding code is depending on the create call running in the foreground (in the EventThread), why is it using `inBackground`?  It'll be flaky any time there's a connection issue or a retry (in which case the create operation was already being put on the background thread by `performBackgroundOperation`). 

Otherwise if the code isn't depending on the create call running in the foreground, is there still an issue with the change?

Randgalt:
The watcher code is counting on starting the ZooKeeper operation from the same thread as the Watcher. You're right the connection issues can get in the way of this. But, to me, this is all moot. This is how Curator works today and it would be very disruptive to change it now. We can't change deep behavior like this without it affecting users of the library.

srdo:
Too bad. Thanks for explaining though. Would you be more in favor of this patch if I added a new `processBackgroundOperation` instead of modifying the existing one? The stack overflow error mentioned in CURATOR-106 is happening because the FailedDeleteManager is assuming that the backgrounded delete actually happens in the background. When it doesn't, the stack may overflow because FailedDeleteManager triggers a delete, which may fail, which causes a call to FailedDeleteManager etc. It likely doesn't happen when retries are enabled, because only the first try happens in the foreground.

Randgalt:
Yeah - if we can address the specific problem without changing expected existing behavior that would be great.

srdo:
@Randgalt I'm probably just going to leave this alone. The stack trace I posted on the JIRA issue was from an application running 2.12.0. I've been unable to provoke the stack overflow on 3.x. It still looks like it's possible for the FailedDeleteManager to recurse a bunch through the same path as in the stack trace, but since I can't provoke the error I'm not sure the bug still exists. Since I can't test a potential fix I'd rather not change the code, at least not unless someone encounters the error on 3.x.

PR #216, opened 2017-04-24
CURATOR-402: Curator should not use QuorumServer.addr when updating the connect string dynamically

srdo:
…e the client connect string, resulting in a client unable to reconnect

See https://issues.apache.org/jira/browse/CURATOR-402.

I took a look at adding a test for this, but both addresses are null when using a single testing server, and I couldn't find a way to start a testing cluster where clientAddr is null.

### DIFF
--- a/curator-framework/src/main/java/org/apache/curator/framework/imps/EnsembleTracker.java
+++ b/curator-framework/src/main/java/org/apache/curator/framework/imps/EnsembleTracker.java
@@ -170,8 +170,10 @@ public class EnsembleTracker implements Closeable, CuratorWatcher
             {
                 sb.append(",");
             }
-            InetSocketAddress address = Objects.firstNonNull(server.clientAddr, server.addr);
-            sb.append(address.getAddress().getHostAddress()).append(":").append(address.getPort());
+            InetSocketAddress clientAddress = server.clientAddr;
+            if(clientAddress != null) {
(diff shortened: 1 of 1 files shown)


lvfangmin:
This fix looks good to time, but I'd like @Randgalt to accept it since he has more context about the changes he made in CURATOR-345.

Randgalt:
I'll look at this soon. This needs to be addressed once and for all.

curator-framework/src/main/java/org/apache/curator/framework/imps/EnsembleTracker.java
```diff
@@ -170,8 +170,10 @@ public static String configToConnectionString(QuorumVerifier data) throws Except
             {
                 sb.append(",");
             }
-            InetSocketAddress address = Objects.firstNonNull(server.clientAddr, server.addr);
-            sb.append(address.getAddress().getHostAddress()).append(":").append(address.getPort());
+            InetSocketAddress clientAddress = server.clientAddr;
```
lvfangmin: After ZooKeeper 3.5, if the zoo.cfg is in old format which has clientPort and server list being defined, then the clientAddr will be 0.0.0.0, which shouldn't be used as the server hostAddress. CURATOR-392 reported some issues similar, and it has handled this case.

srdo: You are right. CURATOR-392 looks like a more complete attempt at fixing this, so this PR doesn't seem necessary.

PR #223, opened 2017-05-16
[CURATOR-362] Use provided ACL when creating parent directories

Zoltan-Szekeres:


Randgalt:
I'm not certain this is the correct behavior. The API contract implies only that the ACL is set for the node being created. The ACL provider's job is to manage any other nodes. This might be a breaking change for existing clients. I'm -1 on this. I'm curious what other committers think.

cammckenzie:
I agree that this is now a 'feature' that the parents don't have the ACL set. Perhaps we need to introduce another option to set the ACL on the parents?

Randgalt:
We can add a new method to `ACLable`:

```
public interface ACLable<T>
{
    public T withACL(List<ACL> aclList);

    public T withACL(List<ACL> aclList, boolean applyToParents);
}
```

szekizoli:
I agree that introducing a new behavior is better through a new method. If everyone likes this solution, I'll implement it.

cammckenzie:
Sounds good to me, thanks @szekizoli

curator-framework/src/main/java/org/apache/curator/framework/imps/ExistsBuilderImpl.java
```diff
@@ -185,7 +185,7 @@ public Stat forPath(String path) throws Exception
             OperationAndData<String> operationAndData = new OperationAndData<String>(this, path, backgrounding.getCallback(), null, backgrounding.getContext(), watching);
             if ( createParentContainersIfNeeded || createParentsIfNeeded )
             {
-                CreateBuilderImpl.backgroundCreateParentsThenNode(client, operationAndData, operationAndData.getData(), backgrounding, createParentContainersIfNeeded);
+                CreateBuilderImpl.backgroundCreateParentsThenNode(client, operationAndData, operationAndData.getData(), backgrounding, client.getAclProvider(), createParentContainersIfNeeded);
```
cammckenzie: Should the ExistsBuilder have the ability to set the ACL provider for the parent containers as well, given that the we provide the option of creating parent containers?

szekizoli: I think yes, it'd better if one can set the ACL as well with the option of creating parents, because the ability to set the ACL when creating a node is the expected behaviour.

Randgalt:
What's the status of this PR? I'd like to get a release out.

szekizoli:
I can prepare an update with the ability to add an ACL provider on the ExistsBuilder if createParentNodesIfNeeded() is called, if you think that's required for this change. It seems more consistent to me with having the option to set the ACL.

Randgalt:
Yes, please. If you do it today it can be in the next release.

szekizoli:
I've updated the PR. I've decided to add the withACL(List<ACL>, boolean) method to a new interface called `ParentACLable`, because it seemed to me that the applyToParents doesn't make sense in all case. For example, in the case of an ExistsBuilder, where only the parent nodes are created. But I have a version with the interface not split as well.

PR #224, opened 2017-06-11
CURATOR-417: Update Guava to 22-android version ( supports jdk 1.7) 

ajayk:
* Update pom.xml to use latest version of guava 
https://github.com/google/guava 
Guava comes in two flavors.
The main flavor requires JDK 1.8 or higher.
If you need support for JDK 1.7 or Android, use the Android flavor. You can find the Android Guava source in the android directory.
* Move from Objects.firstNotNull ( removed from guava ) to MoreObjects.firstNotNull
* Move from MoreExecutors.sameThreadExecutor() ( removed from guava in version 19)  to MoreExecutors. directExecutor()

### DIFF
--- a/curator-framework/src/main/java/org/apache/curator/framework/imps/EnsembleTracker.java
+++ b/curator-framework/src/main/java/org/apache/curator/framework/imps/EnsembleTracker.java
@@ -20,7 +20,7 @@
 package org.apache.curator.framework.imps;
 
 import com.google.common.annotations.VisibleForTesting;
-import com.google.common.base.Objects;
+import com.google.common.base.MoreObjects;
 import com.google.common.base.Preconditions;
 import com.google.common.collect.Maps;
 import org.apache.curator.ensemble.EnsembleProvider;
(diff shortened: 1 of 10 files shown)


Randgalt:
Guava is now shaded in the latest release of Curator so why is this needed? Please see https://cwiki.apache.org/confluence/display/CURATOR/TN13

ajayk:
@Randgalt Thanks for the link i was not aware of the https://cwiki.apache.org/confluence/display/CURATOR/TN13 , 
This would help us to leverage JDK  7 and there are few optimizations on the ImmutableCollection https://github.com/google/guava/wiki/Release22

and when the curator moves to JDK 8 this might be handy 
The intent is that the Android version always be a strict subset of what's available in the main/Java 8 version, so anything compiled against version N of the Android version will work when run against the Java 8 version. 

For the medium-term future, Guava users who need to target JDK 7 can use the Android flavor, which has no Android-specific dependencies. At some point in the future, when Android diverges sufficiently from JDK 7, Guava may stop providing a JDK 7-compatible flavor, at which time those users will have to stay with the latest prior version

Randgalt:
So, I can close this?

ajayk:
Yeah , this can be closed , if we don't want to upgrade guava in the curator

PR #225, opened 2017-06-21
CURATOR-295, removed cache close, catch provider close exceptions

serranom:
we've been running this change in shipped production code for a few months now.


Randgalt:
Thanks for this. This will take me a bit of time to look over.

PR #234, opened 2017-08-11
Fix, maybe cannot deleteChildren completely when multi-client delete concurrently

hebelala:


Randgalt:
Please review https://cwiki.apache.org/confluence/display/CURATOR/Submitting+Pull+Requests

- There must be a JIRA Issue created for your change
- Create a git branch with the same name as the Issue

Randgalt:
Also, I don't know what the need for this is. ZKPaths is an internal utility and the Curator code that uses it already ignores the exception.

hebelala:
use `curatorFramework.delete().deletingChildrenIfNeeded().forPath(path)`, this sync api doesn't ignore the NoNodeException, causes the rest of children nodes will not be deleted perhaps.

`zookeeper.getChildren(path, null)` maybe throw NoNodeException, if the path doesn't exist.

hebelala:
see https://github.com/apache/curator/pull/235, please

PR #235, opened 2017-08-12
[CURATOR-430] deletingChildrenIfNeeded maybe cannot delete children completely when multi-client delete concurrently

hebelala:
use `curatorFramework.delete().deletingChildrenIfNeeded().forPath(path)`, this sync api doesn't ignore the NoNodeException, causes the rest of children nodes will not be deleted perhaps.  

`zookeeper.getChildren(path, null)` maybe throw NoNodeException, if the path doesn't exist.

### DIFF
--- a/curator-client/src/main/java/org/apache/curator/utils/ZKPaths.java
+++ b/curator-client/src/main/java/org/apache/curator/utils/ZKPaths.java
@@ -313,7 +313,13 @@ public class ZKPaths
     {
         PathUtils.validatePath(path);
 
-        List<String> children = zookeeper.getChildren(path, null);
+        List<String> children;
+        try {
+            children = zookeeper.getChildren(path, null);
+        } catch (KeeperException.NoNodeException e) {
+            // someone else has deleted the node it since we checked
+            return;
+        }
         for ( String child : children )
         {
             String fullPath = makePath(path, child);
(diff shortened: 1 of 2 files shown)


Randgalt:
Please include a test (something like the code you posted in the issue) for this. A good place to put the test is TestFrameworkEdges.java.

curator-client/src/main/java/org/apache/curator/utils/ZKPaths.java
```diff
@@ -313,7 +313,13 @@ public static void deleteChildren(ZooKeeper zookeeper, String path, boolean dele
     {
         PathUtils.validatePath(path);
 
-        List<String> children = zookeeper.getChildren(path, null);
+        List<String> children;
+        try {
+            children = zookeeper.getChildren(path, null);
+        } catch (KeeperException.NoNodeException e) {
+            // someone else has deleted the node it since we checked
```
lvfangmin: // someone else has deleted the node since we checked

hebelala: haha...  it's copied

hebelala:
Be sure this issue is bug?  If yes, could you publish a fixed version based on v2.10.0?

JunjianS:
i met the same problem, could u please provide a fixed version ?

Randgalt:
> If yes, could you publish a fixed version based on v2.10.0?

Curator 4.0 is now compatible with both ZK 3.4.x and 3.5.x so the 2.x versions of Curator will not be updated.

Randgalt:
@hebelala hey - the test for this PR keeps failing and as I read it, I don't see how it ever worked. Can you look at this again? The code `client2.delete().forPath("/parent/child" + (childCount / 2));` should actually fail most of the time as the thread has deleted all the children.

hebelala:
@Randgalt I have fix the test case, should I commit a new PR?

PR #236, opened 2017-09-11
CURATOR-431 - Fixed stat population during create

cammckenzie:
-The stat object was not being populated if the create failed due to the node already existing.

### DIFF
--- a/curator-framework/src/main/java/org/apache/curator/framework/imps/CreateBuilderImpl.java
+++ b/curator-framework/src/main/java/org/apache/curator/framework/imps/CreateBuilderImpl.java
@@ -38,6 +38,8 @@ import org.apache.zookeeper.KeeperException;
 import org.apache.zookeeper.Op;
 import org.apache.zookeeper.data.ACL;
 import org.apache.zookeeper.data.Stat;
+import org.apache.zookeeper.server.DataTree;
+
 import java.util.List;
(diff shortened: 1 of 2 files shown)


curator-framework/src/main/java/org/apache/curator/framework/imps/CreateBuilderImpl.java
```diff
@@ -1212,7 +1212,21 @@ public String call() throws Exception
                             {
                                 if ( setDataIfExists )
                                 {
-                                    client.getZooKeeper().setData(path, data, setDataIfExistsVersion);
+                                    Stat setStat = client.getZooKeeper().setData(path, data, setDataIfExistsVersion);
+                                    if(storingStat != null)
+                                    {
+                                        storingStat.setAversion(setStat.getAversion());
```
Randgalt: You can use `DataTree#copyStat()` - FYI

Randgalt: Other than that LGTM

cammckenzie: Thanks, there was another place in CreateBuilderImpl doing the same thing, and I've replaced that with DataTree#copyStat() also.

PR #237, opened 2017-09-14
[CURATOR-432] adds import exclusion to curator-framework bundle for org.apache.curator

christopher-johnson:
org.apache.curator is not an OSGI bundle so it cannot be imported.  This causes unsatisfied requirements exceptions on deployment.  This PR adds the felix plugin configuration for the exclusion in org.apache.curator.framework.  Note that this exclusion could be applied in the parent pom or to each bundle separately, but the same issue exists for other curator bundles.

### DIFF
--- a/curator-framework/pom.xml
+++ b/curator-framework/pom.xml
@@ -37,7 +37,7 @@
 
     <properties>
         <osgi.import.package>
-            *
+            !org.apache.curator,*
         </osgi.import.package>
         <osgi.export.package>
             org.apache.curator.framework*;version="${project.version}";-noimport:=true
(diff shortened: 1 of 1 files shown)


acoburn:
I actually think the issue here is _not_ with the `org.apache.curator/curator-client` bundle (which _is_ an OSGi bundle), but rather with the underlying zookeeper bundle. The issue there is that ZK uses a dash in its versioning scheme (3.5.x-alpha or 3.5.3-beta) and that version becomes part of the `Bundle-Version` in the manifest file. OSGi versioning can't parse that dash; typically, in such situations the dash is replaced with a period for the `Bundle-Version` field in the manifest.

As an experiment, I re-wrapped the ZK bundle (using the version 3.5.3.beta). Then it became possible to deploy ZK 3.5 in Karaf, and once that was in place, all of the 4.0.0 Curator bundles could be both deployed and started in Karaf.

TL;DR -- the Curator bundles are already in fine shape for OSGi, the issue is upstream with Zookeeper, which itself will be resolved as soon as the `-beta` is dropped from the version.

christopher-johnson:
the framework imports the root org.curator.* but the client does not.  Any import dependency on the framework will fail.

curator-framework/pom.xml
```diff
@@ -100,6 +100,27 @@
...
                 </executions>
             </plugin>
+            <plugin>
+                <groupId>org.apache.felix</groupId>
+                <artifactId>maven-bundle-plugin</artifactId>
+                <version>3.3.0</version>
+                <extensions>true</extensions>
+                <configuration>
+                    <instructions>
+                        <Import-Package>!org.apache.curator,*</Import-Package>
```
iocanel: There is no need to define the plugin again. The plugin is defined once in the parent pom and inherited from the rest of the modules.

To customize a modules configuration please set properties:
- `osgi.import.package`
- `osgi.export.package`

PR #238, opened 2017-10-12
[CURATOR-436] getSortedChildren() should ignore NoNode exceptions

Randgalt:
getSortedChildren() should ignore NoNode exceptions and just treat it as if there are no children. This works around issues with container nodes, parents not yet created, etc.


cammckenzie:
Looks good to me.

PR #239, opened 2017-10-19
Add note to zk compatibility page that curator-test should not be upg…

srdo:
…raded to 4.0 when testing against Zookeeper 3.4.x

It seems useful to put a note in about this, since it isn't obvious that you should do this unless you happen to look at https://github.com/apache/curator/blob/master/curator-test-zk34/README.md.


Randgalt:
Duplicate of CURATOR-446 - https://github.com/apache/curator/pull/248

PR #243, opened 2017-11-28
[Fix] Curator-444

borjab:
This ensures that not two nodes are simultaneusly as leaders. 

If setLeadership(false) is redundant no event would be reported anyway so it is quite safe.

### DIFF
--- a/curator-recipes/src/test/java/org/apache/curator/framework/recipes/leader/TestLeaderLatchIsolatedZookeeper.java
+++ b/curator-recipes/src/test/java/org/apache/curator/framework/recipes/leader/TestLeaderLatchIsolatedZookeeper.java
@@ -33,8 +33,8 @@ import org.apache.zookeeper.server.quorum.Leader;
 import org.apache.zookeeper.server.quorum.LearnerHandler;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
-import org.testng.annotations.AfterTest;
-import org.testng.annotations.BeforeTest;
+import org.testng.annotations.AfterMethod;
+import org.testng.annotations.BeforeMethod;
 import org.testng.annotations.Test;
 
 /** Tests issue: CURATOR-444: Zookeeper node is isolated from other zookeepers but not from curator
@@ -51,8 +51,8 @@ public class TestLeaderLatchIsolatedZookeeper {
 	private DummyLeaderLatch firstClient;
 	private DummyLeaderLatch secondClient;
 
-	@BeforeTest
-	public void beforeTest() throws Exception {
+	@BeforeMethod
+	public void beforeMethod() throws Exception {
 		cluster = new TestingCluster(3);
 		cluster.start();
 		firstClient  = new DummyLeaderLatch(cluster.getConnectString(), SESSION_TIMEOUT_MS, CONNECTION_TIMEOUT_MS, "First");
@@ -62,8 +62,8 @@ public class TestLeaderLatchIsolatedZookeeper {
 		logger.info("Session: " + SESSION_TIMEOUT_MS + " connection " + CONNECTION_TIMEOUT_MS );
 	}
 
-	@AfterTest
-	public void afterTest() throws IOException {
+	@AfterMethod
+	public void afterMethod() throws IOException {
 		DummyLeaderLatch.resetHistory();
 		firstClient.stop();
 		secondClient.stop();
@@ -81,15 +81,14 @@ public class TestLeaderLatchIsolatedZookeeper {
 		assertEquals(secondClient.isLeaderAccordingToLatch(), secondClient.isLeaderAccordingToEvents());
 	}
 
-	@Test
+	@Test(invocationCount = 10)
 	public void testThatResistsNetworkGlitches() throws Exception {
 
 		blockLeaderListeningForSomeTime(SESSION_TIMEOUT_MS);
 
 		Thread.sleep(SESSION_TIMEOUT_MS * 3);
 
-		assertTrue(isHistoryValid());
-		logger.info("History :" + DummyLeaderLatch.getEventHistory());
+		assertTrue(isHistoryValid(), "History is not valid: " + DummyLeaderLatch.getEventHistory());
 
 		assertEquals(   firstClient.isLeaderAccordingToLatch(),  firstClient.isLeaderAccordingToEvents());
 		assertEquals(  secondClient.isLeaderAccordingToLatch(), secondClient.isLeaderAccordingToEvents());
@@ -124,10 +123,9 @@ public class TestLeaderLatchIsolatedZookeeper {
 
 		@Override
 		public void run() {
-
 			//suspendThread();
 			//closeSocket();
-			lockSocket();
+			//lockSocket();
 			forceTimeout();
 		}
 
@@ -136,7 +134,7 @@ public class TestLeaderLatchIsolatedZookeeper {
 				Socket socket = learnerHandler.getSocket();
 				int oldTimeout = learnerHandler.getSocket().getSoTimeout();
 				socket.setSoTimeout(1);
-				Thread.sleep(milliseconds * 3);
+				Thread.sleep(SESSION_TIMEOUT_MS);
 				socket.setSoTimeout(oldTimeout);
 
 			}catch (SocketException ex){
@@ -198,6 +196,7 @@ public class TestLeaderLatchIsolatedZookeeper {
 	}
 
 	private boolean isHistoryValid(){
+		logger.info("EVENTS HISTORY :" + DummyLeaderLatch.getEventHistory());
 
 		boolean hadLeader = false;
 		String lastLeaderId = null;
--- a/curator-recipes/src/test/java/org/apache/curator/framework/recipes/leader/TestLeaderLatchIsolatedZookeeper.java
+++ b/curator-recipes/src/test/java/org/apache/curator/framework/recipes/leader/TestLeaderLatchIsolatedZookeeper.java
@@ -1,15 +1,38 @@
+/**
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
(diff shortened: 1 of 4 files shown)


borjab:
Fix now includes test to reproduce error.

curator-recipes/src/test/java/org/apache/curator/framework/recipes/leader/TestLeaderLatchIsolatedZookeeper.java
```diff
@@ -0,0 +1,224 @@
...
+		for(TestingZooKeeperServer server : cluster.getServers()) {
+
+			if (server.getQuorumPeer().leader != null) {
+				Leader leader = server.getQuorumPeer().leader;
+
+				for (LearnerHandler learnerHandler : leader.getLearners()) {
+
+					logger.info("Locking " + learnerHandler.getName());
+					HandlerLocker locker = new HandlerLocker(learnerHandler, milliseconds);
+					locker.start();
```
cammckenzie: These threads never get stopped again do they?

curator-recipes/src/test/java/org/apache/curator/framework/recipes/leader/TestLeaderLatchIsolatedZookeeper.java
```diff
@@ -0,0 +1,224 @@
...
+			this.learnerHandler = learnerHandler;
+			this.milliseconds = milliseconds;
+		}
+
+		@Override
+		public void run() {
+			//suspendThread();
+			//closeSocket();
+			//lockSocket();
+			forceTimeout();
```
cammckenzie: If these commented out things aren't used, then can we remove?

curator-recipes/src/test/java/org/apache/curator/framework/recipes/leader/TestLeaderLatchIsolatedZookeeper.java
```diff
@@ -0,0 +1,224 @@
...
+			}catch (SocketException ex){
+				logger.error("Error forcing timeout", ex);
+			} catch (InterruptedException ex) {
+				logger.error("Interrupted while forcing timeout", ex);
+			}
+		}
+
+
+		private void lockSocket(){
+			try{
```
cammckenzie: Method is not used?

curator-recipes/src/test/java/org/apache/curator/framework/recipes/leader/TestLeaderLatchIsolatedZookeeper.java
```diff
@@ -0,0 +1,224 @@
...
+
+			}catch (InterruptedException ex){
+				String message = "Failed to lock socket for " + milliseconds + " ms";
+				logger.error(message);
+				Thread.currentThread().interrupt();
+				throw new RuntimeException(message, ex);
+			}
+		}
+
+		private void suspendThread(){
```
cammckenzie: Method is not used?

curator-recipes/src/test/java/org/apache/curator/framework/recipes/leader/TestLeaderLatchIsolatedZookeeper.java
```diff
@@ -0,0 +1,224 @@
...
+
+				String message = "Failed to wait " + milliseconds + " ms";
+				logger.error(message);
+				Thread.currentThread().interrupt();
+				throw new RuntimeException(message, ex);
+			}
+
+		}
+
+		private void closeSocket(){
```
cammckenzie: Method is not used?

curator-recipes/src/test/java/org/apache/curator/framework/recipes/leader/testing/DummyLeaderLatch.java
```diff
@@ -0,0 +1,189 @@
...
+package org.apache.curator.framework.recipes.leader.testing;
+
+import static org.apache.curator.framework.recipes.leader.LeaderLatch.CloseMode.NOTIFY_LEADER;
+
+import java.io.IOException;
+import java.util.LinkedList;
+import java.util.List;
+import java.util.concurrent.atomic.AtomicBoolean;
+
+import javax.xml.ws.Holder;
```
cammckenzie: Unused import?

curator-recipes/src/test/java/org/apache/curator/framework/recipes/leader/testing/DummyLeaderLatch.java
```diff
@@ -0,0 +1,189 @@
...
+
+import org.apache.curator.RetryPolicy;
+import org.apache.curator.framework.CuratorFramework;
+import org.apache.curator.framework.CuratorFrameworkFactory;
+import org.apache.curator.framework.recipes.leader.LeaderLatch;
+import org.apache.curator.framework.recipes.leader.LeaderLatchListener;
+import org.apache.curator.framework.recipes.leader.Participant;
+import org.apache.curator.retry.ExponentialBackoffRetry;
+import org.apache.curator.test.Timing;
```
cammckenzie: Unused import?

curator-recipes/src/test/java/org/apache/curator/framework/recipes/leader/TestLeaderLatchIsolatedZookeeper.java
```diff
@@ -0,0 +1,224 @@
...
+	}
+
+	@Test
+	public void testThatStartCoherent() throws Exception {
+		assertEquals(firstClient.isLeaderAccordingToLatch(),  firstClient.isLeaderAccordingToEvents());
+		assertEquals(secondClient.isLeaderAccordingToLatch(), secondClient.isLeaderAccordingToEvents());
+	}
+
+	@Test(invocationCount = 10)
+	public void testThatResistsNetworkGlitches() throws Exception {
```
cammckenzie: Is it possible to make the test more reliable? Running it 10 times at 30 seconds an invocation increases our regression run by 5 minutes, which isn't ideal.

borjab:
Thanks for the feedback. You are right that the test has a lot of room for improvement.  I will work on the tests as requested.

cammckenzie:
Thanks!

curator-recipes/src/test/java/org/apache/curator/framework/recipes/leader/testing/DummyLeaderLatch.java
```diff
@@ -0,0 +1,189 @@
...
+
+	/**
+	 * Synchronously creates all nodes in a path in ZooKeeper if they don't exist.
+	 * <p>
+	 * All API calls provided by Curator for path creation with parents do async
+	 *
+	 * @param path the path
+	 */
+	private void createPath(final String path) throws Exception {
+		final String[] split = path.split("/");
```
Randgalt: why don't you use `create().creatingParentsIfNeeded()`?

curator-recipes/src/test/java/org/apache/curator/framework/recipes/leader/testing/DummyLeaderLatch.java
```diff
@@ -0,0 +1,189 @@
...
+				.connectionTimeoutMs(connectionTimeoutMs)
+				.build();
+		leaderLatch = new LeaderLatch(curatorFramework, LATCH_PATH, instanceId, NOTIFY_LEADER);
+		leaderLatch.addListener(this);
+	}
+
+	public void start() throws Exception {
+		curatorFramework.start();
+		curatorFramework.blockUntilConnected();
+		createPath(LATCH_PATH);
```
Randgalt: The leader latch code will create the path for you - there's no need for this.

egor-ryashin:
We've got continuous issues with that bug (two critical nodes becoming leaders simultaneously begin to damage the work of each other), and I would like this work to be continued, though I see it lost attention some time ago. Right now I'm looking at the code trying to figure out where exactly it's broken and I found this line quite strange https://github.com/apache/curator/pull/243/files#diff-edf135a1ae8f33bd78d8371b25f2d5a6R544
IE, we got 3 children:
> 0
> 1
> 2

And **ourIndex** = 2, why would we add a watcher for the child 1?
> sortedChildren.get(ourIndex - 1)

cammckenzie:
This is just done for efficiency. Each client involved in the leader election only cares when the child before it disconnects, because that implies that it should now become leader. If any of the other children disconnect, the client doesn't care because it doesn't affect whether it should become leader or not.

Randgalt:
> And ourIndex = 2, why would we add a watcher for the child 1?
This prevents thundering herds when ZNodes are deleted. A given client only needs to watch the ZNode previous to it thereby it only gets notified when it will be the next lock/leader holder.

Randgalt:
@borjab and @egor-ryashin I just ran the test in this PR multiple times with the fix in LeaderLatch commented out and the tests always pass. It's important to note that if you don't use the default `ConnectionStateErrorPolicy` and/or elsewhere don't exit your locks, etc. when you receive SUSPENDED you can get in this state. Even then with long GC times there is always the possibility of multiple lock/leader holders. See https://cwiki.apache.org/confluence/display/CURATOR/TN10

egor-ryashin:
@Randgalt I've actually seen several leaders working several minutes till I restart them. 
My guess, right now, that can happen when the leader has lost connection for a spell. For example, for children:
A
B
here B watches A, while A doesn't watch anything, I guess. If A got a connectivity issue, B becomes a leader and set no watches, then A get through and becomes the leader again, due to the sorting rule, while B doesn't receive any notification about it. 
Haven't had the time to prove it yet though.

Randgalt:
@egor-ryashin For `LeaderLatch` calls `setLeadership(false);` when the connection is lost. `LeaderSelector` is a bit more complicated but if you use the recommended `LeaderSelectorListenerAdapter` the active leader will get interrupted when the connection is lost which will result in a release of the its internal lock. In both cases the lock node (`A` in your example) gets deleted on connection loss.

egor-ryashin:
@Randgalt no, it doesn't recover, only a restart helps.

tisonkun:

PR #247, opened 2017-12-17
About InterProcessMutex.lockCount

hxy1991:
The "lockCount" do not share in multithreading, so the "int" type is sufficient.


Randgalt:
I don't know why you think lockCount is not involved in "multithreading" but I assure you that it is.

PR #249, opened 2017-12-20
[CURATOR-388] PathChildrenCache stops working if container node is auto-removed

Randgalt:
PathChildrenCache stops working if container node is auto-removed and later recreated - test and a fix


cammckenzie:
I think this looks OK.

PR #250, opened 2018-01-02
[CURATOR-447] TreeCache: Improve memory usage and concurrent update logic

njhill:
Jira https://issues.apache.org/jira/browse/CURATOR-374 reduced per-node memory usage in `TreeCache`. It can be improved further via removal of the `nodeState` field - its `LIVE` state corresponds exactly to the adjacent `childData` field being non-null, and a sentinel `ChildData` value can be used for the `DEAD` state. This simplification also reduces the room for bugs and state inconsistencies.

Other improvements included:
* More robust cache update logic (in get-children and get-data event callbacks)
* ~~A further simplification to have `TreeNode` extend `AtomicReference`, which obviates the need for an explicit `childData` field~~
* ~~Avoid overhead of incrementing/decrementing the `outstandingOps` atomic integer post-initialization~~

See corresponding JIRA: [CURATOR-447](https://issues.apache.org/jira/browse/CURATOR-447)

### DIFF
--- a/curator-recipes/src/main/java/org/apache/curator/framework/recipes/cache/TreeCache.java
+++ b/curator-recipes/src/main/java/org/apache/curator/framework/recipes/cache/TreeCache.java
@@ -442,7 +442,7 @@ public class TreeCache implements Closeable
                     {
                         final ChildData oldChildData = childData;
                         // Ignore this event if we've already processed a newer update for this node.
-                        if ( (isLive(oldChildData) && newStat.getMzxid() <= oldChildData.getStat().getMzxid()) )
+                        if ( isLive(oldChildData) && newStat.getMzxid() <= oldChildData.getStat().getMzxid() )
                         {
                             break;
                         }
--- a/curator-recipes/src/main/java/org/apache/curator/framework/recipes/cache/TreeCache.java
+++ b/curator-recipes/src/main/java/org/apache/curator/framework/recipes/cache/TreeCache.java
@@ -196,25 +196,18 @@ public class TreeCache implements Closeable
         return new Builder(client, path);
     }
 
-    private enum NodeState
-    {
-        PENDING, LIVE, DEAD
+    private static final ChildData DEAD = new ChildData("/",null,null);
+    
+    private static boolean isLive(ChildData cd) {
+        return cd != null && cd != DEAD;
     }
 
-    private static final AtomicReferenceFieldUpdater<TreeNode, NodeState> nodeStateUpdater =
-            AtomicReferenceFieldUpdater.newUpdater(TreeNode.class, NodeState.class, "nodeState");
+    private static final AtomicReferenceFieldUpdater<TreeNode, ConcurrentMap<String,TreeNode>> childrenUpdater =
+            (AtomicReferenceFieldUpdater)AtomicReferenceFieldUpdater.newUpdater(TreeNode.class, ConcurrentMap.class, "children");
 
-    private static final AtomicReferenceFieldUpdater<TreeNode, ChildData> childDataUpdater =
-            AtomicReferenceFieldUpdater.newUpdater(TreeNode.class, ChildData.class, "childData");
-
-    private static final AtomicReferenceFieldUpdater<TreeNode, ConcurrentMap> childrenUpdater =
-            AtomicReferenceFieldUpdater.newUpdater(TreeNode.class, ConcurrentMap.class, "children");
-
-    private final class TreeNode implements Watcher, BackgroundCallback
+    private final class TreeNode extends AtomicReference<ChildData> implements Watcher, BackgroundCallback
     {
 
-        volatile NodeState nodeState = NodeState.PENDING;
-        volatile ChildData childData;
         final TreeNode parent;
         final String path;
         volatile ConcurrentMap<String, TreeNode> children;
@@ -297,7 +290,11 @@ public class TreeCache implements Closeable
 
         void wasDeleted() throws Exception
         {
-            ChildData oldChildData = childDataUpdater.getAndSet(this, null);
+            ChildData oldChildData;
+            do {
+               if( (oldChildData = get()) == DEAD ) return;
+            } while( !compareAndSet(oldChildData, DEAD) );
+            
             ConcurrentMap<String, TreeNode> childMap = childrenUpdater.getAndSet(this,null);
             if ( childMap != null )
             {
@@ -314,8 +311,7 @@ public class TreeCache implements Closeable
                 return;
             }
 
-            NodeState oldState = nodeStateUpdater.getAndSet(this, NodeState.DEAD);
-            if ( oldState == NodeState.LIVE )
+            if ( isLive(oldChildData) )
             {
                 publishEvent(TreeCacheEvent.Type.NODE_REMOVED, oldChildData);
             }
@@ -377,19 +373,20 @@ public class TreeCache implements Closeable
                 Preconditions.checkState(parent == null, "unexpected EXISTS on non-root node");
                 if ( event.getResultCode() == KeeperException.Code.OK.intValue() )
                 {
-                    nodeStateUpdater.compareAndSet(this, NodeState.DEAD, NodeState.PENDING);
+                    compareAndSet(DEAD, null);
                     wasCreated();
                 }
                 break;
             case CHILDREN:
                 if ( event.getResultCode() == KeeperException.Code.OK.intValue() )
                 {
-                    ChildData oldChildData = childData;
-                    if ( oldChildData != null && oldChildData.getStat().getMzxid() == newStat.getMzxid() )
+                    ChildData oldChildData = get();
+                    //TODO consider doing update of cversion, pzxid, numChildren only
+                    if ( isLive(oldChildData) && oldChildData.getStat().getMzxid() == newStat.getMzxid() )
                     {
                         // Only update stat if mzxid is same, otherwise we might obscure
                         // GET_DATA event updates.
-                        childDataUpdater.compareAndSet(this, oldChildData, new ChildData(oldChildData.getPath(), newStat, oldChildData.getData()));
+                        compareAndSet(oldChildData, new ChildData(oldChildData.getPath(), newStat, oldChildData.getData()));
                     }
 
                     if ( event.getChildren().isEmpty() )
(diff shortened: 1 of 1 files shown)


njhill:
@dragonsinth @Randgalt any interest in these updates? FWIW I've had them running in production for some time.

Randgalt:
I'll defer to @dragonsinth

dragonsinth:
Hi @njhill just got back to work from vacation, and this is a little more than I can just eyeball, I gotta think through it a bit when I have more time.  I think I would like to see the removal of `nodeState` in isolation, perhaps with an explicit `PENDING` sentinel rather than just `null`.

It's not clear to me that `TreeNode extends AtomicReference` and `@Nullable outstandingOps` are wins.  The first one doesn't really save any memory, and for the second one, the ongoing CPU cost to update `outstandingOps` is basically nothing, it's almost certainly immeasurably tiny, and you have to introduce racy field reads to accomplish it.

Have you done memory use measurements?  I would expect the memory footprint of `ConcurrentMap<String, TreeNode> children` to mostly dominate the memory overhead of our own classes, not to mention all the Watchers.

njhill:
Thanks @dragonsinth, there's of course no rush at all!

I agree that `TreeNode extends AtomicReference` isn't a "win" as such, it just looked cleaner to me than using another `AtomicReferenceFieldUpdater`. But it's likely subjective and certainly a minor thing.

The `outstandingOps` change seems clearer to me - the field has no use for most of the life of the cache and yet is incremented/decremented on every single change. It's not the CPU cycle overhead but more the (CPU) cache coherency overhead since it's an `AtomicLong`. I didn't think the change introduces any raciness - the only times when `outstandingOps` needs to be modified (during initialization or re-initialization) it will be guaranteed to be non-null.

> I would expect the memory footprint of `ConcurrentMap<String, TreeNode>` children to mostly dominate the memory overhead of our own classes

Good point. The description of [CURATOR-374](https://issues.apache.org/jira/browse/CURATOR-374) implied this might have _some_ mem overhead benefit for very large caches, but removal of the `nodeState` field was motivated more by state simplification and consolidation of atomic updates. I'll do some quick measurements but I acknowledge percentage-wise the saving might not be significant.

njhill:
@dragonsinth I've removed the `outstandingOps` change commit from the PR, and added a new commit which reverts the `TreeNode extends AtomicReference<ChildData>` change. Hopefully that helps to make this more palatable!

njhill:
@dragonsinth any interest in these changes? Maybe I should rename the PR since as you pointed out the memory saving aspect isn't very significant (I measured about 2% on a cache of 50k empty nodes), but hopefully the state and update logic simplification is beneficial?

dragonsinth:
Sorry, been busy then out of town. I'll look tomorrow.

dragonsinth:
I'm sure this is probably right, but the old code has a _lot_ of miles on it.  I have to read through super carefully and convince myself that there are no behavioral changes.

curator-recipes/src/main/java/org/apache/curator/framework/recipes/cache/TreeCache.java
```diff
@@ -377,15 +375,16 @@ public void processResult(CuratorFramework client, CuratorEvent event) throws Ex
...
+                    childDataUpdater.compareAndSet(this, DEAD, null);
                     wasCreated();
                 }
                 break;
             case CHILDREN:
                 if ( event.getResultCode() == KeeperException.Code.OK.intValue() )
                 {
                     ChildData oldChildData = childData;
-                    if ( oldChildData != null && oldChildData.getStat().getMzxid() == newStat.getMzxid() )
+                    //TODO consider doing update of cversion, pzxid, numChildren only
```
dragonsinth: seems reasonable to me

curator-recipes/src/main/java/org/apache/curator/framework/recipes/cache/TreeCache.java
```diff
@@ -397,14 +396,11 @@ public void processResult(CuratorFramework client, CuratorEvent event) throws Ex
...
-                    if ( childMap == null )
+                    ConcurrentMap<String, TreeNode> childMap;
+                    while ( (childMap = children) == null )
                     {
                         childMap = Maps.newConcurrentMap();
-                        if ( !childrenUpdater.compareAndSet(this, null, childMap) )
-                        {
-                            childMap = children;
-                        }
+                        if ( childrenUpdater.compareAndSet(this, null, childMap) ) break;
```
dragonsinth: I see why you turned this into a loop (nice catch), but I think it reads a little clearly and more deliberateif you revert back to the original code and just change the `if` -> `while`.  Ideally if Curator moves wholesale to Java 1.8,  funcs, like:

```
                    ConcurrentMap<String, TreeNode> childMap = childrenUpdater.updateAndGet(this, (oldChildren) -> {
                        return oldChildren != null ? oldChildren : Maps.newConcurrentMap();
                    });
```

curator-recipes/src/main/java/org/apache/curator/framework/recipes/cache/TreeCache.java
```diff
@@ -704,8 +687,7 @@ private TreeNode find(String findPath)
             {
                 TreeNode childNode = entry.getValue();
                 ChildData childData = childNode.childData;
-                // Double-check liveness after retreiving data.
```
dragonsinth: I kind of liked this comment, although I misspelled "retrieving" here and below

curator-recipes/src/main/java/org/apache/curator/framework/recipes/cache/TreeCache.java
```diff
@@ -728,13 +710,12 @@ private TreeNode find(String findPath)
...
         TreeNode node = find(fullPath);
-        if ( node == null || node.nodeState != NodeState.LIVE )
+        if ( node == null )
         {
             return null;
         }
         ChildData result = node.childData;
-        // Double-check liveness after retreiving data.
-        return node.nodeState == NodeState.LIVE ? result : null;
+        return result != DEAD ? result : null;
```
dragonsinth: shouldn't this be `isLive` for consistency?

curator-recipes/src/main/java/org/apache/curator/framework/recipes/cache/TreeCache.java
```diff
@@ -436,41 +432,28 @@ else if ( event.getResultCode() == KeeperException.Code.NONODE.intValue() )
...
+                    String eventPath = event.getPath();
+                    ChildData toPublish = new ChildData(eventPath, newStat, event.getData());
+                    ChildData toUpdate = cacheData ? toPublish : new ChildData(eventPath, newStat, null);
+                    while (true) {
+                        final ChildData oldChildData = childData;
+                        if ( (isLive(oldChildData) && newStat.getMzxid() <= oldChildData.getStat().getMzxid())
+                                // Ordinary nodes are not allowed to transition from dead -> live;
+                                // make sure this isn't a delayed response that came in after death.
+                                || (parent != null && oldChildData == DEAD) ) {
+                            break;
```
dragonsinth: Can we break up these cases for clarity?

```
                        final ChildData oldChildData = childData;
                        // Ignore this event if we've already processed a newer update for this node.
                        if ( (isLive(oldChildData) && newStat.getMzxid() <= oldChildData.getStat().getMzxid()) )
                        {
                            break;
                        }
                        // Non-root nodes are not allowed to transition from dead -> live;
                        // make sure this isn't a delayed response that came in after death.
                        if ( parent != null && oldChildData == DEAD )
                        {
                            break;
                        }
```

curator-recipes/src/main/java/org/apache/curator/framework/recipes/cache/TreeCache.java
```diff
@@ -436,41 +432,28 @@ else if ( event.getResultCode() == KeeperException.Code.NONODE.intValue() )
...
-                            publishEvent(TreeCacheEvent.Type.NODE_UPDATED, toPublish);
+                            if ( !isLive(oldChildData) )
+                            {
+                                publishEvent(TreeCacheEvent.Type.NODE_ADDED, toPublish);
+                            }
+                            else if ( oldChildData.getStat().getMzxid() != newStat.getMzxid() )
+                            {
+                                publishEvent(TreeCacheEvent.Type.NODE_UPDATED, toPublish);
+                            }
+                            break;
```
dragonsinth: I think the second check is now superfluous since we checked above.

```
                        if ( childDataUpdater.compareAndSet(this, oldChildData, toUpdate) )
                        {
                            publishEvent(isLive(oldChildData) ? TreeCacheEvent.Type.NODE_UPDATED : TreeCacheEvent.Type.NODE_ADDED, toPublish);
                            break;
                        }
```

dragonsinth:
@njhill this looks good to me with a few changes.  I pushed one extra commit onto https://github.com/apache/curator/tree/CURATOR-447 for you to look at.  If you agree with my changes (some of them are to conform to the project style) please sync your branch to mine and I'll merge this.

curator-recipes/src/main/java/org/apache/curator/framework/recipes/cache/TreeCache.java
njhill: The reason I removed it is that unlike before it's not really "double" checking anymore against the separate `nodeState` field, there's only a single check for liveness which is now implicit in the value of `childData`. But I could well have misinterpreted the reason for comment, no problem with leaving it in!

curator-recipes/src/main/java/org/apache/curator/framework/recipes/cache/TreeCache.java
njhill: Reason was that `isLive` also checks for `null` which isn't necessary here (since we're returning `null` anyhow). But no harm in making it consistent.

njhill:
Thanks alot @dragonsinth for going through this so carefully. I like your additional simplifications, and sorry for the project style inconsistencies that I missed!

I've pulled in your commit and added one more just to remove some parentheses no longer needed after splitting those conditions into separate `if`s.

curator-recipes/src/main/java/org/apache/curator/framework/recipes/cache/TreeCache.java
```diff
@@ -702,19 +687,18 @@ private TreeNode find(String findPath)
             ImmutableMap.Builder<String, ChildData> builder = ImmutableMap.builder();
             for ( Map.Entry<String, TreeNode> entry : map.entrySet() )
             {
-                TreeNode childNode = entry.getValue();
-                ChildData childData = childNode.childData;
-                // Double-check liveness after retreiving data.
-                if ( childData != null && childNode.nodeState == NodeState.LIVE )
+                ChildData childData = entry.getValue().childData;
+                // Double-check liveness after retrieving data.
+                if ( isLive(childData) )
```
dragonsinth: It's subtle, but the point I'm trying to make is that you have to read `.childData` exactly once, then check the liveness, then use only that value going forward.  It would be an error to remove the local variable here and reference `entry.getValue().childData` twice.

It's a "double" check because _generally_ there shouldn't be any non-LIVE children in the parent `children` map that we're iterating.  But due to volatility we have to "double check" that child nodes read from the map are still alive after pulling `.childData`

PR #253, opened 2018-02-27
Change visible of constructor/fields

Alykoff:
Change visible for `ServiceInstanceBuilder` constructor/fields. This is need for extensibility (for example in project `https://github.com/hazelcast/hazelcast-zookeeper`). I think that package visibility is bad for extensibility. Could you please also say your opinion about change package visibility in others classes?

PR #254, opened 2018-02-28
Change visible of constructor/fields

Alykoff:
Change visible for ServiceInstanceBuilder constructor/fields. This is need for extensibility (for example in project https://github.com/hazelcast/hazelcast-zookeeper). I think that package visibility is bad for extensibility. Could you please also say your opinion about change package visibility in others classes?

Current pull request matches with https://github.com/apache/curator/pull/253 and has the issue number.


tisonkun:
@Alykoff is this PR still valid for your use case in hazelcase-zookeeper? Could you share a concrete use case then?

Alykoff:
> @Alykoff is this PR still valid for your use case in hazelcast-zookeeper? Could you share a concrete use case then?

There was valid in some project, but now I don't have access to that project, so I don't have information is it valid or not.

tisonkun:
@Alykoff thanks for your updates :)

PR #256, opened 2018-03-04
CURATOR-457. Extension point for discovering paths.

Me1kaa:


curator-x-discovery/src/test/java/org/apache/curator/x/discovery/details/DiscoveryPathConstructorImplTest.java
```diff
@@ -0,0 +1,59 @@
...
+                }
+            }
+        );
+    }
+
+    @Test
+    public void testGetBasePath() throws Exception
+    {
+        Assert.assertEquals(new DiscoveryPathConstructorImpl("/foo/bar").getBasePath(), "/foo/bar");
+        Assert.assertEquals(new DiscoveryPathConstructorImpl("foo/bar").getBasePath(), "/foo/bar");
```
tisonkun: I worried if we should forbid this pattern, said the base dir passed without a leading slash, instead of tolerate it.

It seems somehow informal especially for `foo/bar/` patterns.

tisonkun (changes requested):
Rebased. One comment inline.

Others look good to me.

tisonkun (approved):
Updated to verify `basePath` with `PathUtils` so that we keep consistency instead of using different rules in different code path.

curator-x-discovery/src/main/java/org/apache/curator/x/discovery/DiscoveryPathConstructor.java
```diff
@@ -0,0 +1,49 @@
...
+    String getPathForInstances(String serviceName);
+
+    /**
+     * Return the path where specified instance registered.
+     *
+     * @param serviceName service name
+     * @param instanceId instance id
+     * @return path to concrete instance
+     */
+    String getPathForInstance(String serviceName, String instanceId);
```
kezhuw, line 48: Do we need this ? From javadocs, I concludes to "/base-path/{serviceName}/instance-dir/{instanceId}". Seems that `getPathForInstances` will give us directory for `instanceId` and `ServiceDiscoveryImpl::queryForInstances` depends on this. We need service directory and instance directory, that is all ?

kezhuw (commented):
I support this idea to customize discovery path. But seems that `getPathForInstance` is redundant. IIUC, we need only two directories, one for services and one for service instances. All other paths are constructed from these two.

curator-x-discovery/src/main/java/org/apache/curator/x/discovery/DiscoveryPathConstructor.java
kezhuw: Given what the javadoc says:
1. `getPathForInstances`: Return the path where all instances of the service registered.
2. `getPathForInstance`: Return the path where specified instance registered.

The second can't be an extension point to return arbitrary path. Otherwise, it violates the first.

Also, I don't think this `getPathForInstance` will work. Assumes:

* `getBasePath()` returns `/base-path`.
* `getPathForInstances(serviceName)` returns `/base-path/{serviceName}/instances/`.

Is it possible for `getPathForInstance(serviceName, instanceId)` to return paths other than `/base-path/{serviceName}/instances/{instanceId}` ? If it is possible, how should `Collection<ServiceInstance<T>> queryForInstances(String name)` be implemented ? I expect the current implementation will break.

I think `getPathForInstance(serviceName, instanceId)` should behave same as `ZKPaths.makePath(getPathForInstances(serviceName), instanceId)` for this customization to function correctly. 

> And I'm OK to postpone this ticket unless anyone shows a real-world use case.

The jira already presented a good use case from my perspective. I can continue the rest if above make sense to you and in case you are not available to push forward this. I don't think we should/can expect response from original author given his/her inactivity.

tisonkun: I go through the original ticket an think that it's the implementor's responsibility to implement the method correctly.

The original ticket argues that the default `ZKPaths.makePath` cannot satisfy some cases. And `ServiceDiscoveryImpl` delegates the related methods to the new PathConstructor:

```java
    @Override
    public Collection<String> queryForNames() throws Exception {
        List<String> names = client.getChildren().forPath(pathConstructor.getBasePath());
        return ImmutableList.copyOf(names);
    }
    String pathForName(String serviceName) {
        return pathConstructor.getPathForInstances(serviceName);
    }
    @VisibleForTesting
    String pathForInstance(String serviceName, String instanceId) {
        return pathConstructor.getPathForInstance(serviceName, instanceId);
    }
```

kezhuw: @tisonkun I will open a pull request for discussion.

tisonkun:
This is superseded by https://github.com/apache/curator/pull/473.

PR #257, opened 2018-03-17
CURATOR-458: Fix a bug spotted by error-prone compiler

leventov:
Fix one bug. See http://errorprone.info

### DIFF
--- a/curator-framework/src/main/java/org/apache/curator/framework/schema/Schema.java
+++ b/curator-framework/src/main/java/org/apache/curator/framework/schema/Schema.java
@@ -105,7 +105,7 @@ public class Schema
 
     Schema(String name, Pattern pathRegex, String path, String documentation, SchemaValidator schemaValidator, Allowance ephemeral, Allowance sequential, Allowance watched, boolean canBeDeleted, Map<String, String> metadata)
     {
-        Preconditions.checkNotNull((pathRegex != null) || (path != null), "pathRegex and path cannot both be null");
+        Preconditions.checkArgument((pathRegex != null) || (path != null), "pathRegex and path cannot both be null");
         this.pathRegex = pathRegex;
         this.fixedPath = fixPath(path);
         this.metadata = ImmutableMap.copyOf(Preconditions.checkNotNull(metadata, "metadata cannot be null"));


Randgalt:
Please open an issue to match this PR - see https://cwiki.apache.org/confluence/display/CURATOR/Submitting+Pull+Requests

leventov:
@Randgalt thanks, done

leventov:
@Randgalt what needs to be done to move this PR forward?

curator-framework/src/main/java/org/apache/curator/framework/schema/Schema.java
```diff
@@ -105,7 +105,8 @@ public static SchemaBuilder builderForRecipe(String parentPath)
 
     Schema(String name, Pattern pathRegex, String path, String documentation, SchemaValidator schemaValidator, Allowance ephemeral, Allowance sequential, Allowance watched, boolean canBeDeleted, Map<String, String> metadata)
     {
-        Preconditions.checkNotNull((pathRegex != null) || (path != null), "pathRegex and path cannot both be null");
+        Preconditions.checkNotNull(pathRegex, "pathRegex cannot both be null");
+        Preconditions.checkNotNull(path, "path cannot both be null");
         this.pathRegex = pathRegex;
```
cammckenzie: The existing check here seems like it was incorrect, as it would have returned a boolean which would never be null. This change will fix it, but there error messages are now incorrect. They should just be "pathRegex cannot be null" and "path cannot be null" respectively.

Randgalt:
I don't mind adding a linter to our build but why should we use this one over findbugs or PMD?

leventov:
@Randgalt all static analysis tools are unique in terms of what problems do they find. Errorprone finds some problems that findbugs is not aware of, and vice versa. In practice, I found that errorprone has generally much better signal / noise ratio than findbugs.

Randgalt:
Before we add a new dependency to Curator we should probably vote on it. I'm out of town, I'll look more into this when I get back.

curator-framework/src/main/java/org/apache/curator/framework/schema/Schema.java
leventov: Actually my original fix was wrong logically. What should have been done is just changing `Preconditions.checkNotNull()` call with `Preconditions.checkArgument()`. Fixed now.

leventov:
@Randgalt @cammckenzie I've changed the scope of this issue to just fixing the bug, without introducing the error-prone dependency.

leventov:
@Randgalt @cammckenzie could you please revisit this PR?

cammckenzie:
Any thoughts on this @Randgalt . I've had a look at it, and the changes seem fine, but equally, they're not adding a great deal (other than stopping some error prone complaints).

Randgalt:
Yeah - I don't really see the need for this.

leventov:
This PR fixes one bug.

Randgalt:
> This PR fixes one bug.

I'd rather see a PR with just the bug fix and not coupled with these other changes.

leventov:
@Randgalt changed.

tisonkun (approved):
LGTM. Although, I notice that the usage of `fixedPath` and `pathRegex` may produce NPE in code, which you can give an overall review later.

cc @eolivelli @Randgalt shall we merge this PR?

tisonkun:
Merging...

tisonkun:
> may produce NPE in code

That said, `hashCode` and `equals` may produce NPE.

PR #260, opened 2018-03-18
replace raw Exception thrown by CompressionProvider.compress and decompress with IOException

krichter722:
Throwing a raw `Exception` almost always indicates a design error (there's a lot of them in the project). It makes sense to replace `throws Exception` with the real exceptions thrown over time. This is the first stop.

PRs like #78 indicate that strange if not wrong understanding of exceptions and their treatment in `catch` blocks. Except for some example classes with `main` methods you never want to do something like `catch(Exception ex) {ex.printStackTrace()}`. If `Callable.call` is involved throwing `Exception` is a very bad approach as well and the fact that it throws `Exception` is no reason to throw it from the method calling `Callable.call`. In this case the exception needs to be wrapped into a custom type.

### DIFF
--- a/curator-framework/src/main/java/org/apache/curator/framework/api/CompressionProvider.java
+++ b/curator-framework/src/main/java/org/apache/curator/framework/api/CompressionProvider.java
@@ -18,9 +18,11 @@
  */
 package org.apache.curator.framework.api;
 
+import java.io.IOException;
+
 public interface CompressionProvider
 {
-    public byte[]       compress(String path, byte[] data) throws Exception;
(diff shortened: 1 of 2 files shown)


Randgalt:
Actually,m no it isn't a design error and is intentional. I'm -1 on this PR. I suggest it's better to discuss design issues on the Curator @dev mailing list. Also, I strongly suggest that you read the Apache Code of Conduct (see here: http://www.apache.org/foundation/policies/conduct.html).

tisonkun:
Closed as reason explained above.

PR #263, opened 2018-04-10
[CURATOR-462] InterProcessSemaphoreV2 leaves orphaned lease node if acquiring the semaphore is interrupted at the wrong time

krajcsovszkig-ms:


cammckenzie:
Looks good to me, will merge shortly.

PR #264, opened 2018-05-21
fix typo

zyfworks:


Randgalt:
`iff` means "if and only iff" - not a typo

PR #267, opened 2018-06-11
CURATOR-448 - Include curator framework state in error messages

cjmctague:


curator-framework/src/main/java/org/apache/curator/framework/imps/CuratorFrameworkImpl.java
```diff
@@ -402,75 +402,72 @@ public String getNamespace()
         return (str != null) ? str : "";
     }
 
+    private void checkState()
+    {
+        CuratorFrameworkState state = getState();
+        Preconditions.checkState(state == CuratorFrameworkState.STARTED, "instance state must be %s before calling this method, however its %s", CuratorFrameworkState.STARTED, state);
```
keith-turner: Could make it less wordy, like : `Expected state %s, however it was %s`

PR #270, opened 2018-07-14
A fix for double barrier timeout behavior, and an accompanying test

htuy42:
Assuming this is the desired behavior (ie once we timeout we are no longer considered to be in the barrier), this fix will work. By the time the method returns false, we will no longer be considered a member waiting at the barrier,


Randgalt:
I apologize that no one comment on this in so long... Also, please create an issue in our Jira for this. Please read here: https://cwiki.apache.org/confluence/display/CURATOR/Submitting+Pull+Requests

tisonkun:
Closed as inactive. Feel free to create a JIRA ticket and resubmit this patch.

PR #272, opened 2018-08-05
[CURATOR-473] update the comment of AdvancedTracerDriver.java

nickChenyx:
previous comment was outdated.


curator-client/src/main/java/org/apache/curator/drivers/AdvancedTracerDriver.java
```diff
@@ -26,17 +26,16 @@
 public abstract class AdvancedTracerDriver implements TracerDriver
 {
     /**
-     * Record the given trace event
+     * Record the given operation trace
```
tisonkun, line 29: I think `OperationTrace` can be regarded as "trace event".

tisonkun (commented):
Thanks for your contribution @nickChenyx ! Comments inline.

Also if you can update deprecation comments for the deprecated methods in this file, it will be great.

tisonkun:
Closed as inactive.

PR #275, opened 2018-08-10
[CURATOR-476] PathChildrenCache should check resultCode=-101 for getData(...) done on non-existent child and remove it from it's initialSet

rsouhrain:


cammckenzie:
Thanks for the PR. Looks good to me. I will merge to master shortly.

PR #276, opened 2018-08-17
recipes-locks-LockInternals.internalLockLoop

zcjhwl:
Why not use unsafe to suspend and release threads?
You use wait and notifyAll instead


tisonkun:
Closed as invalid. Please open an JIRA ticket or send an email to the mailing list to describe your motivation first.

PR #277, opened 2018-08-19
[CURATOR-477] add support for configuring Zk Watches

ramaraochavali:
In our use case, we use `TreeCache` to get Zk Data periodically. We start `TreeCache` read data and close it. In this use case, The `ZkWatchManager` of `ZooKeeper` class keeps growing for every TreeCache operation because new `TreeNode` objects are created and added there leading to a memory leak. Also since we do not want the Watcher to periodically watch, this creates  unnecessary background operations.

In this PR, Made the createZkWatches configurable in the Builder, so that use cases like ours can set it to false. The default is true, retaining the current behaviour.

### DIFF
--- a/curator-framework/src/main/java/org/apache/curator/framework/CuratorFrameworkFactory.java
+++ b/curator-framework/src/main/java/org/apache/curator/framework/CuratorFrameworkFactory.java
@@ -149,6 +149,7 @@ public class CuratorFrameworkFactory
         private SchemaSet schemaSet = SchemaSet.getDefaultSchemaSet();
         private boolean zk34CompatibilityMode = isZK34();
         private int waitForShutdownTimeoutMs = 0;
+        private boolean createZkWatches = true;
         /**
          * Apply the current values and build a new CuratorFramework
          *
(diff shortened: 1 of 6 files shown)


Randgalt:
Please open a Jira issue for this. See here - https://cwiki.apache.org/confluence/display/CURATOR/Submitting+Pull+Requests - then maybe @dragonsinth can take a look at this.

dragonsinth:
@Randgalt the use case makes sense, but there's actually no TreeCache changes in this PR, it's all Framework stuff.

ramaraochavali:
@Randgalt Created JIRA issue https://issues.apache.org/jira/browse/CURATOR-477. Please let me know if any thing else needs to be done. Please review and let me know if any changes are required.

ramaraochavali:
@Randgalt  @dragonsinth on a second thought, I think just cleaner to it at the `TreeCache` level. So closing this PR and opening a fresh PR for it with correct branch name as well. The correct PR is https://github.com/apache/curator/pull/278

PR #278, opened 2018-08-20
[CURATOR-477] added support for turning off zk watches

ramaraochavali:
In our use case, we use `TreeCache` to get Zk Data periodically. We start `TreeCache` read data and close it. In this use case, The `ZkWatchManager` of `ZooKeeper` class keeps growing for every TreeCache operation because new `TreeNode` objects are created and added there leading to a memory leak. Also since we do not want the Watcher to periodically watch, this creates  unnecessary background operations.

In this PR, Made the createZkWatches configurable in the Builder, so that use cases like ours can set it to false. The default is true, retaining the current behaviour.
Fixes the jira issue https://issues.apache.org/jira/browse/CURATOR-477

### DIFF
--- a/curator-recipes/src/main/java/org/apache/curator/framework/recipes/cache/TreeCache.java
+++ b/curator-recipes/src/main/java/org/apache/curator/framework/recipes/cache/TreeCache.java
@@ -347,7 +347,11 @@ public class TreeCache implements Closeable
             if ( parent == null )
             {
                 // Root node; use an exist query to watch for existence.
-                client.checkExists().usingWatcher(this).inBackground(this).forPath(path);
+                if (disableZkWatches) {
+                    client.checkExists().forPath(path);
+                } else {
+                    client.checkExists().usingWatcher(this).inBackground(this).forPath(path);
+                }
             }
             else
             {
--- a/curator-recipes/src/main/java/org/apache/curator/framework/recipes/cache/TreeCache.java
+++ b/curator-recipes/src/main/java/org/apache/curator/framework/recipes/cache/TreeCache.java
@@ -347,11 +347,7 @@ public class TreeCache implements Closeable
             if ( parent == null )
             {
                 // Root node; use an exist query to watch for existence.
-                if (disableZkWatches) {
-                    client.checkExists().forPath(path);
-                } else {
-                    client.checkExists().usingWatcher(this).inBackground(this).forPath(path);
-                }
+                maybeWatch(client.checkExists()).forPath(path);
             }
             else
             {
--- a/curator-recipes/src/main/java/org/apache/curator/framework/recipes/cache/TreeCache.java
+++ b/curator-recipes/src/main/java/org/apache/curator/framework/recipes/cache/TreeCache.java
@@ -19,28 +19,11 @@
 
 package org.apache.curator.framework.recipes.cache;
 
-import static com.google.common.base.Preconditions.checkNotNull;
-import static org.apache.curator.utils.PathUtils.validatePath;
-
 import com.google.common.annotations.VisibleForTesting;
 import com.google.common.base.Function;
 import com.google.common.base.Preconditions;
 import com.google.common.collect.ImmutableMap;
 import com.google.common.collect.Maps;
-import java.io.Closeable;
-import java.util.ArrayList;
-import java.util.Collections;
-import java.util.LinkedList;
-import java.util.List;
-import java.util.Map;
-import java.util.concurrent.ConcurrentMap;
-import java.util.concurrent.ExecutorService;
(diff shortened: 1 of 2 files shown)


curator-recipes/src/main/java/org/apache/curator/framework/recipes/cache/TreeCache.java
```diff
@@ -263,11 +281,20 @@ private void doRefreshData() throws Exception
             {
                 if ( dataIsCompressed )
                 {
-                    client.getData().decompressed().usingWatcher(this).inBackground(this).forPath(path);
+                    if (createZkWatches) {
+                        client.getData().decompressed().usingWatcher(this).inBackground(this).forPath(path);
+                    } else {
+                        client.getData().decompressed().inBackground(this).forPath(path);
+                    }
```
dragonsinth: feel like this could all be a little cleaner with a maybeWatch() helper in the class, e.g.

```
maybeWatch(client.getData().decompressed()).inBackground(this).forPath(path)
```

curator-recipes/src/main/java/org/apache/curator/framework/recipes/cache/TreeCache.java
```diff
@@ -84,6 +85,7 @@
         private ExecutorService executorService = null;
         private int maxDepth = Integer.MAX_VALUE;
         private boolean createParentNodes = false;
+        private boolean createZkWatches = true;
```
dragonsinth: I'd probably reverse this as "disableZkWatch" since that's the unusual case.

curator-recipes/src/main/java/org/apache/curator/framework/recipes/cache/TreeCache.java
```diff
@@ -165,6 +167,18 @@ public Builder setCreateParentNodes(boolean createParentNodes)
...
+         * By default, TreeCache creates {@link org.apache.zookeeper.ZooKeeper} watches for every created path.
+         * Change this behavior with this method.
+         * @param createZkWatches true to create watches
+         * @return this for chaining
+         */
+        public Builder setCreateZkWatches(boolean createZkWatches)
+        {
+            this.createZkWatches = createZkWatches;
+            return this;
+        }
```
dragonsinth: and e.g. disableZkWatches(boolean)

ramaraochavali:
@dragonsinth Thanks for your review. addressed the feedback. PTAL

curator-recipes/src/main/java/org/apache/curator/framework/recipes/cache/TreeCache.java
```diff
@@ -263,15 +284,31 @@ private void doRefreshData() throws Exception
...
+
+        private ErrorListenerPathable<byte[]> maybeWatch(GetDataBuilder dataBuilder) {
+            if (disableZkWatches) {
+                return dataBuilder.inBackground(this);
+            } else {
+                return dataBuilder.usingWatcher(this).inBackground(this);
+            }
+        }
+
```
dragonsinth:
```java
        private void doRefreshChildren() throws Exception
        {
            if ( treeState.get() == TreeState.STARTED )
            {
                maybeWatch(client.getChildren()).forPath(path);
            }
        }

        private void doRefreshData() throws Exception
        {
            if ( treeState.get() == TreeState.STARTED )
            {
                if ( dataIsCompressed )
                {
                    maybeWatch(client.getData().decompressed()).forPath(path);
                }
                else
                {
                    maybeWatch(client.getData()).forPath(path);
                }
            }
        }

        private <T, P extends Watchable<BackgroundPathable<T>> & BackgroundPathable<T>> Pathable<T> maybeWatch(P dataBuilder)
        {
            if ( disableZkWatches )
            {
                return dataBuilder.inBackground(this);
            }
            else
            {
                return dataBuilder.usingWatcher(this).inBackground(this);
            }
        }
```

dragonsinth:
TestTreeCache is getting stuck on your branch.  Haven't figured out why yet.  Wanna take a look?

ramaraochavali:
Sorry. My bad. Problem with my change. Fixed. PTAL

dragonsinth:
aha that would do it!  I do think you should update the `maybeWatch` generic method per my other comment

ramaraochavali:
Ahh. Sorry. I missed that comment. I have changed it now. PTAL.

dragonsinth:
@Randgalt LGTM, although I haven't rerun tests yet.

ramaraochavali:
@dragonsinth sorry. Added one more commit for exists check also. can you PTAL?

cammckenzie:
I will merge this shortly, thanks @ramaraochavali and @dragonsinth

PR #281, opened 2018-11-03
[CURATOR-483] Fix path used when re-creating sequential PersistentNode with protection

njhill:
Would previously result in creation of a second non-sequential znode.

See https://issues.apache.org/jira/browse/CURATOR-483.

### DIFF
--- a/curator-recipes/src/main/java/org/apache/curator/framework/recipes/nodes/PersistentNode.java
+++ b/curator-recipes/src/main/java/org/apache/curator/framework/recipes/nodes/PersistentNode.java
@@ -445,7 +445,8 @@ public class PersistentNode implements Closeable
         try
         {
             String existingPath = nodePath.get();
-            String createPath = (existingPath != null && !useProtection) ? existingPath : basePath;
+            String createPath = existingPath == null || (useProtection && !isSequential(mode)) ? basePath
+                    : (useProtection ? basePath + existingPath.substring(existingPath.length()-10) : existingPath);
 
             CreateModable<ACLBackgroundPathAndBytesable<String>> localCreateMethod = createMethod.get();
             if ( localCreateMethod == null )
@@ -462,6 +463,10 @@ public class PersistentNode implements Closeable
             ThreadUtils.checkInterrupted(e);
             throw new RuntimeException("Creating node. BasePath: " + basePath, e);  // should never happen unless there's a programming error - so throw RuntimeException
         }
+    }
+    
+    private static boolean isSequential(CreateMode mode) {
+        return mode == CreateMode.EPHEMERAL_SEQUENTIAL || mode == CreateMode.PERSISTENT_SEQUENTIAL;
     }
 
     private CreateMode getCreateMode(boolean pathIsSet)


cammckenzie:
@njhill Thanks for the PR. Would it be possible to provide a unit test to reproduce the problem and verify the fix?
cheers

njhill:
@cammckenzie sure, I've now added one

curator-recipes/src/main/java/org/apache/curator/framework/recipes/nodes/PersistentNode.java
```diff
@@ -445,7 +445,8 @@ private void createNode()
         try
         {
             String existingPath = nodePath.get();
-            String createPath = (existingPath != null && !useProtection) ? existingPath : basePath;
+            String createPath = existingPath == null || (useProtection && !isSequential(mode)) ? basePath
+                    : (useProtection ? basePath + existingPath.substring(existingPath.length()-10) : existingPath);
```
cammckenzie: I think that this would be cleaner as some if statements. Also, the 10 magic number is not ideal, but I don't believe that we have a constant defining the length of the sequential part of the node name anywhere. Not sure where the best place would be, perhaps just in the CuratorFramework interface. Thoughts @Randgalt ?

njhill: Thanks @cammckenzie. I reworked this to use if statements and defined a constant for the 10 number. I couldn't find such a constant defined elsewhere; it's hardcoded in the zookeeper code: https://github.com/apache/zookeeper/blob/master/zookeeper-server/src/main/java/org/apache/zookeeper/server/PrepRequestProcessor.java#L662.

I also removed the redundant added static `isSequential(CreateMode)` method, didn't notice that this is already a method on the enum.

curator-recipes/src/main/java/org/apache/curator/framework/recipes/nodes/PersistentNode.java
```diff
@@ -444,8 +447,19 @@ private void createNode()
...
+            if ( existingPath != null && !useProtection )
+            {
+                createPath = existingPath;
+            }
+            else
+            {
+                createPath = basePath;
+                if ( useProtection && mode.isSequential() )
+                {
+                    createPath += existingPath.substring(existingPath.length()-SEQUENTIAL_SUFFIX_DIGITS);
```
cammckenzie: This logic is now incorrect. When it's first run the existingPath will be null.

njhill: @cammckenzie :man_facepalming: apologies, now fixed!

curator-recipes/src/main/java/org/apache/curator/framework/recipes/nodes/PersistentNode.java
Randgalt: This might be a good candidate to add to `ZPath.java` - i.e. a method to pull out the sequential number

cammckenzie: Thanks @Randgalt, I figured that there was a good place to put this, but couldn't remember where it should be. @njhill can you refactor the sequential number extraction into org.apache.curator.utils.ZKPaths class? Or, if you don't have time, let me know and I'll get it done.

njhill: @Randgalt @cammckenzie sure, I can do that but am wondering about the best semantics for such a utility method. In particular should it first verify that the path ends in 10 digits? (even in the case that it does, I don't think it's guaranteed to be a sequential node). If so and it does not end in 10 digits, what should be the behaviour?

For this scenario in `PersistentNode` we can be "sure" that it's sequential and hence it's probably reasonable (and more efficient) to omit the check. Given that curator has lasted this long without the need for such a method, I'm wondering whether having it is really warranted. Might it be better to wait until N > 1 (if that ever happens)?

cammckenzie: Good question @njhill  I don't really have a super strong preference, but I'd prefer that the constant defining the size of the sequential bit is not in the PersisteNode class. Perhaps just have a method in ZKPaths that extracts the last 10 digits without any checks. We can make it more defensive down the track if it gets used elsewhere.

I don't think that efficiency is a huge concern here as it's only going to happen when the node is created.

njhill: @cammckenzie @Randgalt I've now updated it per the suggestions, PTAL!

njhill:
@cammckenzie @Randgalt still interested in this fix? :)

curator-recipes/src/main/java/org/apache/curator/framework/recipes/nodes/PersistentNode.java
```diff
@@ -422,6 +422,9 @@ protected void deleteNode() throws Exception
         }
     }
 
+    // Hardcoded in {@link org.apache.zookeeper.server.PrepRequestProcessor}
+    static final int SEQUENTIAL_SUFFIX_DIGITS = 10;
```
Randgalt: I thought we were going to put this in ZKPaths?

Randgalt: Did you push your changes? The PR should update. This is odd.

Randgalt: Wait - I'm seeing it now.

Randgalt:
I'm OK with this now.

cammckenzie:
Yep, I've been meaning to merge, just having issues running some unit tests (which are unrelated to this PR). I will try and merge it today.

njhill:
Great, thanks @cammckenzie @Randgalt. And I just noticed something odd which may be contributing to the confusion ... at least for me right now the commits on the Commits tab are showing in the opposite order to normal i.e. oldest one at the top...

cammckenzie:
Thanks for the fix @njhill

PR #283, opened 2018-12-01
CURATOR-472 - fix deadlocks during tests as well as TTL test issues

Randgalt:
Fixed 2 problems: 1) internalClose can be called from a ZooKeeper background thread (as part of a Watcher callback). Calling ZooKeeper.close with a waitForShutdownTimeoutMs in this case will cause a deadlock as no other events can be processed until the current thread exits. 2) All TTL tests must set the system property "zookeeper.extendedTypesEnabled"

Note: all tests pass for me now


cammckenzie:
Looks good to me. I pushed a fixed to the TestTtlNodes unit test is it was consistently failing for me.

PR #284, opened 2018-12-04
CURATOR-489: CreateBuilderImpl assigns protectedId if doProtected is …

joshng:
This is a light change to correct a bug when `doProtected` is true: the existing code sets `CreateBuilderImpl.doProtected = true`, but assigns `protectedId = null`, resulting in a non-unique protection-token (`_c_null-`) being applied to the name created in zookeeper. Given that all replicas coordinating via zookeeper would use the identical token, this could conceivably lead to incorrect behavior when the protection logic is needed.

While I believe this PR fixes the bug, I'd really suggest a more comprehensive change to eliminate the `CreateBuilderImpl.doProtected` field entirely, instead relying upon `protectedId != null` to convey this intention. However, being unfamiliar with the nuances of the existing implementation, I didn't want to perform that surgery today. If you think that change seems preferable, I'd be happy to do that instead.


Randgalt:
> suggest a more comprehensive change to eliminate the CreateBuilderImpl.doProtected field entirely

Maybe a separate issue for that. 

Do you mind writing a test for this please?

joshng:
Ok, that test is incorporated here now. 👍

PR #279, opened 2018-09-18
[CURATOR-468] release acquired leadership mutex properly

udaysagar2177:
This fixes https://issues.apache.org/jira/browse/CURATOR-468. This PR is based on a solution suggested by cammckenzie & kturner on the jira issue.

PR #285, opened 2018-12-06
[CURATOR-468] LeaderSelector interruption causing spurious log messages, etc.

Randgalt:
Adds a test to what #279 had along with a slightly more complete fix. The finally clause of doWork() releases the mutex. In an interrupted situation this would always fail causing the failed delete manager to handle the work resulting in slightly slower deletions and spurious log messaged. This fixes this by clearing the thread interrupted state prior to releasing the mutex and then resetting afterwards if needed.


cammckenzie:
Looks ok to me.

PR #291, opened 2018-12-09
[CURATOR-493] Fix test case with testDeleteChildrenConcurrently

hebelala:
Link #235

### DIFF
--- a/curator-framework/src/test/java/org/apache/curator/framework/imps/TestFrameworkEdges.java
+++ b/curator-framework/src/test/java/org/apache/curator/framework/imps/TestFrameworkEdges.java
@@ -45,10 +45,13 @@ import org.apache.zookeeper.KeeperException;
 import org.apache.zookeeper.WatchedEvent;
 import org.apache.zookeeper.Watcher;
 import org.apache.zookeeper.data.Stat;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 import org.testng.Assert;
 import org.testng.annotations.BeforeClass;
 import org.testng.annotations.Test;
 import java.util.List;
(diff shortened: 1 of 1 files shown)


curator-framework/src/test/java/org/apache/curator/framework/imps/TestFrameworkEdges.java
```diff
@@ -686,35 +693,66 @@ public void run()
...
+                try
                 {
-                    Assert.fail("client2 delete failed, shouldn't throw NoNodeException", e);
+                    if ( !threadDeleted )
+                    {
+                        Stat stat = client2.checkExists().forPath(child);
+                        if ( stat == null )
+                        {
+                            // the thread client has begin deleted the children
+                            threadDeleted = true;
```
Randgalt: What's the point of this? Once it is set, checkExists() is never called again. That seems incorrect. I removed this and the test still passes. Can you explain?

hebelala: The point is that judge the new thread is running at the loop that is deleting children:      
```ZKPaths.java``` , 341 line
```
        for ( String child : children )
        {
            String fullPath = makePath(path, child);
            deleteChildren(zookeeper, fullPath, true);
        }
```

and then, the main thread is try to delete some child that is still not deleted by the new thread. If the main thread delete child successfully, the new thread should not throw `NoNodeException`, but the old version will throw NoNodeException and cause the remaining children are not deleted, that's the bug situation.

So, it's unnecessary to checkExists() after that `threadDeleted` is true. I just want the main thread delete a child successfully, when the new thread is deleting children.

PR #292, opened 2018-12-09
[CURATOR-405] reset startOfSuspendedEpoch when session expiration is injected

Randgalt:
Once the session expiration is injected, reset startOfSuspendedEpoch so that we don't do this copious times while things are being reset.


Randgalt:
I ran the tests a few times and they all pass.

PR #293, opened 2018-12-09
[CURATOR-490] Make the tests more resilient

Randgalt:
Some of the tests are flakey and fail intermittently. This task will remain open as a container for fixing tests.


Randgalt:
At this point, tests are passing but we still sometimes get at the end of `curator-test-zk34`. I've opened a bug with Maven Surefire for it: https://issues.apache.org/jira/browse/SUREFIRE-1612

PR #297, opened 2018-12-14
[CURATOR-495] Fixes race in many Curator recipes which could block ZK's event thread

Randgalt:
Fixes race in many Curator recipes whereby a pattern was used that called "notifyAll" in a synchronized block in response to a ZooKeeper watcher callback. This created a race and possible deadlock if the recipe instance was already in a synchronized block. This would result in the ZK event thread getting blocked which would prevent ZK connections from getting repaired. This change adds a new executor (available from CuratorFramework) that can be used to do the sync/notify so that ZK's event thread is not blocked.

### DIFF
--- a/curator-framework/src/main/java/org/apache/curator/framework/CuratorFramework.java
+++ b/curator-framework/src/main/java/org/apache/curator/framework/CuratorFramework.java
@@ -25,16 +25,14 @@ import org.apache.curator.framework.api.transaction.CuratorMultiTransaction;
 import org.apache.curator.framework.api.transaction.CuratorOp;
 import org.apache.curator.framework.api.transaction.CuratorTransaction;
 import org.apache.curator.framework.api.transaction.TransactionOp;
-import org.apache.curator.framework.imps.CuratorFrameworkImpl;
 import org.apache.curator.framework.imps.CuratorFrameworkState;
 import org.apache.curator.framework.listen.Listenable;
 import org.apache.curator.framework.schema.SchemaSet;
-import org.apache.curator.framework.state.ConnectionStateListener;
 import org.apache.curator.framework.state.ConnectionStateErrorPolicy;
+import org.apache.curator.framework.state.ConnectionStateListener;
 import org.apache.curator.utils.EnsurePath;
 import org.apache.zookeeper.Watcher;
 import org.apache.zookeeper.server.quorum.flexible.QuorumVerifier;
-
 import java.io.Closeable;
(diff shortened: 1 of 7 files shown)


curator-framework/src/main/java/org/apache/curator/framework/CuratorFramework.java
```diff
@@ -331,4 +329,29 @@
...
     boolean isZk34CompatibilityMode();
+
+    /**
+     * Calls {@link #notifyAll()} on the given object after first synchronizing on it. This is
+     * done from the {@link #runSafe(Runnable)} thread.
+     *
+     * @param monitorHolder object to sync on and notify
+     * @since 4.1.0
+     */
+    default void postSafeNotify(Object monitorHolder)
```
eolivelli: Thinking about the future...
Would it be better to make it return a CompletableFuture?
This way the caller will have a chance to know when eventually the operation has been executed

Randgalt: Good idea - I'll add that

curator-framework/src/main/java/org/apache/curator/framework/CuratorFrameworkFactory.java
```diff
@@ -474,6 +477,28 @@ public Builder schemaSet(SchemaSet schemaSet)
...
+         * to set a custom executor.
+         *
+         * @param runSafeService executor to use for calls to notifyAll from Watcher callbacks etc
+         * @return this
+         * @since 4.1.0
+         */
+        public Builder runSafeService(Executor runSafeService)
+        {
+            this.runSafeService = runSafeService;
+            return null;
```
cammckenzie: This should return this rather than null

Randgalt: damn - good catch

cammckenzie:
I have just run the tests and everything passed, which is a bit of a novelty! Still failed at the end with that surefire issue. I think that this is good to merge though. Nice work!

Randgalt:
For the record: here’s the Surefire big report: https://issues.apache.org/jira/browse/SUREFIRE-1612

eolivelli:
Did you try to upgrade testng and surefire to latest and greatest version?

Randgalt:
FYI - I just tried it with the latest TestNg and it still fails. So, I re-opened my issue there: https://github.com/cbeust/testng/issues/1976

PR #298, opened 2018-12-25
CURATOR-498 LeaderLatch deletes leader and leaves it hung besides a second leader

shayshim:
The Curator app I am working on uses the LeaderLatch to select a leader out of 6 clients.

While testing my app, I noticed that when I make ZK lose its quorum for a while and then restore it, then after Curator in my app restores it's connection to ZK - sometimes not all the 6 clients are found in the latch path (using zkCli.sh). That is, I have 5 instead of 6.

After investigating a little, I have a suspicion that LeaderLatch deleted the leader in method setNode.

To investigate it I copied the LeaderLatch code and added some log messages, and from them it seems like very old create() background callback was surprisingly scheduled and corrupted the current leader with its stale path name. Meaning, this old one called setNode with its stale name, and set itself instead of the leader and deleted the leader. This leaves client running, thinking it is the leader, while another leader is selected.

If my analysis is correct then it seems like we need to make this obsolete create callback cancelled (I think its session was suspended on 22:38:54 and then lost on 22:39:04 - so on SUSPENDED cancel ongoing callbacks).

Please see attached log file and modified LeaderLatch0.

In the log, note that on 22:39:26 it shows that 0000000485 is replaced by 0000000480 and then probably deleted.

Note also that at 22:38:52, 34 seconds before, we can see that it was in the reset() method ("RESET OUR PATH") and possibly triggered the creation of 0000000480 then.

Environment:  ZooKeeper 3.4.13, Curator 4.1.0 (selecting explicitly 3.4.13), Linux 
[HaWatcher.log](https://github.com/apache/curator/files/2709566/HaWatcher.log)
[LeaderLatch0.java.txt](https://github.com/apache/curator/files/2709569/LeaderLatch0.java.txt)


Randgalt:
This is the most minimalist PR we've ever received :D

shayshim:
Hahaha
I copied above the description from JIRA CURATOR-498. Left there both attached files.

Randgalt:
Seriously - is there meant to be a PR here? I only see a 1 line change of a space.

PR #299, opened 2019-01-02
[CURATOR-498] - Fix protection mode race with ephemeral nodes

Randgalt:
"Protection" has a potential bug. If the connection is lost for long enough, Curator will want to kill the session. Session deletions must be handled by the Leader ZK instance. At the same time that the session kill is being processed, Curator's protection mode handling could be calling the follower that it's connected to get the current list of children - this can be handled directly by the follower instance without needing to call the leader. So, in this scenario, the client will get a list of children that includes the ZNode that will get deleted as part of killing the session.
    
This bug has been in Curator since we added the protection feature to it more than 6 years ago. The fix is to include the session ID in the protection ID that is generated for the node name when the create mode is an ephemeral type. Then, if findProtectedNodeInForeground() finds the node in the use-case we've been discussing, it can compare the session ID to the current ZooKeeper handle's session ID and disregard the found node if they don't match.


eolivelli:
@Randgalt why did you close this PR ?

PR #300, opened 2019-01-10
CURATOR-499 Allow creating sequential nodes with an empty name

marquiswang:
Add CreateModable#withMode(CreateMode, PathEncodingType) where
PathEncodingType is an enum with values DEFAULT and ALLOW_EMPTY.

When a node is created with ALLOW_EMPTY, trailing /s will not be
removed from the path.

This should not be used except with sequential CreateModes, as it will
fail with:
java.lang.IllegalArgumentException: Path must not end with / character

Calling the following:

curatorFramework.create()
    .withMode(CreateMode.PERSISTENT_SEQUENTIAL, PathEncodingType.ALLOW_EMPTY)
    .forPath("/foo/bar/, data);

Will create a node "/foo/bar/000000000X", whereas calling:

curatorFramework.create()
    .withMode(CreateMode.PERSISTENT_SEQUENTIAL)
    .forPath("/foo/bar/, data);

Will create a node "/foo/bar000000000X".

### DIFF
--- a/curator-client/src/test/java/org/apache/curator/utils/TestZKPaths.java
+++ b/curator-client/src/test/java/org/apache/curator/utils/TestZKPaths.java
@@ -26,6 +26,18 @@ import java.util.Collections;
 
 public class TestZKPaths
 {
+    @Test
+    public void testFixForNamespaceWithEmptyNodeName()
+    {
+        Assert.assertEquals(ZKPaths.fixForNamespace("foo", "/bar/", true, false), "/foo/bar");
+    }
+
+    @Test
+    public void testFixForNamespaceWithEmptyNodeNameAllowed()
+    {
+        Assert.assertEquals(ZKPaths.fixForNamespace("foo", "/bar/", true, true), "/foo/bar/");
+    }
+
     @SuppressWarnings("NullArgumentToVariableArgMethod")
     @Test
     public void testMakePath()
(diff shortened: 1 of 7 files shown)


Randgalt:
I'm +1 on this change but do you mind writing more tests in `TestZKPaths`? I'd like to see more complete coverage of the new methods.

curator-client/src/test/java/org/apache/curator/utils/TestZKPaths.java
```diff
@@ -26,6 +26,18 @@
 
 public class TestZKPaths
 {
+    @Test
```
Randgalt: I'm +1 on this change but do you mind writing more tests in `TestZKPaths`? I'd like to see more complete coverage of the new methods.

Randgalt: @marquiswang any chance for more tests?

marquiswang: Sorry, I have been on an extended trip and have not had a chance to work on this. I can definitely add tests when I get home in mid March.

tisonkun: ping @marquiswang do you have some spare time to bring back this patch? You can merge master and add tests :)

marquiswang: Hello, blast from the past. I supppooooose I can look into it when I get a moment.

tisonkun: @marquiswang lol. Go ahead!

tisonkun:
After another look, I actually think that we should fail on `create("/zoo/foo/bar/")` with "malformed path" instead of auto strip at first XD. But it's public user-facing behavior now.

PR #280, opened 2018-10-09
CURATOR-481 Remove jackson-mapper-asl-version and update jackson

Max-Pudov:

PR #301, opened 2019-01-21
[CURATOR-503] Update dependencies in January 2019

leventov:
The main motivation for this update is updating Guava from 20.0 to 27.0.1-jre. Between these versions, the implementation of `Maps.newConcurrentMap()` (this method is used in many places in Curator code) was changed and now it returns `ConcurrentHashMap` instead of Guava's own implementation, that was less efficient.

This PR also updates Jackson version, as well as #280, from 2.7 to 2.9. Answering questions raised in the comments to that PR: I believe upgrading Jackson from 2.7 to 2.9 is safe. The compatibility standards in Jackson are much higher than in an average Java project. Most incompatibilities in [2.8](https://github.com/FasterXML/jackson/wiki/Jackson-Release-2.8#changes-compatibility) and [2.9](https://github.com/FasterXML/jackson/wiki/Jackson-Release-2.9#changes-compatibility) are minor tweaks in Jackson's programmatic API (shouldn't be a concern as long as Curator is built and passes tests with the updated dependencies). Also support for some old Java and Android versions is dropped, shouldn't be a concern in Curator either. Finally, seems that there are the only two changes that affect serialization format: [of `java.nio.Path`](https://github.com/FasterXML/jackson-databind/issues/1235) and [of `java.sql.Date`](https://github.com/FasterXML/jackson-databind/issues/219). It also doesn't seem to be relevant to Curator.

Testing: `mvn test` in a private CI server passed.

### DIFF
--- a/curator-x-discovery-server/pom.xml
+++ b/curator-x-discovery-server/pom.xml
@@ -135,8 +135,8 @@
                 <jdk>[1.9,)</jdk>
             </activation>
             <dependencies>
-                <!-- The following dependencies are added to make tests work with JDK 9, as per this
-                     answer: https://stackoverflow.com/questions/43574426#43574427 -->
+                <!-- The following dependencies are added to make tests work with JDK 9 or later,
+                      as per this answer: https://stackoverflow.com/questions/43574426#43574427 -->
 
                 <dependency>
                     <groupId>javax.xml.bind</groupId>
--- a/curator-x-discovery-server/pom.xml
+++ b/curator-x-discovery-server/pom.xml
@@ -127,4 +127,41 @@
             </plugin>
         </plugins>
     </build>
+
+    <profiles>
+        <profile>
+            <id>jdk-9-plus</id>
+            <activation>
+                <jdk>[1.9,)</jdk>
+            </activation>
+            <dependencies>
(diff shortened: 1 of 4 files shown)


curator-x-discovery-server/pom.xml
```diff
@@ -127,4 +127,41 @@
             </plugin>
         </plugins>
     </build>
+
+    <profiles>
+        <profile>
+            <id>jdk-9-plus</id>
```
Randgalt: I'm -1 on this change. I'd rather see a comprehensive Java 11 PR (we have a Jira for this already). Java 9 is not an LTS release.

leventov: This profile is activated whenever the test runtime is Java 9+ (including Java 11). I could change this to Java 11+, so that Java 9 and 10 are not supported at all, but I don't see much sense in it. In fact, it will break the workflow for people who want to run some tests locally and for whatever reason they happen to have Java 9 or 10, but not Java 11 installed (that was my case).

Randgalt:
FYI - Guava has been broken into multiple JARs it seems. I had to update the Guava shading we do. https://github.com/apache/curator/commit/6c0710429cc4767be207073e182eb7a27ee4edc5

PR #302, opened 2019-01-28
[CURATOR-502] Update guava-version to 27.0.1

dawinter:


Randgalt:
This doesn't compile for me - I get a ton of errors. Also, please rename the PR so that to something like: `[CURATOR-502] Update guava-version to 27.0.1`

eolivelli:
@Randgalt as guava has a lot of issues about forward/backward compatibility, did you ever consider to shade it and relocate packages?

Randgalt:
It already is shaded. see here: https://github.com/apache/curator/blob/master/curator-client/pom.xml#L90

eolivelli:
Ops. Sorry.
I didn't notice.

Randgalt:
TBH - this PR is not needed. But, we keep getting requests for it so I'm OK with making the change. But, it's not trivial. Several parts of the codebase need to be updated.

Randgalt:
Also see here: https://cwiki.apache.org/confluence/display/CURATOR/TN13

eolivelli:
@Randgalt thanks for the pointer.
Is there any roadmap towards breaking the API and drop Guava classes at least from the public surface of the project?

Randgalt:
Not at this time. It would be pretty disruptive and there hasn't been a call for it. But, if someone proposes something it will definitely be considered.

Randgalt:
This can be closed in favor of #301

dawinter:
Sure, everything that's update the version is ok for me.

PR #303, opened 2019-01-28
[CURATOR-498] Protected Mode creation can mistake closing session's node causing problems for many recipes such as LeaderLatch

Randgalt:
Kudos to user Shay Shimony for his tireless and excellent work tracking this down. There are two problems addressed here: 1) Protected create mode can potentially find a ZNode that is about to be deleted due to an expired session. CreateBuilderImpl now keeps track of the session ID when the create is initiated. If after a connection loss the session ID has changed, any found protected node is ignored as it will soon be deleted. 2) For ZooKeeper 3.4.x the simulated (via reflection) InjectSessionExpiration was incorrectly setting the connection state to closed which caused the session expiration to be ignored.


Randgalt:
attn: @shayshim

curator-client/src/main/java/org/apache/curator/utils/InjectSessionExpiration.java
```diff
@@ -94,7 +89,7 @@ public static void injectSessionExpiration(ZooKeeper zooKeeper)
             Object eventThread = eventThreadField.get(clientCnxn);
             queueEventMethod.invoke(eventThread, event);
             queueEventOfDeathMethod.invoke(eventThread);
```
shayshim: Why it is important to queue event of death immediately? It will be done for us soon by this flow:
ConnectionState.process -> handleExpiredSession -> ... -> ClientCnxn.disconnect ->...
- because we just queued this WatchedEvent.
Same thing for wakeupCnxnMethod.invoke.

Randgalt: TBH I just wanted to make the minimal change needed to get this to work. It's been this way a very long time.

curator-framework/src/main/java/org/apache/curator/framework/imps/CreateBuilderImpl.java
```diff
@@ -48,19 +50,21 @@
...
     private boolean compress;
     private boolean setDataIfExists;
     private int setDataIfExistsVersion = -1;
-    private String protectedId;
     private ACLing acling;
     private Stat storingStat;
     private long ttl;
+    private boolean doProtected;
+    private String protectedId;
+    private long initialSessionId;
```
cammckenzie: I think that maybe the initialSessionId should have a different name as it's a bit misleading. It's not really the initialSessionId as it gets updated during the edge case for handling protected ephemeral nodes. Really, it's the session ID at the time of creation for a protected ephemeral node isn't it? Perhaps it should be named accordingly and only initialised for the case where it will be needed?

Randgalt: Sure - can you suggest a name?

Randgalt:
It's odd - only 1 test is failing: TestTreeCache.testKilledSession(). I can make it work by special casing via `if ( client.isZk34CompatibilityMode() )` but I wonder why? ZK is sending a node deletion immediately after the fake session closed event is posted. I'm not sure it's worth debugging.

curator-framework/src/main/java/org/apache/curator/framework/imps/CreateBuilderImpl.java
cammckenzie: protectedEphemeralSessionID? And then maybe move the initialisation check so that it only occurs if Curator is actually creating a protected ephemeral node, rather than always initialising on the first forPath() call?

Randgalt:
@cammckenzie have a look at this new version. I refactored protected mode into a separate class. Also, everyone, have a look at TestTreeCache.testKilledSession - it makes me a bit nervous that this had to be done but it's only for ZK 3.4 and may not be worth the effort to fix.

curator-framework/src/main/java/org/apache/curator/framework/imps/CreateBuilderImpl.java
```diff
@@ -610,18 +610,18 @@ private String protectedPathInForeground(String adjustedPath, byte[] data, List<
         {
             ThreadUtils.checkInterrupted(e);
             if ( ( e instanceof KeeperException.ConnectionLossException ||
-                !( e instanceof KeeperException )) && protectedId != null )
+                !( e instanceof KeeperException )) && protectedMode.doProtected() )
```
Randgalt: Note: this is a subtle change. The previous version checked `protectedId != null` and I don't remember why.

curator-recipes/src/test/java/org/apache/curator/framework/recipes/cache/TestTreeCache.java
```diff
@@ -426,10 +426,20 @@ public void testKilledSession() throws Exception
         assertEvent(TreeCacheEvent.Type.NODE_ADDED, "/test/me");
 
         KillSession2.kill(client.getZookeeperClient().getZooKeeper());
-        assertEvent(TreeCacheEvent.Type.CONNECTION_LOST);
-        assertEvent(TreeCacheEvent.Type.CONNECTION_RECONNECTED);
-        assertEvent(TreeCacheEvent.Type.INITIALIZED);
-        assertEvent(TreeCacheEvent.Type.NODE_REMOVED, "/test/me", "data".getBytes());
+        if ( client.isZk34CompatibilityMode() )
```
cammckenzie: Is this a race condition? I reverted this test and it seems to run fine for me (the 5 times I've run it)

shayshim:
For ZK 3.5, ZooKeeperTestable is used to do similar work like InjectSessionExpiration does for 3.4 (or maybe vise versa). And similarly to the bug in InjectSessionExpiration, ZooKeeperTestable also sets ClintCnxn.state to CLOSED, which I believe would lead to the same bug in Curator when 3.5 is used. 
Maybe better to use the reflective version also for 3.5 as well, and not ZooKeeperTestable.

curator-framework/src/main/java/org/apache/curator/framework/imps/ProtectedMode.java
```diff
@@ -0,0 +1,112 @@
...
+/**
+ * manage the protected mode state for {@link org.apache.curator.framework.imps.CreateBuilderImpl}
+ */
+class ProtectedMode
+{
+    private final Logger log = LoggerFactory.getLogger(getClass());
+    private boolean doProtected = false;
+    private String protectedId = null;
+    private long sessionId = 0;
+
```
shayshim: ProtectedMode is accessed by potentially more than one thread, no? In that case, doesn't it need to use some concurrency synchronization mechanism like a lock or volatile for it's fields, at least to assure visibility?

curator-framework/src/main/java/org/apache/curator/framework/imps/ProtectedMode.java
```diff
@@ -0,0 +1,112 @@
...
+     * @param createMode create mode in use
+     * @param foundNode the found node
+     * @return either the found node or null - client should always use the returned value
+     * @throws Exception errors
+     */
+    String validateFoundNode(CuratorFrameworkImpl client, CreateMode createMode, String foundNode) throws Exception
+    {
+        if ( doProtected && createMode.isEphemeral() )
+        {
+            if ( sessionId != client.getZooKeeper().getSessionId() )
```
shayshim: It is hard to believe that the session would be changed between lines 98, 100 & 108 (and even if it does - it still seems to be correct), so it is mostly matter of style to prefer saving the result of getSessionId to local variable and continue all local work against that variable.

curator-recipes/src/test/java/org/apache/curator/framework/recipes/cache/TestTreeCache.java
Randgalt: I'm going to re-work this test

Randgalt:
@shayshim and @cammckenzie - I've applied your comments and refactored some code, etc. Please re-review. Note: I'm now doing as Shay suggests and only inserted a session expired event for both 3.4 and 3.5. This means we can no longer use the Testable facade I added to ZK long ago. For future safety, I'm asking ZK to help us out via: https://issues.apache.org/jira/browse/ZOOKEEPER-3269

cammckenzie:
@Randgalt the changes seem OK, but the tests for various things are now failing. Discovery for example:

[INFO] -------------------------------------------------------
[INFO]  T E S T S
[INFO] -------------------------------------------------------
[INFO] Running org.apache.curator.x.discovery.server.jetty_jersey.TestMapsWithJersey
[ERROR] Tests run: 1, Failures: 1, Errors: 0, Skipped: 0, Time elapsed: 1.391 s <<< FAILURE! - in org.apache.curator.x.discovery.server.jetty_jersey.TestMapsWithJersey
[ERROR] testRegisterService(org.apache.curator.x.discovery.server.jetty_jersey.TestMapsWithJersey)  Time elapsed: 0.349 s  <<< FAILURE!
com.sun.jersey.api.client.UniformInterfaceException: PUT http://localhost:42063/v1/service/test/b290e974-b95c-486a-9cf9-487177b19b1f returned a response status of 404 Not Found
	at org.apache.curator.x.discovery.server.jetty_jersey.TestMapsWithJersey.testRegisterService(TestMapsWithJersey.java:136)

[INFO] Running org.apache.curator.x.discovery.server.jetty_jersey.TestObjectPayloadWithJersey
[ERROR] Tests run: 1, Failures: 1, Errors: 0, Skipped: 0, Time elapsed: 1.363 s <<< FAILURE! - in org.apache.curator.x.discovery.server.jetty_jersey.TestObjectPayloadWithJersey
[ERROR] testRegisterService(org.apache.curator.x.discovery.server.jetty_jersey.TestObjectPayloadWithJersey)  Time elapsed: 0.351 s  <<< FAILURE!
com.sun.jersey.api.client.UniformInterfaceException: PUT http://localhost:37965/v1/service/test/38a4bed0-50a5-43c9-b214-4eaa96757f45 returned a response status of 404 Not Found
	at org.apache.curator.x.discovery.server.jetty_jersey.TestObjectPayloadWithJersey.testRegisterService(TestObjectPayloadWithJersey.java:134)

[INFO] Running org.apache.curator.x.discovery.server.jetty_jersey.TestStringsWithJersey
[ERROR] Tests run: 2, Failures: 2, Errors: 0, Skipped: 0, Time elapsed: 1.532 s <<< FAILURE! - in org.apache.curator.x.discovery.server.jetty_jersey.TestStringsWithJersey
[ERROR] testEmptyServiceNames(org.apache.curator.x.discovery.server.jetty_jersey.TestStringsWithJersey)  Time elapsed: 0.14 s  <<< FAILURE!
com.sun.jersey.api.client.UniformInterfaceException: GET http://localhost:36037/v1/service returned a response status of 404 Not Found
	at org.apache.curator.x.discovery.server.jetty_jersey.TestStringsWithJersey.testEmptyServiceNames(TestStringsWithJersey.java:164)

[ERROR] testRegisterService(org.apache.curator.x.discovery.server.jetty_jersey.TestStringsWithJersey)  Time elapsed: 0.238 s  <<< FAILURE!
com.sun.jersey.api.client.UniformInterfaceException: PUT http://localhost:45169/v1/service/test/7fc15020-0c05-47d3-91a7-a6788442caf9 returned a response status of 404 Not Found
	at org.apache.curator.x.discovery.server.jetty_jersey.TestStringsWithJersey.testRegisterService(TestStringsWithJersey.java:130)

curator-client/src/main/java/org/apache/curator/utils/Compatibility.java
```diff
@@ -18,29 +18,49 @@
...
+            localHasZooKeeperAdmin = false;
+            logger.info("Running in ZooKeeper 3.4.x compatibility mode");
+        }
+        hasZooKeeperAdmin = localHasZooKeeperAdmin;
+
+        Method localQueueEventMethod;
+        try
+        {
+            Class<?> testableClass = Class.forName("org.apache.zookeeper.Testable");
+            localQueueEventMethod = testableClass.getMethod("queueEvent", WatchedEvent.class);
```
shayshim: Did you actually want to try calling getMethod("queueEvent", ) on the ZooKeeper class? There is no queueEvent method in Testable, but there is in ZooKeeper of 3.5.

shayshim: I didn't understand - Testable has no method "queueEvent"..

Randgalt: It will - https://github.com/apache/zookeeper/pull/799

Randgalt:
@cammckenzie Jenkins tests are passing. Those tests also passed on my machine. However, I'll run a few more times to make sure.

Randgalt:
FYI (cc @cammckenzie @shayshim) tests just passed again on Jenkins: https://builds.apache.org/job/Curator/101/

Randgalt:
@cammckenzie what do you think? Can I merge this?

cammckenzie:
I think the changes are OK. I'm still having issues with the tests, but maybe that's something local. So, merge away.

TisonKun:
@Randgalt does Curator 2.x suffer this issue and if so, will we make a backport bugfix?

Randgalt:
> does Curator 2.x suffer this issue and if so, will we make a backport bugfix?
No, I don't think it does. Curator 2.0 does not do "fake" session expiration. That was added in 3.x.

PR #304, opened 2019-02-07
[CURATOR-505] WIP - Circuit breaking connection state listener decorator

Randgalt:
A decorator/proxy for connection state listeners that adds circuit breaking behavior. During network outages ZooKeeper can become very noisy sending connection/disconnection events in rapid succession. Curator recipes respond to these messages by resetting state, etc. E.g. LeaderLatch must delete its lock node and try to recreate it in order to try to re-obtain leadership, etc.

This noisy herding can be avoided by using the circuit breaking listener decorator. When it receives ConnectionState.SUSPENDED, the circuit becomes "open" (based on the provided RetryPolicy) and will ignore future connection state changes until RetryPolicy timeout has elapsed. Note: however, if the connection goes from ConnectionState.SUSPENDED to ConnectionState.LOST the first LOST state is sent.

For reviewers:
- [CircuitBreakingConnectionStateListener.java ](
https://github.com/apache/curator/pull/304/files#diff-7de291537103ffec9bfbbc51fd74923d) - the main worker
- [CircuitBreaker.java](
https://github.com/apache/curator/pull/304/files#diff-9552eec2f1af385e82e77b0c1463e142) - the circuit breaking semantics
- [TestLeaderLatch.java](https://github.com/apache/curator/pull/304/files#diff-d289c49b2bf336605b989bb7da8b3356R64) - shows how it works in practice. The server is stopped/restarted in fast succession 5 times. Without the circuit breaker it would result in 6 resets (1 for the initial reset). With the circuit breaker, it's only 2 no matter how many stops/restarts are executed.

### DIFF
--- a/curator-framework/src/main/java/org/apache/curator/framework/state/CircuitBreaker.java
+++ b/curator-framework/src/main/java/org/apache/curator/framework/state/CircuitBreaker.java
@@ -20,6 +20,7 @@ package org.apache.curator.framework.state;
 
 import org.apache.curator.RetryPolicy;
 import org.apache.curator.RetrySleeper;
+import java.time.Duration;
 import java.util.Objects;
 import java.util.concurrent.ScheduledExecutorService;
 import java.util.concurrent.TimeUnit;
@@ -77,7 +78,8 @@ class CircuitBreaker
 
         long[] sleepTimeNanos = new long[]{0L};
         RetrySleeper retrySleeper = (time, unit) -> sleepTimeNanos[0] = unit.toNanos(time);
-        if ( retryPolicy.allowRetry(retryCount, System.nanoTime() - startNanos, retrySleeper) )
+        Duration elapsedTime = Duration.ofNanos(System.nanoTime() - startNanos);
+        if ( retryPolicy.allowRetry(retryCount, elapsedTime.toMillis(), retrySleeper) )
         {
             ++retryCount;
             service.schedule(completion, sleepTimeNanos[0], TimeUnit.NANOSECONDS);
--- a/curator-framework/src/main/java/org/apache/curator/framework/state/CircuitBreaker.java
+++ b/curator-framework/src/main/java/org/apache/curator/framework/state/CircuitBreaker.java
@@ -14,7 +14,7 @@ class CircuitBreaker
 
     private boolean isOpen = false;
     private int retryCount = 0;
-    private long openStartNanos = 0;
+    private long startNanos = 0;
 
     CircuitBreaker(RetryPolicy retryPolicy, ScheduledExecutorService service)
     {
@@ -41,13 +41,13 @@ class CircuitBreaker
 
         isOpen = true;
         retryCount = 0;
-        openStartNanos = System.nanoTime();
-        if ( !tryToRetry(completion) )
+        startNanos = System.nanoTime();
+        if ( tryToRetry(completion) )
         {
-            close();
-            return false;
+            return true;
         }
-        return true;
+        close();
+        return false;
     }
 
     boolean tryToRetry(Runnable completion)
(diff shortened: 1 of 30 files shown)


curator-framework/src/main/java/org/apache/curator/framework/state/CircuitBreaker.java
```diff
@@ -0,0 +1,79 @@
...
+    boolean tryToRetry(Runnable completion)
+    {
+        if ( !isOpen )
+        {
+            return false;
+        }
+
+        long[] sleepTimeNanos = new long[]{0L};
+        RetrySleeper retrySleeper = (time, unit) -> sleepTimeNanos[0] = unit.toNanos(time);
+        if ( !retryPolicy.allowRetry(retryCount, System.nanoTime() - openStartNanos, retrySleeper) )
```
Randgalt: Note: this is a little clumsy given how Curator's RetryPolicy works. But, this is getting the amount of sleep the retry desires (internally the retry policy calls the sleeper with a sleep amount)

curator-framework/src/main/java/org/apache/curator/framework/state/CircuitBreakingConnectionStateListener.java
```diff
@@ -0,0 +1,177 @@
...
+                log.info("Circuit is opening. State: {} post-retryCount: {}", newState, circuitBreaker.getRetryCount());
+                circuitLastState = circuitInitialState = newState;
+                circuitLostHasBeenSent = (newState == ConnectionState.LOST);
+            }
+            else
+            {
+                log.debug("Could not open circuit breaker. State: {}", newState);
+            }
+        }
+        callListener(circuitInitialState);
```
Randgalt: should be `newState`

curator-recipes/src/main/java/org/apache/curator/framework/recipes/leader/LeaderLatch.java
```diff
@@ -149,6 +141,7 @@ public LeaderLatch(CuratorFramework client, String latchPath, String id, CloseMo
         this.latchPath = PathUtils.validatePath(latchPath);
         this.id = Preconditions.checkNotNull(id, "id cannot be null");
         this.closeMode = Preconditions.checkNotNull(closeMode, "closeMode cannot be null");
+        listener = client.decorateConnectionStateListener((__, newState) -> handleStateChange(newState));
```
Randgalt: Note: I'm going to make the decoration automatic. It's silly to force this on clients.

curator-framework/src/main/java/org/apache/curator/framework/state/CircuitBreaker.java
```diff
@@ -0,0 +1,79 @@
...
+    boolean tryToRetry(Runnable completion)
+    {
+        if ( !isOpen )
+        {
+            return false;
+        }
+
+        long[] sleepTimeNanos = new long[]{0L};
+        RetrySleeper retrySleeper = (time, unit) -> sleepTimeNanos[0] = unit.toNanos(time);
+        if ( retryPolicy.allowRetry(retryCount, System.nanoTime() - startNanos, retrySleeper) )
```
Randgalt: Note: this is a little clumsy given how Curator's RetryPolicy works. But, this is getting the amount of sleep the retry desires (internally the retry policy calls the sleeper with a sleep amount)

yuri-tceretian: According to the documentation of `RetryPolicy.allowRetry`, the second argument is supposed to be in milliseconds. You provide nanos instead.

Randgalt: Java's `TimeUnit` can convert to any unit. So, for future benefit I'm converting to nanos via `unit.toNanos`.

yuri-tceretian: If I read the code correctly, it can cause problems with policies like `RetryUntilElapsed`. Please correct me if I am wrong

Randgalt: Doh - I misread your comment. I see the problem now and it's fixex.

yuri-tceretian:
> I could make the CircuitBreaker itself configurable.

I think another circuit breaker would be better. Anyway, that's just an idea. The current implementation of circuit breaker is a good one.

PR #305, opened 2019-03-21
CURATOR-511: Add toString to ZKPaths PathAndNode

BELUGABEHR:


curator-client/src/main/java/org/apache/curator/utils/ZKPaths.java
```diff
@@ -154,6 +154,12 @@ public String getNode()
         {
             return node;
         }
+
+        @Override
+        public String toString()
+        {
+            return "PathAndNode [path=" + path + ", node=" + node + "]";
```
Randgalt: Why not return `ZKPaths.makePath(path, node)`?

Randgalt:
While we're at it maybe we should add `hashCode()` and `equals`?

curator-client/src/main/java/org/apache/curator/utils/ZKPaths.java
BELUGABEHR: Because I want to know what the node was and the path was. I already know what I passed in :)

BELUGABEHR:
@Randgalt  Thank you for your time, feedback, and interest.  I have pushed a new updated with hashcode and equals.

PR #308, opened 2019-03-25
CURATOR-515: Remove use of deprecated Throwables method

belugabehr:


tisonkun (changes requested):
This may cause class cast issue if `e` is neither `RuntimeException`, `Error` nor `Exception`.

I'd prefer you simply replace `Throwables.propagate(e);` with:

```java
Throwables.throwIfUnchecked(e);
throw new RuntimeException(e);
```

as is. And handle all occurrences among the Curator project.

PR #309, opened 2019-03-25
CURATOR-516: Remove Deprecated newSetFromMap

belugabehr:


BELUGABEHR:
Hello @Randgalt.

You were able to help me with a previous Curator PR.  Can I get your help once more?  Thanks.

Randgalt:
What's the benefit of this?

tisonkun:
@Randgalt Guava's `Sets.newSetFromMap` is a wrapper of `Collections.newSetFromMap` and there is a deprecation comments: `Use {@link Collections#newSetFromMap} instead.`

However, this change is not quite necessary as we can always handle it when a version bump of Guava cause this method removed or we'd like to (it's less possible, although) remove dependency of Guava.

tisonkun:
Closed as Not A Problem.

PR #311, opened 2019-05-13
[CURATOR-429] Make Curator 4.x compatible with Zookeeper 3.4.x in OSGi too

JiriOndrusek:
Issue: https://issues.apache.org/jira/browse/CURATOR-429

Curator 4.0 supports ZooKeeper 3.4.x (https://curator.apache.org/zk-compatibility.html)
This change allows this dependency also in OSGi. (where it required 3.5)

### DIFF
--- a/curator-client/pom.xml
+++ b/curator-client/pom.xml
@@ -36,7 +36,7 @@
 
     <properties>
         <osgi.import.package>
-            *
+            org.apache.zookeeper.*;version="[3.4,4.0)", *
         </osgi.import.package>
         <osgi.export.package>
             org.apache.curator*;version="${project.version}";-noimport:=true
--- a/curator-framework/pom.xml
+++ b/curator-framework/pom.xml
@@ -37,7 +37,7 @@
 
     <properties>
         <osgi.import.package>
-            *
+            org.apache.zookeeper.*;version="[3.4,4.0)", *
         </osgi.import.package>
         <osgi.export.package>
             org.apache.curator.framework*;version="${project.version}";-noimport:=true
--- a/curator-recipes/pom.xml
+++ b/curator-recipes/pom.xml
@@ -37,7 +37,7 @@
 
     <properties>
         <osgi.import.package>
-            *
+            org.apache.zookeeper.*;version="[3.4,4.0)", *
         </osgi.import.package>
         <osgi.export.package>
             org.apache.curator.framework.recipes*;version="${project.version}";-noimport:=true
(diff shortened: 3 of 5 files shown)


oscerd:
Thanks @JiriOndrusek

curator-client/pom.xml
```diff
@@ -36,7 +36,7 @@
 
     <properties>
         <osgi.import.package>
-            *
+            org.apache.zookeeper.*;version="[3.4,4.0)", *
```
shayshim: I have no background with OSGi. Could you explain this change? I mean, why "osgi.import.package" with '*' didn't do what you need and this change fixes it? Usually '*' is most general.

oscerd:
Without applying a version range the result is that curator will work only with zookeeper 3.5, because this the dependency declared in curator. It will use what is located in the POM. Relaxing the requirement by introducing a range is usually the best way to manage this situation. This will made true the sentence "curator can work with zookeeper 3.4 and 3.5, from version 4.x". Actually this is not true for OSGi container.

shayshim:
Thank you @oscerd, could you explain what do you mean by "Without applying a version range the result is that curator will work only with zookeeper 3.5"? It is possible to use Curator 4.x with zookeeper 3.4.x, as described here: https://curator.apache.org/zk-compatibility.html (in fact, we are using it like that). Could you elaborate a bit what this maven change does and why needed for OSGi?

grgrzybek:
@shayshim OSGi is generally very confusing runtime/technology to enforce the most sane and obvious rules that should always be used.
In other words, if you assume _semantic versioning_ (which tells you that between 1.0.0 and 2.0.0 you have breaking changes and between 1.0.0 and 1.1.0 you have some changes and between 1.0.0 and 1.0.1 you have only bug fixes), then you should not be able to blindly upgrade some dependencies if they have some breaking or medium API changes.

OSGi (here, using maven-bundle-plugin and bndlib) gives you strict rules - it can analyze your POM and decide about the safe version _range_ you can use.
By default, if Curator 4.1.1 has dependency on Zookeeper 3.5.4-beta, the generated/assumed/default _Import range_ is `[3.5,4.0)`, which again - is reasonable. maven-bundle-plugin doesn't know that Curator is also compatible with Zookeeper 3.4.x - in this case you have to be explicit and use `[3.4,4.0)` range instead of `*` which generates `[3.5,4.0)`.

I hope this clears confusion. Also @oscerd provided good explanation.

oscerd:
Quote @grgrzybek

shayshim:
@grgrzybek  @oscerd thanks!

PR #315, opened 2019-07-03
[CURATOR-530] Documentation on InterProcessSemaphoreMutex is misleading

emretetik96:


curator-recipes/src/main/java/org/apache/curator/framework/recipes/locks/InterProcessMultiLock.java
```diff
@@ -145,6 +145,9 @@ public boolean acquire(long time, TimeUnit unit) throws Exception
      * {@inheritDoc}
      *
      * <p>NOTE: locks are released in the reverse order that they were acquired.</p>
+     *	
+     * @throws Exception ZK errors, interruptions, current thread does not own the lock
```
shayshim: Seems right, but don't you need to add now same "@throws" text also for InterProcessMutex.release()? (This is also the actual code that throws IllegalMonitorStateException in case of different thread, and used internally here)

curator-recipes/src/main/java/org/apache/curator/framework/recipes/locks/InterProcessLock.java
```diff
@@ -44,7 +44,7 @@
     /**
      * Perform one release of the mutex.
      *
-     * @throws Exception ZK errors, interruptions, current thread does not own the lock
+     * @throws Exception ZK errors, interruptions
```
shayshim: On the one hand it seems true that not all implementations throw exception in case of different thread, on the other hand, after removing this comment from the interface - its users cannot be aware of the possibility that this exception can be thrown (depending on the implementation). So user that did not create this InterProcessLock, but calls release() from different thread, might be surprised of the IllegalMonitorStateException (in case it is InterProcessMutex, for instance).
Maybe add above the interface InterProcessLock top level javadoc (empty currently):
NOTE: depending on its implementation, InterProcessLock#release() might throw exception if current thread does not own the lock

emretetik96: That makes sense, updating the PR

curator-recipes/src/main/java/org/apache/curator/framework/recipes/locks/InterProcessMultiLock.java
emretetik96: @shayshim `InterProcessMutex.release()` already has that `@throws` in its javadoc. See https://github.com/emretetik96/curator/blob/77470cbbe0befa26500c6d18e6fc93c29f831725/curator-recipes/src/main/java/org/apache/curator/framework/recipes/locks/InterProcessMutex.java#L126

curator-recipes/src/main/java/org/apache/curator/framework/recipes/locks/InterProcessLock.java
emretetik96: updated

curator-recipes/src/main/java/org/apache/curator/framework/recipes/locks/InterProcessMultiLock.java
shayshim: @emretetik96 right, my mistake

PR #320, opened 2019-07-28
[CURATOR-533] - improve circuit breaking behavior

Randgalt:
CURATOR-505 introduced circuit breaking behavior via `CircuitBreakingConnectionStateListener` and `ConnectionStateListenerDecorator`. Elastic has been using it to success but reports that the implementation can be improved. The existing implementation uses a new `CircuitBreaker` for each `ConnectionStateListener` set in a Curator client. It turns out that this is not ideal. Instead, a shared `CircuitBreaker` should be used per Curator client.

Unfortunately, the best way to do this is to remove the `ConnectionStateListenerDecorator` semantics and use a different mechanism. This Issue proposes to do this and remove `ConnectionStateListenerDecorator`. This is a breaking change but given the short amount of time it's been in Curator it's unlikely that it's been widely adopted.

In this commit, `ConnectionStateListenerDecorator` is removed in favor of `ConnectionStateListenerManagerFactory`. `ConnectionStateManager` uses this factory to create the container to hold registered `ConnectionStateListeners`. A new `CircuitBreakerManager` now manages the circuit breaking behavior using a shared `CircuitBreaker`.


curator-framework/src/main/java/org/apache/curator/framework/state/ConnectionStateListener.java
```diff
@@ -31,14 +31,14 @@
...
-     * for doNotDecorate(), it will not be passed through the decorator.
+     * ConnectionStateListener managers set via {@link org.apache.curator.framework.CuratorFrameworkFactory.Builder#connectionStateListenerManagerFactory(ConnectionStateListenerManagerFactory)}
+     * are allowed to proxy (etc.) ConnectionStateListeners as needed. If this method returns <code>true</code>
+     * the ConnectionStateListener manager must <em>not</em> proxy the listener as it's a vital internal
+     * listener used by Curator.
      *
      * @return true/false
      */
-    default boolean doNotDecorate()
+    default boolean doNotProxy()
```
shayshim: I think proxy is a bit vague. Why not something like useCircuitBreaker() instead of doNotProxy(), to be more clear about the context it is used in? Unless you see future possibilities to call it also for other 'proxies' than the circuit breaker.

curator-framework/src/main/java/org/apache/curator/framework/state/CircuitBreakingManager.java
```diff
@@ -0,0 +1,91 @@
...
+        else
+        {
+            mainContainer.addListener(listener, executor);
+        }
+    }
+
+    @Override
+    public void removeListener(ConnectionStateListener listener)
+    {
+        mainContainer.addListener(listener);
```
shayshim: Probably should remove?

curator-recipes/src/test/java/org/apache/curator/framework/recipes/leader/TestLeaderLatch.java
```diff
@@ -43,60 +43,119 @@
...
-        {
+            .build();
+        try {
             client.start();
-            AtomicInteger resetCount = new AtomicInteger(0);
-            try ( LeaderLatch latch = new LeaderLatch(client, "/foo/bar")
+            client.create().forPath("/hey");
+
+            Semaphore lostSemaphore = new Semaphore(0);
+            ConnectionStateListener undecoratedListener = new ConnectionStateListener()
```
shayshim: unproxiedListener/nonCircuitBreakingListener?

curator-recipes/src/test/java/org/apache/curator/framework/recipes/leader/TestLeaderLatch.java
```diff
@@ -43,60 +43,119 @@
...
 public class TestLeaderLatch extends BaseClassForTests
 {
     private static final String PATH_NAME = "/one/two/me";
     private static final int MAX_LOOPS = 5;
 
+    private static class Holder
+    {
+        final AtomicInteger stateChangeCount = new AtomicInteger(0);
+        final CountDownLatch isLockedLatch = new CountDownLatch(1);
+        volatile LeaderLatch latch;
```
shayshim: Could be simply final with constructor

curator-recipes/src/test/java/org/apache/curator/framework/recipes/leader/TestLeaderLatch.java
```diff
@@ -43,60 +43,119 @@
...
+                executorService.submit(() -> {
+                    holder.latch.start();
+                    Assert.assertTrue(holder.latch.await(timing.forWaiting().milliseconds(), TimeUnit.MILLISECONDS));
+                    holder.isLockedLatch.countDown();
+                    return null;
+                });
+                timing.awaitLatch(holder.isLockedLatch);
+            });
+
+            for ( int i = 0; i < 5; ++i )
```
shayshim: 5 is threadQty, right?

curator-recipes/src/test/java/org/apache/curator/framework/recipes/leader/TestLeaderLatch.java
```diff
@@ -43,60 +43,119 @@
...
+                server.stop();
+                Assert.assertTrue(timing.acquireSemaphore(lostSemaphore));
+                server.restart();
+                timing.sleepABit();
             }
+
+            for ( Holder holder : holders )
+            {
+                Assert.assertTrue(holder.latch.await(timing.forWaiting().milliseconds(), TimeUnit.MILLISECONDS));
+                Assert.assertEquals(holder.stateChangeCount.get(), 3);  // 1 suspend, 1 lost, 1 reconnected
```
shayshim: Could be bolder and collect the actual states into a list and assert them to be exactly these ones :)

curator-framework/src/main/java/org/apache/curator/framework/state/CircuitBreakingManager.java
```diff
@@ -0,0 +1,91 @@
...
+    public int size()
+    {
+        return mainContainer.size() + doNotProxyContainer.size();
+    }
+
+    @Override
+    public void forEach(Consumer<ConnectionStateListener> function)
+    {
+        doNotProxyContainer.forEach(function);
+        function.accept(masterListener);
```
shayshim: function.accept(masterListener) is probably the same as mainContainer.forEach(function), right? but I think the second is clearer and symmetric with doNotProxyContainer.forEach(function)

curator-framework/src/main/java/org/apache/curator/framework/state/CircuitBreakingManager.java
Randgalt: Yikes! Good catch

curator-framework/src/main/java/org/apache/curator/framework/state/ConnectionStateListener.java
Randgalt: It’s not specific to circuit breaking.

curator-framework/src/main/java/org/apache/curator/framework/state/CircuitBreakingManager.java
Randgalt: No, it's not the same. `masterListener` proxies the listeners in the `mainContainer`

curator-recipes/src/test/java/org/apache/curator/framework/recipes/leader/TestLeaderLatch.java
Randgalt: Actually not as the override on `handleStateChange` needs the `holder` instance.

curator-recipes/src/test/java/org/apache/curator/framework/recipes/leader/TestLeaderLatch.java
Randgalt: actually not - just a coincidence. I'll make it `6` so this is clear

curator-recipes/src/test/java/org/apache/curator/framework/recipes/leader/TestLeaderLatch.java
shayshim: Oh, right

curator-framework/src/main/java/org/apache/curator/framework/state/ConnectionStateListener.java
shayshim: Gotcha

curator-framework/src/main/java/org/apache/curator/framework/state/CircuitBreakingManager.java
shayshim: True. Shame on me :)

curator-recipes/src/test/java/org/apache/curator/framework/recipes/leader/TestLeaderLatch.java
Randgalt: done

Randgalt:
@shayshim comments addressed and rebased

PR #322, opened 2019-08-08
[CURATOR-536] break infinite loop in internalLeave() on timeout

abcdenis:
Hi,
please merge it :)


shayshim:
@abcdenis thanks for this fix

PR #324, opened 2019-09-02
[CURATOR-537] make getOurPath() public

Randgalt:
make getOurPath() public for reference purposes


cammckenzie:
Looks fine to me.

PR #326, opened 2019-09-21
[CURATOR-541] Fix retry code in BaseClassForTests

Randgalt:
The retry code in BaseClassForTests was hopelessly broken - I don't know for how long. Reworked it so that it does the right thing now (hopefully). Storing the retry count as an attribute wasn't working. The new method stores it as a field in the test class and makes sure that it's always correct. It should only ever be true when actually retrying and, given that TestNG always calls the retry method, it can be reset after a retry fails.


cammckenzie:
Looks good to me.

PR #328, opened 2019-10-11
CURATOR-547 change JAX-RS reader/writer to reuse Jackson ObjectMapper

cowtowncoder:
also upgrade Jackson version from 2.9.8 to 2.9.10 (latest 2.9.x patch).


Randgalt:
merged

PR #329, opened 2019-10-11
[CURATOR-546] - ModeledCacheImpl.currentData() was removing from entries instead of calling get()

Randgalt:
Bad copy/paste bug. ModeledCacheImpl.currentData() was removing from entries instead of calling get()


cammckenzie:
Looks good to me.

PR #331, opened 2019-10-28
Fix comment

yaojingguo:
1. Use internal consistently since PathChildrenCache uses internal.
2. Fix a typo.


curator-x-discovery/src/main/java/org/apache/curator/x/discovery/ServiceCacheBuilder.java
```diff
@@ -49,7 +49,7 @@
     public ServiceCacheBuilder<T> threadFactory(ThreadFactory threadFactory);
 
     /**
-     * Optional ExecutorService to use for the cache's background thread. The specified ExecutorService
+     * Optional ExecutorService to use for the cache's internal thread. The specified ExecutorService
```
tisonkun, line 52: I don't think this is a necessary change. You may update this PR to keep the typo fix change and that looks good to me.

yaojingguo:
A new commit is ready for review.

tisonkun (approved):

PR #334, opened 2019-11-03
[CURATOR-549] Support ZooKeeper 3.6.0 Persistent Recursive Watchers

Randgalt:

Support persistent watchers in ZK 3.6+ while maintaining background compatibility with previous versions of ZK. Adds Curator Framework DSL calls to ZooKeeper's `addWatch()` method. Subsequent PRs will add recipes. Both the older Framework and the async Framework now have methods to add persistent watchers.

### DIFF
--- a/curator-framework/src/main/java/org/apache/curator/framework/api/AddWatchBuilder.java
+++ b/curator-framework/src/main/java/org/apache/curator/framework/api/AddWatchBuilder.java
@@ -16,20 +16,17 @@
  * specific language governing permissions and limitations
  * under the License.
  */
-package org.apache.curator.zk35;
+package org.apache.curator.framework.api;
 
-import org.apache.curator.utils.Compatibility;
-import org.testng.Assert;
-import org.testng.annotations.Test;
+import org.apache.zookeeper.AddWatchMode;
 
-public class TestIs35
+public interface AddWatchBuilder extends AddWatchBuilder2
 {
-    @Test
-    public void testIsZk35()
-    {
-        Assert.assertFalse(Compatibility.hasGetReachableOrOneMethod());
-        Assert.assertTrue(Compatibility.hasAddrField());
-        Assert.assertFalse(Compatibility.hasPersistentWatchers());
-    }
+    /**
+     * The mode to use. By default, {@link org.apache.zookeeper.AddWatchMode#PERSISTENT_RECURSIVE} is used
+     *
+     * @param mode mode to use
+     * @return this
+     */
+    AddWatchBuilder2 withMode(AddWatchMode mode);
 }
-
--- a/curator-framework/src/main/java/org/apache/curator/framework/imps/CuratorFrameworkImpl.java
+++ b/curator-framework/src/main/java/org/apache/curator/framework/imps/CuratorFrameworkImpl.java
@@ -41,6 +41,7 @@ import org.apache.curator.framework.state.ConnectionState;
 import org.apache.curator.framework.state.ConnectionStateErrorPolicy;
 import org.apache.curator.framework.state.ConnectionStateListener;
 import org.apache.curator.framework.state.ConnectionStateManager;
+import org.apache.curator.utils.Compatibility;
 import org.apache.curator.utils.DebugUtils;
 import org.apache.curator.utils.EnsurePath;
 import org.apache.curator.utils.ThreadUtils;
@@ -556,6 +557,13 @@ public class CuratorFrameworkImpl implements CuratorFramework
     public RemoveWatchesBuilder watches()
     {
         return new RemoveWatchesBuilderImpl(this);
+    }
+
+    @Override
+    public WatchesBuilder watchers()
+    {
+        Preconditions.checkState(Compatibility.hasPersistentWatchers(), "watchers() is not supported in the ZooKeeper library being used. Use watches() instead.");
+        return new WatchesBuilderImpl(this);
     }
 
     protected void internalSync(CuratorFrameworkImpl impl, String path, Object context)
--- a/curator-framework/src/test/java/org/apache/curator/framework/imps/TestWatchesBuilder.java
+++ b/curator-framework/src/test/java/org/apache/curator/framework/imps/TestWatchesBuilder.java
@@ -33,11 +33,14 @@ import org.apache.curator.retry.RetryOneTime;
 import org.apache.curator.test.Timing;
 import org.apache.curator.test.compatibility.CuratorTestBase;
 import org.apache.curator.utils.CloseableUtils;
+import org.apache.curator.utils.ZookeeperFactory;
+import org.apache.zookeeper.AddWatchMode;
 import org.apache.zookeeper.KeeperException;
 import org.apache.zookeeper.WatchedEvent;
 import org.apache.zookeeper.Watcher;
 import org.apache.zookeeper.Watcher.Event.EventType;
 import org.apache.zookeeper.Watcher.WatcherType;
+import org.apache.zookeeper.ZooKeeper;
 import org.testng.Assert;
 import org.testng.annotations.Test;
 import java.util.concurrent.CountDownLatch;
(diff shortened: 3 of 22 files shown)


curator-framework/src/main/java/org/apache/curator/framework/api/AddWatchBuilder.java
```diff
@@ -0,0 +1,32 @@
...
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package org.apache.curator.framework.api;
+
+import org.apache.zookeeper.AddWatchMode;
+
+public interface AddWatchBuilder extends AddWatchBuilder2
```
shayshim: Why do we need AddWatchBuilder2 and AddWatchBuilder separated? Why not merge them into one interface? Did you want to clean AddWatchBuilder from the parent interfaces?

curator-x-async/src/main/java/org/apache/curator/x/async/api/AsyncCuratorFrameworkDsl.java
```diff
@@ -112,4 +112,11 @@
      * @return builder object
      */
     AsyncRemoveWatchesBuilder removeWatches();
+
+    /**
+     * Start an add watch builder
+     *
```
shayshim: Consider adding something like: supported only when ZooKeeper client of version 3.6 or above is used, throws IllegalStateException for ZooKeeper client 3.5 or below

curator-framework/src/main/java/org/apache/curator/framework/imps/CuratorFrameworkImpl.java
```diff
@@ -572,6 +572,13 @@ public RemoveWatchesBuilder watches()
         return new RemoveWatchesBuilderImpl(this);
     }
 
+    @Override
+    public WatchesBuilder watchers()
+    {
+        Preconditions.checkState(Compatibility.hasPersistentWatchers(), "addWatch() is not supported in the ZooKeeper library being used. Use watches() instead.");
```
shayshim: The exception message: "addWatch()" -> "watchers()"

curator-framework/src/main/java/org/apache/curator/framework/api/AddWatchBuilder.java
Randgalt: `AddWatchBuilder` has the "withMode" method which is optional.

curator-x-async/src/main/java/org/apache/curator/x/async/api/AsyncCuratorFrameworkDsl.java
```diff
@@ -112,4 +112,11 @@
...
+     * @return builder object
+     */
+    AsyncWatchBuilder addWatch();
```
shayshim: Shouldn't addWatch() be named watchers(), line in CuratorFramwork, to be consistent with the naming?

curator-framework/src/test/java/org/apache/curator/framework/imps/TestWatchesBuilder.java
```diff
@@ -611,6 +614,56 @@ public void pathAddedForGuaranteedOperation(
...
+        try ( CuratorFramework client = CuratorFrameworkFactory.newClient(server.getConnectString(), new RetryOneTime(1)) )
+        {
+            client.start();
+            client.blockUntilConnected();
+
+            CountDownLatch latch = new CountDownLatch(5);
+            Watcher watcher = event -> latch.countDown();
+            client.watchers().add().withMode(AddWatchMode.PERSISTENT_RECURSIVE).usingWatcher(watcher).forPath("/test");
+
+            client.create().forPath("/test");
```
shayshim: Not interesting to test here also other recursive events like data changes and remove nodes?
Also, might be good to create and modify recursively "/not-the-test" and make sure nothing comes from there.
And remove watcher tests shouldn't be repeated for persistent/recursive too?

curator-framework/src/main/java/org/apache/curator/framework/api/AddWatchBuilder.java
shayshim: OK, it is just a little hard for me with using the same name but with 2 as suffix, but I don't have a better name to offer :)

curator-x-async/src/main/java/org/apache/curator/x/async/api/AsyncCuratorFrameworkDsl.java
Randgalt: No, the DSL verbs are different in for async and sync

curator-framework/src/test/java/org/apache/curator/framework/imps/TestWatchesBuilder.java
Randgalt: I was thinking it isn't necessary to test functionality that's already tested in ZK itself. But, this is a good point. I'll add some more tests.

Randgalt:
> In the case that ZK server is using the Optimized WatchManager it would not be able to support PersistentRecursive watches.
> How can we deal with this problem ?

Hmm - good call. I'll consider this.

Randgalt:
attn: @eolivelli @cammckenzie @shayshim @TisonKun

curator-x-async/src/main/java/org/apache/curator/x/async/details/AsyncCuratorFrameworkImpl.java
```diff
@@ -151,6 +153,13 @@ public AsyncRemoveWatchesBuilder removeWatches()
         return new AsyncRemoveWatchesBuilderImpl(client, filters);
     }
 
+    @Override
+    public AsyncWatchBuilder addWatch()
+    {
+        Preconditions.checkState(Compatibility.hasPersistentWatchers(), "addWatch() is not supported in the ZooKeeper library being used.");
```
eolivelli: I would throw UnsupportedOperationException

Randgalt: it's always been this way - I'd rather leave it for historical reasons

curator-framework/src/test/java/org/apache/curator/framework/imps/TestWatchesBuilder.java
```diff
@@ -610,6 +613,122 @@ public void pathAddedForGuaranteedOperation(
...
+
+    @Test(groups = CuratorTestBase.zk36Group)
+    public void testPersistentWatchInBackground() throws Exception
+    {
+        try ( CuratorFramework client = CuratorFrameworkFactory.newClient(server.getConnectString(), new RetryOneTime(1)) )
+        {
+            client.start();
+            client.blockUntilConnected();
+
+            CountDownLatch backgroundLatch = new CountDownLatch(1);
```
cammckenzie: What's the purpose of the backgroundLatch? It doesn't seem to actually get used (beyond being counted down).

curator-framework/src/test/java/org/apache/curator/framework/imps/TestWatchesBuilder.java
```diff
@@ -610,6 +613,122 @@ public void pathAddedForGuaranteedOperation(
...
+
+    @Test(groups = CuratorTestBase.zk36Group)
+    public void testPersistentRecursiveWatchInBackground() throws Exception
+    {
+        try ( CuratorFramework client = CuratorFrameworkFactory.newClient(server.getConnectString(), new RetryOneTime(1)) )
+        {
+            client.start();
+            client.blockUntilConnected();
+
+            CountDownLatch backgroundLatch = new CountDownLatch(1);
```
cammckenzie: As for previous backgroundLatch comment

Randgalt: Good catch on both

curator-recipes/src/test/java/org/apache/curator/framework/recipes/leader/TestLeaderSelectorEdges.java
```diff
@@ -26,6 +26,7 @@
 import org.apache.curator.framework.state.ConnectionState;
 import org.apache.curator.retry.RetryNTimes;
 import org.apache.curator.test.BaseClassForTests;
+import org.apache.curator.test.compatibility.CuratorTestBase;
```
TisonKun: Unused import

PR #335, opened 2019-11-03
[CURATOR-549] Recipes based on Persistent Recursive Watchers

Randgalt:

This PR adds several recipes that use the new ZOOKEEPER-1416 Persistent Recursive watches from ZooKeeper 3.6.0.

## PersistentWatcher

A wrapper recipe that keeps a persistent (single or recursive) watch set and active through disconnections, etc.

See: [PersistentWatcher.java](https://github.com/apache/curator/blob/CURATOR-549-zk36-persistent-watcher-recipes/curator-recipes/src/main/java/org/apache/curator/framework/recipes/watch/PersistentWatcher.java)

## CuratorCache

Completely re-written cache recipe that will replace `TreeCache`, `NodeCache` and `PathChildrenCache`. With the benefit of persistent recursive watchers and the new PersistentWatcher recipe, the implementation is far simpler, will use significantly less resources and network calls, be easier to support and should be more stable and performant.

Wrappers for the older cache's listeners are provided to help with transitions.

See:

- [CuratorCache](https://github.com/apache/curator/blob/CURATOR-549-zk36-persistent-watcher-recipes/curator-recipes/src/main/java/org/apache/curator/framework/recipes/cache/CuratorCache.java) - this is the API
- [CuratorCacheImpl](https://github.com/apache/curator/blob/CURATOR-549-zk36-persistent-watcher-recipes/curator-recipes/src/main/java/org/apache/curator/framework/recipes/cache/CuratorCacheImpl.java) - this is the implementation

### DIFF
--- a/curator-recipes/src/main/java/org/apache/curator/framework/recipes/cache/CuratorCache.java
+++ b/curator-recipes/src/main/java/org/apache/curator/framework/recipes/cache/CuratorCache.java
@@ -1,0 +1,136 @@
+/**
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+package org.apache.curator.framework.recipes.cache;
+
+import org.apache.curator.framework.CuratorFramework;
+import org.apache.curator.framework.listen.Listenable;
+import java.io.Closeable;
+import java.util.Optional;
+import java.util.stream.Stream;
+
+/**
+ * <p>
+ *     A utility that attempts to keep the data from a node locally cached. Optionally the entire
+ *     tree of children below the node can also be cached. Will respond to update/create/delete events, pull
+ *     down the data, etc. You can register listeners that will get notified when changes occur.
+ * </p>
+ *
+ * <p>
+ *     <b>IMPORTANT</b> - Due to how ZooKeeper works you will not get notified of every single event.
+ *     For example during a network partition the cache will not get events. Imagine the following scenario:
+ *
+ *     <ul>
+ *         <li>Pre-network partition the cache contains "/foo" and "/foo/bar"</li>
+ *         <li>A network partition occurs and your Curator client loses connection to the server</li>
+ *         <li>Image another client that isn't partitioned, deletes "/foo/bar" and then a third client re-creates "/foo/bar"</li>
+ *         <li>Your client's partition is fixed. The cache will only see <em>one</em> change - the third client's re-create</li>
+ *     </ul>
+ *
+ *     Additionally, remember that ZooKeeper is an eventual consistent system. Always use ZNode version
+ *     numbers when updating nodes.
+ * </p>
+ */
+public interface CuratorCache extends Closeable, CuratorCacheAccessor
+{
+    /**
+     * cache build options
+     */
+    enum Options
+    {
+        /**
+         * Normally the entire tree of nodes starting at the given node are cached. This option
+         * causes only the given node to be cached (i.e. a single node cache)
+         */
+        SINGLE_NODE_CACHE,
+
+        /**
+         * Decompress data via {@link org.apache.curator.framework.api.GetDataBuilder#decompressed()}
+         */
+        COMPRESSED_DATA,
+
+        /**
+         * Normally, when the cache is closed via {@link CuratorCache#close()}, the storage is cleared
+         * via {@link CuratorCacheStorage#clear()}. This option prevents the storage from being cleared.
+         */
+        DO_NOT_CLEAR_ON_CLOSE
+    }
+
+    /**
+     * Return a Curator Cache for the given path with the given options using a standard storage instance
+     *
+     * @param client Curator client
+     * @param path path to cache
+     * @param options any options
+     * @return cache (note it must be started via {@link #start()}
+     */
+    static CuratorCache build(CuratorFramework client, String path, Options... options)
+    {
+        return builder(client, path).withOptions(options).build();
+    }
+
+    /**
+     * Start a Curator Cache builder
+     *
+     * @param client Curator client
+     * @param path path to cache
+     * @return builder
+     */
+    static CuratorCacheBuilder builder(CuratorFramework client, String path)
+    {
+        return new CuratorCacheBuilderImpl(client, path);
+    }
+
+    /**
+     * Start the cache. This will cause a complete refresh from the cache's root node and generate
+     * events for all nodes found, etc.
+     */
+    void start();
+
+    /**
+     * Close the cache, stop responding to events, etc.
+     */
+    @Override
+    void close();
+
+    /**
+     * Return the listener container so that listeners can be registered to be notified of changes to the cache
+     *
+     * @return listener container
+     */
+    Listenable<CuratorCacheListener> listenable();
+
+    /**
+     * {@inheritDoc}
+     */
+    @Override
+    Optional<ChildData> get(String path);
+
+    /**
+     * {@inheritDoc}
+     */
+    @Override
+    int size();
+
+    /**
+     * {@inheritDoc}
+     */
+    @Override
+    Stream<ChildData> stream();
+}
--- a/curator-recipes/src/main/java/org/apache/curator/framework/recipes/cache/CuratorCacheImpl.java
+++ b/curator-recipes/src/main/java/org/apache/curator/framework/recipes/cache/CuratorCacheImpl.java
@@ -1,0 +1,301 @@
+/**
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package org.apache.curator.framework.recipes.cache;
+
+import com.google.common.annotations.VisibleForTesting;
+import com.google.common.base.Preconditions;
+import com.google.common.collect.Sets;
+import org.apache.curator.framework.CuratorFramework;
+import org.apache.curator.framework.api.BackgroundCallback;
+import org.apache.curator.framework.api.CuratorEvent;
+import org.apache.curator.framework.listen.Listenable;
+import org.apache.curator.framework.listen.StandardListenerManager;
+import org.apache.curator.framework.recipes.watch.PersistentWatcher;
+import org.apache.curator.utils.ThreadUtils;
+import org.apache.curator.utils.ZKPaths;
+import org.apache.zookeeper.KeeperException;
+import org.apache.zookeeper.WatchedEvent;
+import org.apache.zookeeper.data.Stat;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import java.util.Collections;
+import java.util.Optional;
+import java.util.Set;
+import java.util.concurrent.atomic.AtomicReference;
+import java.util.function.Consumer;
+import java.util.stream.Stream;
+
+import static org.apache.curator.framework.recipes.cache.CuratorCacheListener.Type.*;
+import static org.apache.zookeeper.KeeperException.Code.NONODE;
+import static org.apache.zookeeper.KeeperException.Code.OK;
+
+class CuratorCacheImpl implements CuratorCache
+{
+    private final Logger log = LoggerFactory.getLogger(getClass());
+    private final AtomicReference<State> state = new AtomicReference<>(State.LATENT);
+    private final PersistentWatcher persistentWatcher;
+    private final CuratorFramework client;
+    private final CuratorCacheStorage storage;
+    private final String path;
+    private final boolean recursive;
+    private final boolean compressedData;
+    private final boolean clearOnClose;
+    private final StandardListenerManager<CuratorCacheListener> listenerManager = StandardListenerManager.standard();
+    private final Consumer<Exception> exceptionHandler;
+    private final OutstandingOps outstandingOps = new OutstandingOps(() -> callListeners(CuratorCacheListener::initialized));
+
+    private enum State
+    {
+        LATENT,
+        STARTED,
+        CLOSED
+    }
+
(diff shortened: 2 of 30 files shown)


Randgalt:
Attn: @cammckenzie @shayshim @TisonKun - reviews would be appreciated

TisonKun:
Thanks for your work! I'm reviewing now.

curator-recipes/src/main/java/org/apache/curator/framework/recipes/cache/CuratorCache.java
```diff
@@ -0,0 +1,133 @@
...
+     * {@inheritDoc}
+     */
+    @Override
+    Stream<ChildData> stream();
+
+    /**
+     * {@inheritDoc}
+     */
+    @Override
+    Stream<ChildData> streamImmediateChildren(String fromParent);
```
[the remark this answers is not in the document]
Randgalt: I'm not happy with this API. I'm going to change it. Please refer to the next commit.

Randgalt:
@TisonKun 2 comments have been addressed.

curator-recipes/src/main/java/org/apache/curator/framework/recipes/cache/OutstandingOps.java
```diff
@@ -0,0 +1,60 @@
...
+ * under the License.
+ */
+package org.apache.curator.framework.recipes.cache;
+
+import java.util.concurrent.atomic.AtomicLong;
+
+class OutstandingOps
+{
+    private final Runnable completionProc;
+    private volatile AtomicLong count = new AtomicLong(0);
```
eolivelli: What not final?

curator-recipes/src/main/java/org/apache/curator/framework/recipes/cache/OutstandingOps.java
```diff
@@ -0,0 +1,60 @@
...
+    }
+
+    void decrement()
+    {
+        AtomicLong localCount = count;
+        if ( localCount != null )
+        {
+            if ( (localCount.decrementAndGet() == 0) )
+            {
+                count = null;
```
eolivelli: I am not sure this trick works.

You can be here when another thread calls increment function.
You are not inside a lock

Randgalt: Look at the next line `if ( localCount.compareAndSet(0, Long.MIN_VALUE) )` - that ensures consistency (please review and confirm).

curator-recipes/src/main/java/org/apache/curator/framework/recipes/cache/OutstandingOps.java
Randgalt: It gets set to null after `completionProc` is called to signal decrement/increment are no longer necessary.

curator-recipes/src/main/java/org/apache/curator/framework/recipes/cache/OutstandingOps.java
Randgalt: Hmm - I think there's a race here - I'm going to address.

Randgalt: @eolivelli I reworked it - please re-check

curator-recipes/src/main/java/org/apache/curator/framework/recipes/cache/OutstandingOps.java
```diff
@@ -0,0 +1,55 @@
...
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package org.apache.curator.framework.recipes.cache;
+
+import java.util.concurrent.atomic.AtomicLong;
+import java.util.concurrent.atomic.AtomicReference;
+
+class OutstandingOps
```
TisonKun: An alternative is that `java.util.concurrent.Phaser` which already implements semantic like `OutstandingOps`.

```java
Runnable r = () -> callListeners(CuratorCacheListener::initialized);
Phaser p = new Phaser() {
  @Override
            protected boolean onAdvance(int phase, int registeredParties) {
                if (registeredParties > 0) {
                    return false;
                }
                r.run();
                return true;
            }
}
```

and replace `increment` as `register` while `decrement` as `arriveAndDeregister`

Randgalt: Ooh - TIL - I didn't know about that. Thanks. Looking into it now.

Randgalt: Phaser looks to be a kind of barrier. How would it be used in this context @TisonKun ?

TisonKun: It seems `OutstandingOps` is a one-shot barrier, so that as the code snippet above when we have new outstanding op we register a party and terminate the phaser at the first time it is advanced. Is it the semantic of `OutstandingOps`?

Randgalt: It's the same idea as in TreeCache. When the cache is started each outstanding operation increments a counter and when the op completes it decrements the counter. It's not exact, but it provides a way to handle the "cache is initialized" feature. Once the outstanding ops counter gets back to 0 it's no longer used.

TisonKun: Yes that is the same as things in my mind. Since `OutstandingOps` implemented correctly now, we can have a follow-up to see if it's worth we replace it with Java builtins and I will share a patch with description there. At most it is an "improvement" and should not block this PR to merge :)

For the "how", it seems to me that https://github.com/apache/curator/pull/335#discussion_r399591042 is clear.

curator-recipes/src/main/java/org/apache/curator/framework/recipes/cache/CuratorCacheImpl.java
```diff
@@ -0,0 +1,301 @@
...
+
+    private void rebuild()
+    {
+        if ( state.get() != State.STARTED )
+        {
+            return;
+        }
+
+        // rebuild from the root first
+        nodeChanged(path);
```
shayshim: This call is going to also get all data of children recursively, and update the storage accordingly, so why do we need to pass the storage again as below?

curator-recipes/src/main/java/org/apache/curator/framework/recipes/cache/CuratorCacheImpl.java
```diff
@@ -0,0 +1,301 @@
...
+import java.util.Set;
+import java.util.concurrent.atomic.AtomicReference;
+import java.util.function.Consumer;
+import java.util.stream.Stream;
+
+import static org.apache.curator.framework.recipes.cache.CuratorCacheListener.Type.*;
+import static org.apache.zookeeper.KeeperException.Code.NONODE;
+import static org.apache.zookeeper.KeeperException.Code.OK;
+
+class CuratorCacheImpl implements CuratorCache
```
shayshim: This cache does not allow to maintain only the immediate children? I noticed `parentPathFilter`, but shouldn't we allow that for efficiancy? (for cases of many grandchildren and so deeper that their updates are not interesting in this cache context.) Maybe not useful enough to bother, just asking.

curator-recipes/src/main/java/org/apache/curator/framework/recipes/cache/OutstandingOps.java
```diff
@@ -0,0 +1,55 @@
...
+    {
+        if ( active )
+        {
+            count.incrementAndGet();
+        }
+    }
+
+    void decrement()
+    {
+        if ( active && (count.decrementAndGet() == 0) )
```
eolivelli: Much better!

curator-recipes/src/main/java/org/apache/curator/framework/recipes/cache/CuratorCacheImpl.java
```diff
@@ -0,0 +1,301 @@
...
+                {
+                    handleException(event);
+                }
+                outstandingOps.decrement();
+            };
+
+            outstandingOps.increment();
+            if ( compressedData )
+            {
+                client.getData().decompressed().inBackground(callback).forPath(fromPath);
```
shayshim: General question about the inBackground callbacks - they are all going to be serialized on the single ZK event thread?

curator-recipes/src/main/java/org/apache/curator/framework/recipes/cache/CuratorCacheImpl.java
```diff
@@ -0,0 +1,301 @@
...
+    }
+
+    private void handleException(CuratorEvent event)
+    {
+        handleException(KeeperException.create(KeeperException.Code.get(event.getResultCode())));
+    }
+
+    private void handleException(Exception e)
+    {
+        ThreadUtils.checkInterrupted(e);
```
shayshim: In case of `InterruptedException` you both reset the interrupt on the thread and propagate the exception to user. Shouldn't you do only one of them, but not both?

curator-recipes/src/main/java/org/apache/curator/framework/recipes/cache/CuratorCacheImpl.java
Randgalt: Consider the case where the cache is in sync and then there is a network partition for a period of time. When the connection is restored, every node in the cache has to be re-considered as the new state of the server is not known. You can't just rebuild the cache from zero either as the client still needs to get events for changes. Neither can we just rebuild from the root node as checkChildrenChanged() stops if the cVersion is the same. E.g.

Pre network partition:

```
root
      Node A1
            Node B1
                Node C1
      Node A2
```

Post network partition:

```
root
      Node A1
            Node B1
                Node C2       --- C1 deleted, C2 is new
      Node A2
```

Rebuilding only from the Root would not detect that C1 was deleted and C2 was added because the cVersion of A1 would be the same.

curator-recipes/src/main/java/org/apache/curator/framework/recipes/cache/CuratorCacheImpl.java
Randgalt: If someone really needs to emulate PathChildrenCache it could be done with a custom implementation of `CuratorCacheStorage`. I thought it added too much weight to have APIs that only cache 1 level.

curator-recipes/src/main/java/org/apache/curator/framework/recipes/cache/CuratorCacheImpl.java
Randgalt: Yes, background callbacks are called from the single ZK event thread. Is there a concern?

curator-recipes/src/main/java/org/apache/curator/framework/recipes/cache/CuratorCacheImpl.java
Randgalt: It's important to reset the interrupted flag. I don't think we should count on the end user doing that. Curator has had this behavior for a long time. Maybe we can re-visit it in another issue.

curator-recipes/src/main/java/org/apache/curator/framework/recipes/cache/CuratorCacheImpl.java
```diff
@@ -0,0 +1,301 @@
...
+    public void start()
+    {
+        Preconditions.checkState(state.compareAndSet(State.LATENT, State.STARTED), "Already started");
+        persistentWatcher.start();
+    }
+
+    @Override
+    public void close()
+    {
+        if ( state.compareAndSet(State.STARTED, State.CLOSED) )
```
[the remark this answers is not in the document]
Randgalt: This pattern is used throughout Curator. It's actually an error to call `close()` without first calling `start()`. Maybe it should be an exception. We can deal with all instances of this pattern in a separate issue.

Randgalt:
> The patch is so huge.
> I don't feel I can give a real +1
> 
> I like this work.
> 
> I would suggest to create little patches for the future. this way reviews can be narrower.

This is already a broken up patch. There was a PR that preceded it. There will be another PR after this. So, this is as small as I could get this portion. Though, I guess I could have done just the PersistentWatcher class and then this.

Randgalt:
@shayshim I reworked the doc for `CuratorCache`. Please re-review.

curator-recipes/src/main/java/org/apache/curator/framework/recipes/cache/CuratorCacheImpl.java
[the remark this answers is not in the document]
shayshim: If that's the pattern everyone already used to, then I think it is better to leave it without change. It is minor consideration anyways, and always can be overridden with a boolean paired to the cache from the outside.

curator-recipes/src/main/java/org/apache/curator/framework/recipes/cache/CuratorCacheImpl.java
shayshim: No concern, just wanted to make sure. `addListener(T listener, Executor executor)` can be used to offload heavy listeners to another single threaded `Executor`.

curator-recipes/src/main/java/org/apache/curator/framework/recipes/cache/CuratorCacheImpl.java
Randgalt: Also - the new CuratorCache uses a Persistent watcher internally (which I added to ZK for 3.6). Those watchers only have 2 modes - single node watch or entire tree watch. So, the new CuratorCache matches this.

shayshim: Yeah, it is kind of limitation of the Persistent Watcher itself - not be able to be Persistent only for immediate children of the node. Potentially for client that needs only immediate children, it can make both ZK server and client work harder for potentially many/large not-interesting sub-trees events. But maybe in real life it is not useful enough.

curator-recipes/src/main/java/org/apache/curator/framework/recipes/cache/CuratorCacheListenerBuilder.java
```diff
@@ -0,0 +1,129 @@
...
+     */
+    CuratorCacheListenerBuilder forChanges(ChangeListener listener);
+
+    /**
+     * Add a listener only both {@link Type#NODE_CREATED} and {@link Type#NODE_CHANGED}
+     *
+     * @param listener listener to add
+     * @return this
+     */
+    CuratorCacheListenerBuilder forCreatesAndChanges(ChangeListener listener);
```
shayshim: Instead of having these multiple combinations (and the user will still need to check type for each event e.g. `type == NODE_CHANGED`) you could introduce interface that allows the user implement only the methods that are interesting for him/her, and have the burden of type -> action mapping specified in one location for all users:
`public interface CacheListener { 
  void onCacheInitialized();
  void onNodeCreate(ChildData data);
  void onNodeChanged(ChildData oldData, ChildData data);
  void onNodeDeleted(ChildData data);
}`
(you could introduce empty implementation for this interface for convenience).
It is approach that doesn't embrace lambdas, but in this case of multiple event types it might be more comfortable for users.

Randgalt: We could have both actually. We can add a `CacheListenerAdaptor` that looks like what you have. However, since Java 8, haven't most devs moved to the neo-functional style? The benefit of how I did the listener builder is that you don't have to use cumbersome anonymous classes. i.e. you can do:

```java
CuratorCacheListener.builder()
    .forDeletes(node -> handleDelete(node))
    .forInitialized(() -> handleInitialized())
    ... etc ...
```

This is superior to anonymous inner classes.

Randgalt: The adaptor would look like this. Should I add this? Is it gilding the Lilly? https://gist.github.com/Randgalt/cbe38cb3cdf1825f73dd165d997a2248

shayshim: No, I take it back. Didn't understand fully your builder until now.

PR #336, opened 2019-11-03
[CURATOR-549] Bridge facade for ModeledCache and ServiceCache, etc.

Randgalt:
Note: this PR is based off #335 

It would be nice if Curator's internal use of caches got the benefit of CuratorCache/Persistent Watches. This PR creates a simple bridge that, when using ZK 3.6.0 creates a `CuratorCache`, and for earlier versions creates a `TreeCache`. The `curator-test-zk35` module ensures that both code paths are tested.

Important files:

- [CompatibleCuratorCacheBridge.java](https://github.com/apache/curator/blob/d16271a644d62f8bdd9470dcb1cc30fafdc3f247/curator-recipes/src/main/java/org/apache/curator/framework/recipes/cache/CompatibleCuratorCacheBridge.java) - this is the wrapper around TreeCache
- [CuratorCacheBridgeBuilderImpl.java](https://github.com/apache/curator/blob/d16271a644d62f8bdd9470dcb1cc30fafdc3f247/curator-recipes/src/main/java/org/apache/curator/framework/recipes/cache/CuratorCacheBridgeBuilderImpl.java) - the builder that returns either a CuratorCache or a CompatibleCuratorCacheBridge
- [GroupMember.java](https://github.com/apache/curator/blob/d16271a644d62f8bdd9470dcb1cc30fafdc3f247/curator-recipes/src/main/java/org/apache/curator/framework/recipes/nodes/GroupMember.java) - updated GroupMember recipe that uses the bridge
- [ModeledCacheImpl.java](https://github.com/apache/curator/blob/d16271a644d62f8bdd9470dcb1cc30fafdc3f247/curator-x-async/src/main/java/org/apache/curator/x/async/modeled/details/ModeledCacheImpl.java) - updated ModeledCacheImpl that uses the bridge
- [ServiceCacheImpl.java](https://github.com/apache/curator/blob/d16271a644d62f8bdd9470dcb1cc30fafdc3f247/curator-x-discovery/src/main/java/org/apache/curator/x/discovery/details/ServiceCacheImpl.java) - updated ServiceCacheImpl that uses the bridge
- [ServiceDiscoveryImpl.java](https://github.com/apache/curator/blob/d16271a644d62f8bdd9470dcb1cc30fafdc3f247/curator-x-discovery/src/main/java/org/apache/curator/x/discovery/details/ServiceDiscoveryImpl.java) - updated ServiceDiscoveryImpl that uses the bridge

### DIFF
--- a/curator-recipes/src/main/java/org/apache/curator/framework/recipes/cache/CuratorCacheBridge.java
+++ b/curator-recipes/src/main/java/org/apache/curator/framework/recipes/cache/CuratorCacheBridge.java
@@ -1,0 +1,36 @@
+/**
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package org.apache.curator.framework.recipes.cache;
+
+/**
+ * A facade that uses {@link org.apache.curator.framework.recipes.cache.CuratorCache} if
+ * persistent watches are available or a {@link org.apache.curator.framework.recipes.cache.TreeCache}
+ * otherwise
+ */
+@SuppressWarnings("deprecation")
+public interface CuratorCacheBridge extends CuratorCache
+{
+    /**
+     * Returns true if the underlying cache is {@link org.apache.curator.framework.recipes.cache.CuratorCache} (i.e. ZooKeeper 3.6+).
+     * Otherwise it is {@link org.apache.curator.framework.recipes.cache.TreeCache} (i.e. ZooKeeper 3.5.x)
+     *
+     * @return true/false
+     */
+    boolean isCuratorCache();
+}
--- a/curator-recipes/src/main/java/org/apache/curator/framework/recipes/cache/CuratorCacheBridgeBuilder.java
+++ b/curator-recipes/src/main/java/org/apache/curator/framework/recipes/cache/CuratorCacheBridgeBuilder.java
@@ -1,0 +1,57 @@
+/**
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+package org.apache.curator.framework.recipes.cache;
+
+import java.util.concurrent.ExecutorService;
+
+public interface CuratorCacheBridgeBuilder
+{
+    /**
+     * @param options any options
+     * @return this
+     */
+    CuratorCacheBridgeBuilder withOptions(CuratorCache.Options... options);
+
+    /**
+     * The bridge cache will not retain the data bytes. i.e. ChildData objects
+     * returned by the cache will always return {@code null} for {@link ChildData#getData()}
+     *
+     * @return this
+     */
+    CuratorCacheBridgeBuilder withDataNotCached();
+
+    /**
+     * If the old {@link org.apache.curator.framework.recipes.cache.TreeCache} is used by the bridge
+     * (i.e. you are using ZooKeeper 3.5.x) then this executor service is passed to {@link org.apache.curator.framework.recipes.cache.TreeCache.Builder#setExecutor(java.util.concurrent.ExecutorService)}.
+     * For {@link org.apache.curator.framework.recipes.cache.CuratorCache} this is not used and will be ignored (a warning will be logged).
+     *
+     * @param executorService executor to use for ZooKeeper 3.5.x
+     * @return this
+     */
+    @SuppressWarnings("deprecation")
+    CuratorCacheBridgeBuilder withExecutorService(ExecutorService executorService);
+
+    /**
+     * Return a new Curator Cache Bridge based on the builder methods that have been called
+     *
+     * @return new Curator Cache Bridge
+     */
+    CuratorCacheBridge build();
+}
--- a/curator-recipes/src/main/java/org/apache/curator/framework/recipes/cache/CuratorCacheBridgeBuilderImpl.java
+++ b/curator-recipes/src/main/java/org/apache/curator/framework/recipes/cache/CuratorCacheBridgeBuilderImpl.java
@@ -1,0 +1,77 @@
+/**
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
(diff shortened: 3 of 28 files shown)


Randgalt:
attn: @cammckenzie @shayshim @TisonKun @eolivelli

curator-recipes/src/main/java/org/apache/curator/framework/recipes/cache/CuratorCacheBridgeBuilder.java
```diff
@@ -0,0 +1,57 @@
...
+     */
+    CuratorCacheBridgeBuilder withOptions(CuratorCache.Options... options);
+
+    /**
+     * The bridge cache will not retain the data bytes. i.e. ChildData objects
+     * returned by the cache will always return {@code null} for {@link ChildData#getData()}
+     *
+     * @return this
+     */
+    CuratorCacheBridgeBuilder withBytesNotCached();
```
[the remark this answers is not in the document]
Randgalt: Yeah - sure I can rename it.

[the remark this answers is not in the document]
cammckenzie: I just noticed that it is named that all through the CuratorCache implementation as well. Maybe it's not worth bothering with. Having withDataNotCached, may be a confusing name anyway.

curator-recipes/src/main/java/org/apache/curator/framework/recipes/cache/CuratorCacheBridge.java
```diff
@@ -0,0 +1,36 @@
...
+@SuppressWarnings("deprecation")
+public interface CuratorCacheBridge extends CuratorCache
+{
+    /**
+     * Returns true if the underlying cache is {@link org.apache.curator.framework.recipes.cache.CuratorCache} (i.e. ZooKeeper 3.6+).
+     * Otherwise it is {@link org.apache.curator.framework.recipes.cache.TreeCache} (i.e. ZooKeeper 3.5.x)
+     *
+     * @return true/false
+     */
+    boolean isCuratorCache();
```
shayshim, line 35: I don't see why this method is necessary? Why would users need to know that?

curator-recipes/src/main/java/org/apache/curator/framework/recipes/cache/CuratorCacheBridgeBuilderImpl.java
```diff
@@ -0,0 +1,77 @@
...
+    }
+
+    @Override
+    public CuratorCacheBridge build()
+    {
+        if ( !forceTreeCache && Compatibility.hasPersistentWatchers() )
+        {
+            if ( executorService != null )
+            {
+                LoggerFactory.getLogger(getClass()).warn("CuratorCache does not support custom ExecutorService");
```
shayshim, line 70: Consider advising in the warning that custom executors can be used through `Listenable.addListener(T listener, Executor executor);`

curator-recipes/src/main/java/org/apache/curator/framework/recipes/nodes/GroupMember.java
```diff
@@ -121,8 +128,11 @@ public void close()
     {
         ImmutableMap.Builder<String, byte[]> builder = ImmutableMap.builder();
         boolean thisIdAdded = false;
-        for ( ChildData data : cache.getCurrentData() )
+
+        Iterator<ChildData> iterator = cache.stream().filter(parentPathFilter(membershipPath)).iterator();
```
shayshim, line 132: Not sure if it is better, just want to mention you could stay with the more pleasant for each syntax by relying on `Iterable` like that:
Iterable<ChildData> iterable = getChildDataIterable();
        for (ChildData data: iterable)
While `getChildDataIterable` is:
```
private Iterable<ChildData> getChildDataIterable() {
        Stream<ChildData> onlyMembers = cache.stream().filter(parentPathFilter(membershipPath));
        return onlyMembers::iterator;
    }
```

curator-recipes/src/test/java/org/apache/curator/framework/recipes/cache/TestCuratorCacheBridge.java
```diff
@@ -0,0 +1,65 @@
...
+import org.apache.curator.framework.CuratorFrameworkFactory;
+import org.apache.curator.retry.RetryOneTime;
+import org.apache.curator.test.compatibility.CuratorTestBase;
+import org.apache.curator.utils.Compatibility;
+import org.testng.Assert;
+import org.testng.annotations.Test;
+
+public class TestCuratorCacheBridge extends CuratorTestBase
+{
+    public void testImplementationSelection()
```
shayshim, line 32: Did you forget to add `@Test`?

curator-recipes/src/test/java/org/apache/curator/framework/recipes/nodes/TestGroupMember.java
```diff
@@ -25,11 +25,13 @@
 import org.apache.curator.retry.RetryOneTime;
 import org.apache.curator.test.BaseClassForTests;
 import org.apache.curator.test.Timing;
+import org.apache.curator.test.compatibility.CuratorTestBase;
 import org.apache.curator.utils.CloseableUtils;
 import org.testng.Assert;
 import org.testng.annotations.Test;
 import java.util.Map;
 
+@Test(groups = CuratorTestBase.zk35TestCompatibilityGroup)
```
shayshim, line 34: Can you remind why do we need it now? It will make the test run if this env has 3.5 and up?

curator-x-async/src/main/java/org/apache/curator/x/async/modeled/details/ModeledCacheImpl.java
```diff
@@ -42,11 +46,12 @@
 
 class ModeledCacheImpl<T> implements TreeCacheListener, ModeledCache<T>
 {
-    private final TreeCache cache;
+    private final CuratorCacheBridge cache;
     private final Map<ZPath, Entry<T>> entries = new ConcurrentHashMap<>();
     private final ModelSerializer<T> serializer;
     private final StandardListenerManager<ModeledCacheListener<T>> listenerContainer = StandardListenerManager.standard();
     private final ZPath basePath;
+    private final EnsureContainers ensureContainers;
```
shayshim, line 54: This is more question for the previous review, but why `TreeCache` support creating the parent nodes of the path (as containers) and `CuratorCache` isn't?

curator-x-async/src/main/java/org/apache/curator/x/async/modeled/details/ModeledCacheImpl.java
```diff
@@ -148,38 +165,32 @@ public void childEvent(CuratorFramework client, TreeCacheEvent event)
         }
     }
 
-    private void internalChildEvent(TreeCacheEvent event) throws Exception
+    private void internalChildEvent(TreeCacheEvent event)
```
shayshim, line 168: In case of `TreeCache`, this code can called from `Executor`, but in case of `CuratorCache`, it can be called only on the event thread. No way it can be an issue? 
It could add itself as listener also using `addListener(T listener, Executor executor)`

curator-recipes/src/main/java/org/apache/curator/framework/recipes/cache/CuratorCacheBridge.java
Randgalt: Technically, no. TBH this class will mostly be used internally.

curator-recipes/src/test/java/org/apache/curator/framework/recipes/cache/TestCuratorCacheBridge.java
Randgalt: Good catch

curator-recipes/src/test/java/org/apache/curator/framework/recipes/nodes/TestGroupMember.java
Randgalt: We now have a (new) curator-test-zk35 module that tests ZooKeeper 3.5.x compatibility. This release will use ZooKeeper 3.6.x as the base.

curator-x-async/src/main/java/org/apache/curator/x/async/modeled/details/ModeledCacheImpl.java
Randgalt: I think it was copying PathChildrenCache. With PersistentWatchers it's not necessary to have the ZNodes exist as the watcher is independent. So, it's not needed with CuratorCache.

curator-x-async/src/main/java/org/apache/curator/x/async/modeled/details/ModeledCacheImpl.java
Randgalt: > but in case of CuratorCache, it can be called only on the event thread

No, it isn't. Listeners are called via `client.runSafe(() -> listenerManager.forEach(proc));`

PR #338, opened 2019-11-04
[CURATOR-549] Prep for Persistent Watchers by moving to ZK 3.6.0

Randgalt:
### 3.6.0 has many changes

ZooKeeper 3.6.0 has some significant changes from previous versions and this PR is only bringing Curator up to date. Some of these changes are:

- The reconfig APIs have moved into a new class, `ZooKeeperAdmin`. This class existed in 3.5.x but wasn't required. Now it is. 
- `DefaultZookeeperFactory` now checks `Compatibility.hasZooKeeperAdmin()` and creates a `ZooKeeperAdmin` if it's available
- A bunch of little things changed in the ZK server code which affected Curator's test classes. I moved it all into reflection based calls in `Compatibility.java` in the test module. We now have modules that test ZK 3.4, 3.5 and 3.6 so we're safe with compatibility.
- ZooKeeper's `MultiTransactionRecord` has been removed it seems. That forced `CuratorMultiTransactionRecord` to be re-written. It's not a public class so hopefully it won't affect anyone.

### New module

There is a new module, `curator-test-zk35`. It forces ZooKeeper 3.5.6 and performs selected tests from the other modules to ensure compatibility. Tests annotated with TestNG groups `zk35` and `zk35Compatibility` are tested. Group `zk36` is excluded. Note: these tests will only run from Maven. I don't think IntelliJ/Eclipse support the Maven syntax I used.

### 3 More PRs coming

I have 3 more PRs that will be separate from this to avoid too much review pain:

- #334
- #335 
- #336

### DIFF
--- a/curator-client/src/main/java/org/apache/curator/utils/Compatibility.java
+++ b/curator-client/src/main/java/org/apache/curator/utils/Compatibility.java
@@ -1,0 +1,104 @@
+/**
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+package org.apache.curator.utils;
+
+import org.apache.zookeeper.server.quorum.QuorumPeer;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import java.lang.reflect.Field;
+import java.lang.reflect.Method;
+import java.net.InetSocketAddress;
+
+/**
+ * Utils to help with ZK version compatibility
+ */
+public class Compatibility
+{
+    private static final Method getReachableOrOneMethod;
+    private static final Field addrField;
+
+    private static final Logger logger = LoggerFactory.getLogger(Compatibility.class);
+
+    static
+    {
+        Method localGetReachableOrOneMethod;
+        try
+        {
+            Class<?> multipleAddressesClass = Class.forName("org.apache.zookeeper.server.quorum.MultipleAddresses");
+            localGetReachableOrOneMethod = multipleAddressesClass.getMethod("getReachableOrOne");
+            LoggerFactory.getLogger(Compatibility.class).info("Using org.apache.zookeeper.server.quorum.MultipleAddresses");
+        }
+        catch ( ReflectiveOperationException ignore )
+        {
+            localGetReachableOrOneMethod = null;
+        }
+        getReachableOrOneMethod = localGetReachableOrOneMethod;
+
+        Field localAddrField;
+        try
+        {
+            localAddrField = QuorumPeer.QuorumServer.class.getField("addr");
+        }
+        catch ( NoSuchFieldException e )
+        {
+            localAddrField = null;
+            LoggerFactory.getLogger(Compatibility.class).error("Could not get addr field! Reconfiguration fail!");
+        }
+        addrField = localAddrField;
+    }
+
+    public static boolean hasGetReachableOrOneMethod()
+    {
+        return (getReachableOrOneMethod != null);
+    }
+
+    public static boolean hasAddrField()
+    {
+        return (addrField != null);
+    }
+
+    public static String getHostAddress(QuorumPeer.QuorumServer server)
+    {
+        InetSocketAddress address = null;
+        if ( getReachableOrOneMethod != null )
+        {
+            try
+            {
+                address = (InetSocketAddress)getReachableOrOneMethod.invoke(server.addr);
+            }
+            catch ( Exception e )
+            {
+                logger.error("Could not call getReachableOrOneMethod.invoke({})", server.addr, e);
+            }
+        }
+        else if (addrField != null)
+        {
+            try
+            {
+                address = (InetSocketAddress)addrField.get(server);
+            }
+            catch ( Exception e )
+            {
+                logger.error("Could not call addrField.get({})", server, e);
+            }
+        }
+        return (address != null) ? address.getAddress().getHostAddress() : "unknown";
+    }
+}
--- a/curator-framework/src/main/java/org/apache/curator/framework/imps/CuratorMultiTransactionRecord.java
+++ b/curator-framework/src/main/java/org/apache/curator/framework/imps/CuratorMultiTransactionRecord.java
@@ -16,49 +16,57 @@
  * specific language governing permissions and limitations
  * under the License.
  */
+
 package org.apache.curator.framework.imps;
 
 import com.google.common.collect.Lists;
 import org.apache.curator.framework.api.transaction.OperationType;
 import org.apache.curator.framework.api.transaction.TypeAndPath;
-import org.apache.zookeeper.MultiTransactionRecord;
 import org.apache.zookeeper.Op;
 import java.security.MessageDigest;
+import java.util.ArrayList;
+import java.util.Iterator;
 import java.util.List;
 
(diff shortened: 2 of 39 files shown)


curator-framework/src/main/java/org/apache/curator/framework/imps/CuratorFrameworkImpl.java
```diff
@@ -620,6 +621,15 @@ ZooKeeper getZooKeeper() throws Exception
         return client.getZooKeeper();
     }
 
+    Object getZooKeeperAdmin() throws Exception
```
shayshim: If we return client.getZooKeeper(), just like in getZooKeeper(), then this method seems unneeded to me. Clients should call getZooKeeper() and cast to ZooKeeperAdmin if they need (like in ReconfigBuilderImpl). 
You also could promise in its javadoc to return ZooKeeperAdmin for versions 3.5 and up.

Randgalt: This method does the `isZk34CompatibilityMode()` check though. You don't see value in that?

shayshim: Only if the caller knows it has 3.5+ then it will be able to cast to ZooKeeperAdmin (which it currently has to anyways) and consequently use it. If it has 3.4 - it won't be able to cast - at compile time (no such class), which is safer.

curator-framework/src/main/java/org/apache/curator/framework/imps/CuratorFrameworkImpl.java
```diff
@@ -830,6 +840,12 @@ public boolean isZk34CompatibilityMode()
         return zk34CompatibilityMode;
     }
 
+    @Override
+    public boolean isZk35CompatibilityMode()
+    {
+        return !zk34CompatibilityMode && !Compatibility.hasPersistentWatchers();
```
shayshim: BTW why do we need zk34CompatibilityMode as insatnce field? Couldn't we call Compatibility.isZK34() instead?

Randgalt: I forget TBH. As I recall, I wanted users to be able to force 3.4 mode regardless of the library. So, someone could use ZK 3.5 client library against a 3.4 server. But this likely isn't important anymore. We should probably consider removing 3.4 support altogether at some point.

curator-framework/src/main/java/org/apache/curator/framework/imps/CuratorMultiTransactionRecord.java
```diff
@@ -16,49 +16,57 @@
...
 {
-    private final List<TypeAndPath>     metadata = Lists.newArrayList();
-
-    @Override
-    public final void add(Op op)
-    {
-        throw new UnsupportedOperationException();
-    }
+    private final List<TypeAndPath> metadata = Lists.newArrayList();
+    private List<Op> ops = new ArrayList<>();
```
shayshim: Can be final too

Randgalt:
FYI - Please DO NOT MERGE this yet

cammckenzie:
Looks ok to me.

curator-test/src/main/java/org/apache/curator/test/compatibility/CuratorTestBase.java
```diff
@@ -21,8 +21,11 @@
 import org.apache.curator.test.BaseClassForTests;
 import org.testng.annotations.Listeners;
 
-@Listeners(Zk35MethodInterceptor.class)
 public class CuratorTestBase extends BaseClassForTests
 {
+    public static final String zk35Group = "zk35";
+    public static final String zk36Group = "zk36";
+    public static final String zk35CompatibilityGroup = "zk35Compatibility";
```
shayshim: What's the difference between zk35 and zk35CompatibilityGroup?

curator-test-zk34/pom.xml
```diff
@@ -160,7 +160,7 @@
                         <dependency>org.apache.curator:curator-framework</dependency>
```
shayshim: Didn't happen in this PR, but this file has 2 curator-recipes dependencies

shayshim: Is there a reason why we don't have curator-x-async dependency here?

shayshim:
Looks fine to me

curator-test/src/main/java/org/apache/curator/test/compatibility/CuratorTestBase.java
Randgalt: `zk35` is for tests that only work with ZooKeeper 3.5 or above (it's a pre-existing group). `zk35CompatibilityGroup` is a new group for tests that should be run by the `curator-test-zk35` module. I'll try to come up with better names.

curator-test-zk34/pom.xml
Randgalt: It's not tested for zk34 compatibility. We don't make any guarantees that that module works with zk 3.4

Randgalt:
@shayshim @cammckenzie @TisonKun this PR is ready for review/re-review

curator-client/src/main/java/org/apache/curator/utils/Compatibility.java
```diff
@@ -0,0 +1,104 @@
...
+
+        Field localAddrField;
+        try
+        {
+            localAddrField = QuorumPeer.QuorumServer.class.getField("addr");
+        }
+        catch ( NoSuchFieldException e )
+        {
+            localAddrField = null;
+            LoggerFactory.getLogger(Compatibility.class).error("Could not get addr field! Reconfiguration fail!");
```
TisonKun:
```suggestion
            logger.error("Could not get addr field! Reconfiguration fail!");
```

And most where in Curator we call `log`.

curator-framework/src/main/java/org/apache/curator/framework/imps/ReconfigBuilderImpl.java
```diff
@@ -268,7 +269,7 @@ public void processResult(int rc, String path, Object ctx, byte[] bytes, Stat st
                     client.processBackgroundOperation(data, event);
                 }
             };
-            client.getZooKeeper().reconfig(joining, leaving, newMembers, fromConfig, callback, backgrounding.getContext());
+            ((ZooKeeperAdmin)client.getZooKeeper()).reconfigure(joining, leaving, newMembers, fromConfig, callback, backgrounding.getContext());
```
TisonKun: Shall we change the return type of `getZooKeeper` to `ZooKeeperAdmin`?

I don't see an instantiate code path of `NonAdminZookeeperFactory` and from 3.5.3 there is always `ZooKeeperAdmin`. From another perspective, what if `getZooKeeper` didn't return `ZooKeeperAdmin` and we reach to this line? It seems the program exits exceptionally.

curator-recipes/src/test/java/org/apache/curator/framework/recipes/leader/TestLeaderSelectorEdges.java
```diff
@@ -26,6 +26,7 @@
 import org.apache.curator.framework.state.ConnectionState;
 import org.apache.curator.retry.RetryNTimes;
 import org.apache.curator.test.BaseClassForTests;
+import org.apache.curator.test.compatibility.CuratorTestBase;
```
TisonKun: Unused import.

curator-test/src/main/java/org/apache/curator/test/BaseClassForTests.java
```diff
@@ -123,7 +123,16 @@ public void setup() throws Exception
...
-        createServer();
+        try
+        {
+            createServer();
+        }
+        catch ( FailedServerStartException ignore )
+        {
+            log.warn("Failed to start server - retrying");
+            closeServer();
+            createServer();
```
TisonKun: Nice to document that we try exactly one more time for a best effort flaky test auto-recovery. It is confused at the first glance(why retry exactly once) and over-engineering if we set the retry time as a field or even configurable.

curator-framework/src/main/java/org/apache/curator/framework/imps/ReconfigBuilderImpl.java
Randgalt: No - via the `ZookeeperFactory` clients can change which ZooKeeper class is returned. In fact, I added `NonAdminZookeeperFactory` that returns the non-admin ZooKeeper. Of course, clients will need to know they can't use the reconfig APIs when this is done.

Randgalt:
Closing in favor of #350 - this should be a Pr for CURATOR-558

PR #341, opened 2019-12-20
[CURATOR-553] Use slf4j place-holder instead String.format and concat.

PhantomThief:


tisonkun (commented):
Thanks for your contribution @PhantomThief ! The patch looks good. Could you rebase the patch onto master? I'll give it a review then and you can ping me after rebase.

tisonkun:
Closed as inactive. This patch is good to continue, but cannot be merged as is. Feel free to create a new one based on the latest master.

PR #343, opened 2020-01-27
[CURATOR-557] ServiceCacheImpl does not close ExecutorService

naude-r:
- ServiceCacheImpl should allow the self-created executor service to be closed
- Allow an optional executor service for ServiceProvider

fix #CURATOR-557

### DIFF
--- a/curator-x-discovery/src/main/java/org/apache/curator/x/discovery/ServiceCacheBuilder.java
+++ b/curator-x-discovery/src/main/java/org/apache/curator/x/discovery/ServiceCacheBuilder.java
@@ -45,7 +45,9 @@ public interface ServiceCacheBuilder<T>
      *
      * @param threadFactory factory
      * @return this
+     * @deprecated use {@link #executorService(ExecutorService)} instead
      */
+    @Deprecated
     public ServiceCacheBuilder<T> threadFactory(ThreadFactory threadFactory);
 
     /**
--- a/curator-x-discovery/src/main/java/org/apache/curator/x/discovery/ServiceProviderBuilder.java
+++ b/curator-x-discovery/src/main/java/org/apache/curator/x/discovery/ServiceProviderBuilder.java
@@ -18,7 +18,9 @@
  */
 package org.apache.curator.x.discovery;
 
+import org.apache.curator.utils.CloseableExecutorService;
 import org.apache.curator.x.discovery.strategies.RoundRobinStrategy;
+import java.util.concurrent.ExecutorService;
 import java.util.concurrent.ThreadFactory;
 
 public interface ServiceProviderBuilder<T>
@@ -47,11 +49,14 @@ public interface ServiceProviderBuilder<T>
     public ServiceProviderBuilder<T> providerStrategy(ProviderStrategy<T> providerStrategy);
 
     /**
-     * optional - the thread factory to use for creating internal threads
+     * optional - the thread factory to use for creating internal threads. The specified ThreadFactory overrides
+     * any prior ThreadFactory or ClosableExecutorService set on the ServiceProviderBuilder
      *
      * @param threadFactory factory to use
      * @return this
+     * @deprecated use {@link #executorService(ExecutorService)} instead
      */
+    @Deprecated
     public ServiceProviderBuilder<T> threadFactory(ThreadFactory threadFactory);
 
     /**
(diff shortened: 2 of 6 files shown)


curator-x-discovery/src/main/java/org/apache/curator/x/discovery/ServiceCacheBuilder.java
```diff
@@ -46,6 +46,7 @@
      * @param threadFactory factory
      * @return this
      */
```
shayshim: Please add to the Javadoc:
`* @deprecated use {@link #executorService(ExecutorService)} instead`

curator-x-discovery/src/main/java/org/apache/curator/x/discovery/ServiceProviderBuilder.java
```diff
@@ -71,4 +74,23 @@
      * @return this
      */
     public ServiceProviderBuilder<T> additionalFilter(InstanceFilter<T> filter);
+
+    /**
+     * Optional ExecutorService to use for the cache's background thread. The specified ExecutorService
+     * will be wrapped in a CloseableExecutorService and overrides any prior ThreadFactory or ExecutorService
```
shayshim: Should be `... and overrides any prior ThreadFactory or CloseableExecutorService` (instead of `ExecutorService`), no?
Same question for next method.

curator-x-discovery/src/main/java/org/apache/curator/x/discovery/ServiceProviderBuilder.java
```diff
@@ -52,6 +54,7 @@
      * @param threadFactory factory to use
      * @return this
      */
```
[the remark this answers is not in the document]
naude-r: done

curator-x-discovery/src/main/java/org/apache/curator/x/discovery/details/ServiceProviderBuilderImpl.java
```diff
@@ -102,4 +105,18 @@
         filters.add(filter);
         return this;
     }
+
```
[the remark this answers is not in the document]
naude-r: done

curator-x-discovery/src/main/java/org/apache/curator/x/discovery/details/ServiceProviderImpl.java
```diff
@@ -43,13 +46,21 @@
     private final ProviderStrategy<T> providerStrategy;
     private final DownInstanceManager<T> downInstanceManager;
 
-    public ServiceProviderImpl(ServiceDiscoveryImpl<T> discovery, String serviceName, ProviderStrategy<T> providerStrategy, ThreadFactory threadFactory, List<InstanceFilter<T>> filters, DownInstancePolicy downInstancePolicy)
+    public ServiceProviderImpl(ServiceDiscoveryImpl<T> discovery, String serviceName, ProviderStrategy<T> providerStrategy, ThreadFactory threadFactory, CloseableExecutorService executorService, List<InstanceFilter<T>> filters, DownInstancePolicy downInstancePolicy)
```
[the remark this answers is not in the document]
naude-r: done. constructor was made protected so that:
* it can be extended in future
* ServiceProviderBuilderImpl can access it.

curator-x-discovery/src/main/java/org/apache/curator/x/discovery/ServiceCacheBuilder.java
naude-r: done

curator-x-discovery/src/main/java/org/apache/curator/x/discovery/ServiceProviderBuilder.java
naude-r: done for both.

naude-r:
@shayshim changes made. have not squashed the commit. mostly so that review comments stay intact. can do this if required.

curator-x-discovery/src/main/java/org/apache/curator/x/discovery/ServiceCacheBuilder.java
shayshim: Sorry if I wasn't clear, but I didn't mean to drop the `@Deprecated`. `@Deprecated` and `@deprecated` should be used together. Please refer to:
https://docs.oracle.com/javase/8/docs/technotes/guides/javadoc/deprecation/deprecation.html
Their conclusion is "Thus, using the `@Deprecated` annotation to generate warnings is more portable that relying on the `@deprecated` Javadoc tag."

Same for `ServiceProviderBuilder.threadFactory(...)`.

shayshim: Also, it turns out that deprecation of interface method doesn't affect overriding methods, as said in the same link:
"It is possible for a non-deprecated property to hide or override a deprecated one, which removes deprecation. As developer of an API, it is your responsibility to deprecate overrides of a deprecated method, if in fact they should be deprecated."
Could you deprecate also the two `threadFactory(...)` overriding methods?

shayshim:
@naude-r yes, please squash the commits before your next push. Thanks!

curator-x-discovery/src/main/java/org/apache/curator/x/discovery/ServiceCacheBuilder.java
naude-r: done.

naude-r:
@shayshim review comments addressed and commits squashed.

PR #345, opened 2020-02-03
[CURATOR-551] Fix Handle Holder new connection string

Randgalt:
Commit 26364c6186fc7c09a9462557b1ca791e9aa70006 (Sat Sep 26 13:13:02 2015) changed HandleHolder.getNewConnectionString() was changed to return the new connection string instead of just a boolean. I believe the value returned should have been ensembleProvider.getConnectionString() not helper.getConnectionString(). TBH I no longer remember the genesis of this change but I can't make the current implementation make sense.

Additionally, a change was made to optionally call zooKeeper.updateServerList(). When this path is taken the handle holder's connection needs to be updated as well or we'll get an infinite loop of changes. The path that runs when ensembleProvider.updateServerListEnabled() is false ends up setting handle holder's connection to ensembleProvider.getConnectionString().

I'm loathe to make such low level changes in code that's existed for a long time. But, my investigation shows that this is how it should be. Hopefully, users can do testing.

### DIFF
--- a/curator-client/src/main/java/org/apache/curator/ConnectionState.java
+++ b/curator-client/src/main/java/org/apache/curator/ConnectionState.java
@@ -48,7 +48,7 @@ class ConnectionState implements Watcher, Closeable
     private static final int MAX_BACKGROUND_EXCEPTIONS = 10;
     private static final boolean LOG_EVENTS = Boolean.getBoolean(DebugUtils.PROPERTY_LOG_EVENTS);
     private static final Logger log = LoggerFactory.getLogger(ConnectionState.class);
-    private final HandleHolder zooKeeper;
+    private final HandleHolder handleHolder;
     private final AtomicBoolean isConnected = new AtomicBoolean(false);
     private final AtomicInteger lastNegotiatedSessionTimeoutMs = new AtomicInteger(0);
     private final EnsembleProvider ensembleProvider;
@@ -73,7 +73,7 @@ class ConnectionState implements Watcher, Closeable
             parentWatchers.offer(parentWatcher);
         }
 
-        zooKeeper = new HandleHolder(zookeeperFactory, this, ensembleProvider, sessionTimeoutMs, canBeReadOnly);
+        handleHolder = new HandleHolder(zookeeperFactory, this, ensembleProvider, sessionTimeoutMs, canBeReadOnly);
     }
 
     ZooKeeper getZooKeeper() throws Exception
@@ -96,7 +96,7 @@ class ConnectionState implements Watcher, Closeable
             checkTimeouts();
         }
 
-        return zooKeeper.getZooKeeper();
+        return handleHolder.getZooKeeper();
     }
 
     boolean isConnected()
@@ -122,7 +122,7 @@ class ConnectionState implements Watcher, Closeable
         CloseableUtils.closeQuietly(ensembleProvider);
         try
         {
-            zooKeeper.closeAndClear(waitForShutdownTimeoutMs);
+            handleHolder.closeAndClear(waitForShutdownTimeoutMs);
         }
         catch ( Exception e )
         {
@@ -173,7 +173,7 @@ class ConnectionState implements Watcher, Closeable
                 connectionStartMs = System.currentTimeMillis();
                 if ( newIsConnected )
                 {
-                    lastNegotiatedSessionTimeoutMs.set(zooKeeper.getNegotiatedSessionTimeoutMs());
+                    lastNegotiatedSessionTimeoutMs.set(handleHolder.getNegotiatedSessionTimeoutMs());
                     log.debug("Negotiated session timeout: " + lastNegotiatedSessionTimeoutMs.get());
                 }
             }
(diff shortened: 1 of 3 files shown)


curator-client/src/main/java/org/apache/curator/ConnectionState.java
```diff
@@ -48,7 +48,7 @@
     private static final int MAX_BACKGROUND_EXCEPTIONS = 10;
     private static final boolean LOG_EVENTS = Boolean.getBoolean(DebugUtils.PROPERTY_LOG_EVENTS);
     private static final Logger log = LoggerFactory.getLogger(ConnectionState.class);
-    private final HandleHolder zooKeeper;
+    private final HandleHolder handleHolder;
```
Randgalt: Note: renamed for clarity

curator-client/src/main/java/org/apache/curator/HandleHolder.java
```diff
@@ -34,15 +34,6 @@
 
     private volatile Helper helper;
 
-    private interface Helper
```
Randgalt: Moved to a top level class so imp can be shared and it's easier to read.

curator-client/src/main/java/org/apache/curator/HandleHolder.java
```diff
@@ -70,7 +61,16 @@ String  getConnectionString()
     String getNewConnectionString()
     {
         String helperConnectionString = (helper != null) ? helper.getConnectionString() : null;
-        return ((helperConnectionString != null) && !ensembleProvider.getConnectionString().equals(helperConnectionString)) ? helperConnectionString : null;
+        String ensembleProviderConnectionString = ensembleProvider.getConnectionString();
```
Randgalt: Note: this is the major change

curator-client/src/main/java/org/apache/curator/ConnectionState.java
```diff
@@ -356,6 +356,7 @@ private void handleNewConnectionString(String newConnectionString)
                 if ( ensembleProvider.updateServerListEnabled() )
                 {
                     zooKeeper.updateServerList(newConnectionString);
+                    handleHolder.resetConnectionString(newConnectionString);
```
Randgalt: Also here - this is the meat of the change

curator-client/src/main/java/org/apache/curator/HandleHolder.java
shayshim: I have a question. `helper` is `null` until `closeAndReset()` is called, and new connection string can be sensed only if `helper` is not `null`. Isn't it possible to have new connection string (from the `EnsembleTracker`) before `closeAndReset()` was ever called? If it is possible, then it seems like the new connection string won't be handled, since `helper` would be `null`.

Randgalt: Thanks @shayshim I'll look into this possiblity

cammckenzie: The whole flow of how things interact is very confusing, but I don't think that this is an issue. `closeAndReset()` gets called from `ConnectionState.reset()` which gets called from the `ConnectionState.start()` method.  I'm not 100% sure there's not a race condition here, but from what I can see, it seems improbable.

Randgalt: > whole flow of how things interact is very confusing

Too true. What do we do? Should we attempt a more aggressive change?

shayshim: I see, thanks @cammckenzie

cammckenzie: Maybe we should just merge this change and look at cleaning up this stuff in a future PR? I think it would be a pretty significant change.

Randgalt: I'm OK with that. @shayshim ? Maybe you can both approve this PR when you're OK

Randgalt: > I have a question. `helper` is `null` until `closeAndReset()` is called, and new connection string can be sensed only if `helper` is not `null`. Isn't it possible to have new connection string (from the `EnsembleTracker`) before `closeAndReset()` was ever called? If it is possible, then it seems like the new connection string won't be handled, since `helper` would be `null`.

`closeAndReset()` will cause `ensembleProvider.getConnectionString()` to get called anyway. Does that answer your question? I think the intent of the above code is to force-close the connection when there's a new connection string.

shayshim: > I'm OK with that. @shayshim ? Maybe you can both approve this PR when you're OK

I agree  - better to continue refactor in different PR, dedicated for that

shayshim: > `closeAndReset()` will cause `ensembleProvider.getConnectionString()` to get called anyway. Does that answer your question? I think the intent of the above code is to force-close the connection when there's a new connection string.
In short I wanted to make sure  that `closeAndReset()` (create `helper`) is called before any new connection string is set . 
As @cammckenzie said, it is called from `ConnectionState.start()`, which is called (eventually) from `CuratorFrameworkImpl.start()`, from line 338 using `client.start()` - before `ensembleTracker.start()` in line 353, which allows to set new connection strings. 
So any call to `getNewConnectionString()` will meet non `null` `helper`. 
In addition to that, `helper.getZooKeeper()` is also called eventually from 338, so `data.connectionString` is set as needed, before `getNewConnectionString()` can be called. 
So I think we have no issue here.

PR #346, opened 2020-02-18
[CURATOR-559] Fix nested retry loops

Randgalt:
The retry loop mechanism ended up getting nested multiple times causing exponential calls to the retry policy and violating a given policy's limits. Use a thread local to mitigate this so that a retry loop is reused for nested API calls, etc.


Randgalt:
attn: @shayshim and @cammckenzie

PR #348, opened 2020-03-07
CURATOR-562 Remove ConnectionHandlingPolicy

TisonKun:
See also the description in [JIRA ticket](https://issues.apache.org/jira/browse/CURATOR-562) and commit message of [the last commit](https://github.com/apache/curator/commit/cc89762e4d08af30e27b8f58a2a47c166c160569)


curator-client/src/main/java/org/apache/curator/ConnectionState.java
```diff
@@ -218,7 +216,7 @@ public String call()
         };
         int lastNegotiatedSessionTimeoutMs = getLastNegotiatedSessionTimeoutMs();
         int useSessionTimeoutMs = (lastNegotiatedSessionTimeoutMs > 0) ? lastNegotiatedSessionTimeoutMs : sessionTimeoutMs;
-        ConnectionHandlingPolicy.CheckTimeoutsResult result = connectionHandlingPolicy.checkTimeouts(hasNewConnectionString, connectionStartMs, useSessionTimeoutMs, connectionTimeoutMs);
+        CheckTimeoutsResult result = checkTimeouts(hasNewConnectionString, connectionStartMs, useSessionTimeoutMs, connectionTimeoutMs);
```
Randgalt: To my eye, the separate method doesn't add much and hurts readability. To the point, `RESET_CONNECTION`, `CONNECTION_TIMEOUT` and `SESSION_TIMEOUT` are no longer possible. Only `NOP` and `NEW_CONNECTION_STRING` are possible so `CheckTimeoutsResult` is no longer needed.

TisonKun: Yes. I mentioned it in the JIRA and commit message.

>Besides, some values in CheckTimeoutsResult seems out of day that we might have a follow-up to handle it.

I'd prefer do it in another cycle so that I have cycle to understand the impact by removing it :-)

TisonKun: If you think we'd better include the removal of `CheckTimeoutsResult` in this pr, we can hold it and let me try to investigate in time/

Randgalt: Yes, it would be better in this PR

TisonKun:
@Randgalt `CheckTimeoutsResult` end up with replaced with an if-else block. However, tarvis fails on Java 8 Env at

>[ERROR] Tests run: 2, Failures: 1, Errors: 0, Skipped: 0, Time elapsed: 24.828 s <<< FAILURE! - in org.apache.curator.framework.imps.TestWithCluster
[ERROR] testSessionSurvives(org.apache.curator.framework.imps.TestWithCluster)  Time elapsed: 13.484 s  <<< FAILURE!
java.lang.AssertionError: expected object to not be null
	at org.apache.curator.framework.imps.TestWithCluster.testSessionSurvives(TestWithCluster.java:77)

I cannot reproduce the problem locally, also don't have an idea the relation to this pr. Could you take a look?

Randgalt:
@TisonKun our tests are flakey unfortunately. I tried this test locally with your latest commit and it passes. Does it pass for you?

TisonKun:
@Randgalt yes, as I said "I cannot reproduce the problem locally", it passed in my environment.

Randgalt:
Sorry - I wasn't sure what you meant by that. OK. I'm not concerned. I'll continue with the review.

curator-client/src/main/java/org/apache/curator/ConnectionState.java
```diff
@@ -18,7 +18,6 @@
  */
```
Randgalt: Man - this class has become nearly incomprehensible. This really needs to be addressed. I'm thinking about it.

TisonKun: I agree and the first time I notice this `ConnectionState` with `curator-framwork`'s `ConnectionState`, it really confused me. But we can separated the refactor or anything on this `ConnectionState` in another pass IMO.

Randgalt:
@TisonKun I sent you a minor PR - other than that I think this is ready

TisonKun:
@Randgalt Thanks for your review and PR! I've merged it.

Randgalt:
merged

PR #350, opened 2020-03-15
[CURATOR-558] - Updates for ZooKeeper 3.6.0

Randgalt:
### 3.6.0 has many changes

ZooKeeper 3.6.0 has some significant changes from previous versions and this PR is only bringing Curator up to date. Some of these changes are:

- The reconfig APIs have moved into a new class, `ZooKeeperAdmin`. This class existed in 3.5.x but wasn't required. Now it is. 
- `DefaultZookeeperFactory` now always creates a `ZooKeeperAdmin`. There is a new factory, `NonAdminZookeeperFactory`, if users don't want the admin version of ZooKeeper.
- A bunch of little things changed in the ZK server code which affected Curator's test classes. I moved it all into reflection based calls in `Compatibility.java` in the test module. We now have modules that test 3.5 and 3.6 so we're safe with compatibility.
- ZooKeeper's `MultiTransactionRecord` has been removed it seems. That forced `CuratorMultiTransactionRecord` to be re-written. It's not a public class so hopefully it won't affect anyone.

### New module

There is a new module, `curator-test-zk35`. It forces ZooKeeper 3.5.7 and performs selected tests from the other modules to ensure compatibility. Tests annotated with TestNG group `zk35Compatibility` are tested. Group `zk36` is excluded. Note: these tests will only run from Maven. I don't think IntelliJ/Eclipse support the Maven syntax I used.

### 3 More PRs coming

I have 3 more PRs that will be separate from this to avoid too much review pain:

- #334
- #335 
- #336

### DIFF
--- a/curator-client/src/main/java/org/apache/curator/utils/Compatibility.java
+++ b/curator-client/src/main/java/org/apache/curator/utils/Compatibility.java
@@ -43,7 +43,7 @@ public class Compatibility
         {
             Class<?> multipleAddressesClass = Class.forName("org.apache.zookeeper.server.quorum.MultipleAddresses");
             localGetReachableOrOneMethod = multipleAddressesClass.getMethod("getReachableOrOne");
-            LoggerFactory.getLogger(Compatibility.class).info("Using org.apache.zookeeper.server.quorum.MultipleAddresses");
+            log.info("Using org.apache.zookeeper.server.quorum.MultipleAddresses");
         }
         catch ( ReflectiveOperationException ignore )
         {
@@ -59,7 +59,7 @@ public class Compatibility
         catch ( NoSuchFieldException e )
         {
             localAddrField = null;
-            LoggerFactory.getLogger(Compatibility.class).error("Could not get addr field! Reconfiguration fail!");
+            log.error("Could not get addr field! Reconfiguration fail!");
         }
         addrField = localAddrField;
     }
--- a/curator-client/src/main/java/org/apache/curator/utils/Compatibility.java
+++ b/curator-client/src/main/java/org/apache/curator/utils/Compatibility.java
@@ -1,0 +1,104 @@
+/**
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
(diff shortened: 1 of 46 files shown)


Randgalt:
@cammckenzie, @shayshim and @TisonKun - I want this PR to be against CURATOR-558 so I closed #338 and opened this.

curator-client/src/main/java/org/apache/curator/utils/Compatibility.java
```diff
@@ -0,0 +1,104 @@
...
+
+        Field localAddrField;
+        try
+        {
+            localAddrField = QuorumPeer.QuorumServer.class.getField("addr");
+        }
+        catch ( NoSuchFieldException e )
+        {
+            localAddrField = null;
+            LoggerFactory.getLogger(Compatibility.class).error("Could not get addr field! Reconfiguration fail!");
```
TisonKun:
```suggestion
            log.error("Could not get addr field! Reconfiguration fail!");
```

cherry-pick the comment, `log` has been initialized here, hasn't it?

curator-client/src/main/java/org/apache/curator/utils/Compatibility.java
```diff
@@ -0,0 +1,104 @@
...
+    private static final Logger log = LoggerFactory.getLogger(Compatibility.class);
+
+    static
+    {
+        Method localGetReachableOrOneMethod;
+        try
+        {
+            Class<?> multipleAddressesClass = Class.forName("org.apache.zookeeper.server.quorum.MultipleAddresses");
+            localGetReachableOrOneMethod = multipleAddressesClass.getMethod("getReachableOrOne");
+            LoggerFactory.getLogger(Compatibility.class).info("Using org.apache.zookeeper.server.quorum.MultipleAddresses");
```
TisonKun:
```suggestion
            log.info("Using org.apache.zookeeper.server.quorum.MultipleAddresses");
```

ditto

curator-framework/src/main/java/org/apache/curator/framework/api/CuratorEventType.java
```diff
@@ -96,5 +96,10 @@
     /**
      * Event sent when client is being closed
      */
-    CLOSING
+    CLOSING,
+
+    /**
+     * Corresponds to {@link org.apache.curator.framework.CuratorFramework#watchers()}
```
TisonKun:
```suggestion
     * Corresponds to {@link CuratorFramework#watchers()}
```

consistent

curator-test-zk35/pom.xml
```diff
@@ -0,0 +1,161 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
TisonKun: Some of our `pom.xml` have an Apache LICENSE header. I'm unclear where we should include it but a reminder.

PR #354, opened 2020-03-26
[CURATOR-565] check useContainerParentsIfAvailable for ExistsBuilderImpl.

PhantomThief:


Randgalt:
Please provide a test case and/or more information on why this is needed.

curator-framework/src/main/java/org/apache/curator/framework/imps/ExistsBuilderImpl.java
```diff
@@ -67,7 +67,9 @@ public ACLableExistBuilderMain creatingParentsIfNeeded()
     @Override
     public ACLableExistBuilderMain creatingParentContainersIfNeeded()
     {
-        createParentContainersIfNeeded = true;
+        if (client.useContainerParentsIfAvailable()) {
```
Randgalt, line 70: this should be on the `creatingParentsIfNeeded()` method

eolivelli:
@PhantomThief do you have cycles to follow up on this patch ?

tisonkun:
@PhantomThief this patch may make sense as we use different logic handle these flags between `CreateBuilder` and `ExistBuilder`. However, it may not be resolved in this way but align to how `CreateBuilderImpl` handles `createParentsIfNeeded` and `createParentsAsContainers`.

tisonkun:
Closed as stale. Feel free to resubmit it if you're still willing to proceed.

PR #357, opened 2020-04-01
[CURATOR-525] Fix for Zombie LOST Curator

Randgalt:
There is a race whereby the ZooKeeper connection can be healed before Curator is finished processing the new connection state. When this happens the Curator instance becomes a Zombie stuck in the LOST state. This fix is a "hack". ConnectionStateManager will notice that the connection state is LOST but that the Curator instance reports that it is connected. When this happens, it is logged and the connection is reset.

### DIFF
--- a/curator-framework/src/main/java/org/apache/curator/framework/state/ConnectionStateManager.java
+++ b/curator-framework/src/main/java/org/apache/curator/framework/state/ConnectionStateManager.java
@@ -291,16 +291,9 @@ public class ConnectionStateManager implements Closeable
                     if ( (currentConnectionState == ConnectionState.LOST) && client.getZookeeperClient().isConnected() )
                     {
                         // CURATOR-525 - there is a race whereby LOST is sometimes set after the connection has been repaired
-                        // this "hack" fixes it by resetting the connection
-                        log.warn("ConnectionState is LOST but isConnected() is true. Resetting connection.");
-                        try
-                        {
-                            client.getZookeeperClient().reset();
-                        }
-                        catch ( Exception e )
-                        {
-                            log.error("Could not reset connection after LOST/isConnected mismatch");
-                        }
+                        // this "hack" fixes it by forcing the state to RECONNECTED
+                        log.warn("ConnectionState is LOST but isConnected() is true. Forcing RECONNECTED.");
+                        addStateChange(ConnectionState.RECONNECTED);
                     }
                 }
             }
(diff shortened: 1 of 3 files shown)


Randgalt:
attn: @jschlather and @chuchao333

curator-framework/src/main/java/org/apache/curator/framework/state/ConnectionStateManager.java
```diff
@@ -319,8 +319,18 @@ else if ( currentConnectionState == ConnectionState.LOST )
         {
             try
             {
-                // give ConnectionState.checkTimeouts() a chance to run, reset ensemble providers, etc.
-                client.getZookeeperClient().getZooKeeper();
+                if ( client.getZookeeperClient().isConnected() )
```
chuchao333, line 322: right now, the `checkSessionExperiation` is only called `if sessionExpirationPercent > 0`, we should also change that?

Randgalt: Maybe - I feel like this is such a hack I wanted to limit it. But, maybe you're right.

Randgalt: Check now @chuchao333

chuchao333: LGTM now, thanks for the fix.

Randgalt: TBH - this makes me nervous. I worry about an infinite loop of resets. I'll think about this some more before I merge.

chuchao333: @Randgalt yep, I agree. I would also prefer if we could fix this without firing the unexpected LOST in the first place.

Also TBH, some of the curator code sounds "no obvious defect" instead of "obviously no defect" to me :) I do understand it's complex enough given all the edge cases it has to handle.

Randgalt: Much of that backend code needs to be reconsidered. If you're interested it would be a great project. I could help. If, I find the time, I'll rework some of it. For example, the ConnectionState code belongs in the client code, not the framework code. Pulling all that out and re-writing it directly in the client module would simplify things and give us an opportunity to remove some races due to their being two different threads (ZK's and Curator's).

Moonergfp:
https://issues.apache.org/jira/browse/CURATOR-561
I think that this fixed code may ultimately  correct the state of Curator.

PR #359, opened 2020-04-19
[CURATOR-559] testThreadedRecursingRetry

Randgalt:
test used InterProcessReadWriteLock - the problem is that the retry can end up in background loop which has its own thread thereby spoiling the test. Instead use a foreground operation for consistent tests


cammckenzie:
LGTM

PR #363, opened 2020-05-04
CURATOR-544: SessionFailedRetryPolicy

TisonKun:
More details on [JIRA](https://issues.apache.org/jira/browse/CURATOR-544#).

The breaking parts are we remove `RetryLoop.shouldRetry` & `RetryLoop. isRetryException` but it is no harm to add them back if we want to keep public APIs.

cc @Randgalt @shayshim

### DIFF
--- a/curator-client/src/main/java/org/apache/curator/RetryPolicy.java
+++ b/curator-client/src/main/java/org/apache/curator/RetryPolicy.java
@@ -52,8 +52,7 @@ public interface RetryPolicy
             return (rc == KeeperException.Code.CONNECTIONLOSS.intValue()) ||
                     (rc == KeeperException.Code.OPERATIONTIMEOUT.intValue()) ||
                     (rc == KeeperException.Code.SESSIONMOVED.intValue()) ||
-                    (rc == KeeperException.Code.SESSIONEXPIRED.intValue()) ||
-                    (rc == -13); // KeeperException.Code.NEWCONFIGNOQUORUM.intValue()) - using hard coded value for ZK 3.4.x compatibility
+                    (rc == KeeperException.Code.SESSIONEXPIRED.intValue());
         }
         return false;
     }
--- a/curator-client/src/main/java/org/apache/curator/RetryPolicy.java
+++ b/curator-client/src/main/java/org/apache/curator/RetryPolicy.java
@@ -18,6 +18,8 @@
  */
 package org.apache.curator;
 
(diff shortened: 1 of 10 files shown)


Randgalt:
Please add details on the breaking changes to https://github.com/apache/curator/blob/master/src/site/confluence/breaking-changes.confluence

curator-framework/src/main/java/org/apache/curator/framework/imps/CuratorFrameworkImpl.java
```diff
@@ -637,7 +637,8 @@ boolean useContainerParentsIfAvailable()
         boolean doQueueOperation = false;
         do
         {
-            if ( RetryLoop.shouldRetry(event.getResultCode()) )
+            final KeeperException ke = KeeperException.create(event.getResultCode());
```
Randgalt, line 640: nit - for me creating `ke` is unneeded. Create the exception directly as the argument

curator-client/src/main/java/org/apache/curator/RetryPolicy.java
```diff
@@ -33,5 +35,26 @@
...
+    default boolean allowRetry(Throwable exception)
+    {
+        if ( exception instanceof KeeperException)
+        {
+            final int rc = ((KeeperException) exception).code().intValue();
+            return (rc == KeeperException.Code.CONNECTIONLOSS.intValue()) ||
+                    (rc == KeeperException.Code.OPERATIONTIMEOUT.intValue()) ||
+                    (rc == KeeperException.Code.SESSIONMOVED.intValue()) ||
+                    (rc == KeeperException.Code.SESSIONEXPIRED.intValue()) ||
+                    (rc == -13); // KeeperException.Code.NEWCONFIGNOQUORUM.intValue()) - using hard coded value for ZK 3.4.x compatibility
```
cammckenzie, line 56: @Randgalt Can we move this to a constant now that we're no longer supporting 3.4?

Randgalt: Yes for sure

TisonKun: Thanks for spotting this. Will update.

TisonKun:
Thanks for your reviews! Will update in hours.

TisonKun:
Updated

TisonKun:
Thanks!

PR #364, opened 2020-05-06
CURATOR-568-option-CF-skip-ensemble-tracking

chevaris:
CURATOR-568
Adding ensembleTracker(boolean) and withEnsembleTracker() methods to
CuratorFrameworkFactory.builder() that allows enabling/disabling
ensemble tracking


chevaris:
Hi there,
I would appreciate your comments on this commit.
CI is failing when running with JDK 8 (passed OK with JDK 11)(org.apache.curator.framework.recipes.locks.TestInterProcessMutex.testReentrantSingleLock), but I do not think it is related with the commit itself (The test is consistently running OK in my laptop with both Java 8 and 11, and I do not see how this commit could influence that test). I would appreciate if someone can take a look

Thanks in advance

PR #365, opened 2020-05-08
Replace OutstandingOps with JDK bundled Phaser

TisonKun:
It might or might not be an improvement. But let's verify it with test suite first.


tisonkun:
Originally discussed at https://github.com/apache/curator/pull/335#discussion_r399591042

eolivelli:
Committed to master branch

Please update JIRA

PR #366, opened 2020-05-10
[CURATOR-569] API TO HANDLE PROTECTED MODE ZNODE NAMES

chevaris:
I would appreciate your comments on this commit to solve CURATOR-569

### DIFF
--- a/curator-framework/src/main/java/org/apache/curator/framework/api/CreateBuilderMain.java
+++ b/curator-framework/src/main/java/org/apache/curator/framework/api/CreateBuilderMain.java
@@ -20,6 +20,7 @@ package org.apache.curator.framework.api;
 
 import java.util.UUID;
 
+import org.apache.curator.framework.imps.ProtectedUtils;
 import org.apache.zookeeper.CreateMode;
 
 public interface CreateBuilderMain extends
@@ -29,55 +30,6 @@ public interface CreateBuilderMain extends
     Compressible<CreateBackgroundModeStatACLable>,
     Statable<CreateProtectACLCreateModePathAndBytesable<String>>
 {
-
-    /**
-     * First 3 characters in the prefix on ZNode names when using {@link #withProtection()}
-     */
-    static final String PROTECTED_PREFIX = "_c_";
-
-    /**
-     * Last character used in the prefix on ZNode names when using {@link #withProtection()}
-     */
-    static final char PROTECTED_SEPARATOR = '-';
-
-    /**
-     * Prefix length on ZNode name when using {@link #withProtection()}
-     */
-    static final int PROTECTED_PREFIX_WITH_UUID_LENGTH = 
-            PROTECTED_PREFIX.length() 
-            + 36 // UUID canonical text representation produced by {@link UUID#toString()}
-            + 1; // Separator length
-
-    /**
-     * Utility method to determine if a given ZNode name starts with Curator's generated protected prefix.
-     * 
-     * @param znodeName ZNode name
-     * @return {@code true} if the given ZNode name starts with Curator's generated protected prefix
-     */
-    public static boolean isProtectedZNode(final String znodeName) {
-        if (znodeName.length() > PROTECTED_PREFIX_WITH_UUID_LENGTH
-                && znodeName.startsWith(PROTECTED_PREFIX)
-                && znodeName.charAt(PROTECTED_PREFIX_WITH_UUID_LENGTH-1) == PROTECTED_SEPARATOR
-           ) {
-            try {
-                UUID.fromString(znodeName.substring(PROTECTED_PREFIX.length(), PROTECTED_PREFIX_WITH_UUID_LENGTH-2));
-                return true;
-            } catch (IllegalArgumentException e) {
-                // Just false
-            }
-        }
-        return false;
-    }
-
-    /**
-     * Utility method to remove Curator's generated protected prefix if present
-     * 
-     * @param znodeName ZNode name
-     * @return string without Curator's generated protected prefix if present; original string if prefix not present
-     */
-    public static String removeProtectedPrefix(final String znodeName) {
-        return isProtectedZNode(znodeName) ? znodeName.substring(PROTECTED_PREFIX_WITH_UUID_LENGTH) : znodeName;
-    }
 
     /**
      * Causes any parent nodes to get created if they haven't already been
@@ -128,9 +80,9 @@ public interface CreateBuilderMain extends
      *     Putting the create builder into protection mode works around this.
      *     The name of the node that is created is prefixed with a 40 characters string that is the concatenation of
      *     <ul>
-     *         <li>{@value #PROTECTED_PREFIX}
+     *         <li>{@value ProtectedUtils#PROTECTED_PREFIX}
      *         <li>Canonical text representation of a random generated UUID as produced by {@link UUID#toString()}
-     *         <li>{@value #PROTECTED_SEPARATOR}
+     *         <li>{@value ProtectedUtils#PROTECTED_SEPARATOR}
      *     </ul>
      *     If node creation fails the normal retry mechanism will occur. On the retry, the parent path is first searched
      *     for a node that has previous described prefix in it. If that node is found, it is assumed to be the lost
--- a/curator-framework/src/main/java/org/apache/curator/framework/api/CreateBuilderMain.java
+++ b/curator-framework/src/main/java/org/apache/curator/framework/api/CreateBuilderMain.java
@@ -18,6 +18,8 @@
  */
 package org.apache.curator.framework.api;
 
+import java.util.UUID;
+
 import org.apache.zookeeper.CreateMode;
 
 public interface CreateBuilderMain extends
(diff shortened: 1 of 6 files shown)


Randgalt:
> Looks OK to me. I wonder if this should be moved to some sort of util class though rather than packaging it within the CreateBuilder.

I was thinking the same thing. `CreateBuilder` is hidden and not normally used.

curator-framework/src/main/java/org/apache/curator/framework/imps/CreateBuilderImpl.java
```diff
@@ -751,7 +748,7 @@ public String forPath(String path) throws Exception {
 
     private static String getProtectedPrefix(String protectedId)
```
Randgalt, line 749: This method should be public as well. Also, move the whole bit (all 4 APIs into a helper util class in the root).

curator-framework/src/main/java/org/apache/curator/framework/api/CreateBuilderMain.java
```diff
@@ -27,6 +29,56 @@
...
+    public static boolean isProtectedZNode(final String znodeName) {
+        if (znodeName.length() > PROTECTED_PREFIX_WITH_UUID_LENGTH
+                && znodeName.startsWith(PROTECTED_PREFIX)
+                && znodeName.charAt(PROTECTED_PREFIX_WITH_UUID_LENGTH-1) == PROTECTED_SEPARATOR
+           ) {
+            try {
+                UUID.fromString(znodeName.substring(PROTECTED_PREFIX.length(), PROTECTED_PREFIX_WITH_UUID_LENGTH-2));
+                return true;
+            } catch (IllegalArgumentException e) {
+                // Just false
```
TisonKun, line 66: nit - (personal taste) explicitly return `false` does no harm and more expressive

curator-framework/src/main/java/org/apache/curator/framework/api/CreateBuilderMain.java
```diff
@@ -27,6 +29,56 @@
...
+        return false;
+    }
+
+    /**
+     * Utility method to remove Curator's generated protected prefix if present
+     * 
+     * @param znodeName ZNode name
+     * @return string without Curator's generated protected prefix if present; original string if prefix not present
+     */
+    public static String removeProtectedPrefix(final String znodeName) {
```
TisonKun, line 78: What if returns `Optional<String>` which equals `isProtectedZNode(znodeName) ? Optional.of(znodeName.substring(PROTECTED_PREFIX_WITH_UUID_LENGTH)) : Optional.empty()`.

For API semantic it fits in "if present" and user can easily get the original sting by `.orElse` but hard to filter non-protected znodes with this API(the user of this method seems interested in protected znodes only.)

chevaris: The purpose it is just the opposite and not to worry if the ZNode is protected or not when using the API. Why is hard to filter non-protected znodes? You pass the ZNode name and if the ZNode is not protected you get the same.
Basically, this method provides you a way to normalize the names and remove the non-semantic part of it (does not matter if the node is protected or not)
I do not think that Optional is a good fit.  The method provides a mean  to get the normalized path (without prefix) given a real path, so it can never be null. All valid ZNodes names will have a normalized form without the prefix.

chevaris:
Thanks all for your comments.

I see consesus on the comments to move the new static methods to a new helper class (e.g. ProtectedUtils), and some of previous methods present in CreateBuilderImpl. I will do that!!!

TisonKun proposed an alternative for the API; I disagree with the proposal but I would appreciate if maybe Jordan or cammckenzie  could comment on that.

curator-framework/src/main/java/org/apache/curator/framework/api/CreateBuilderMain.java
TisonKun: Thanks for your clarification. It makes sense we add a method for normalizing which works as you described. What about name it `normalize` and document the behavior as you said.

Normalizing is the purpose and remove prefix if it is a protected node is the implementation.

Randgalt: Yeah - let's leave it as it.

chevaris:
Provided new code addressing previous comments

curator-framework/src/main/java/org/apache/curator/framework/imps/ProtectedUtils.java
```diff
@@ -0,0 +1,95 @@
...
+            + 1; // Separator length
+
+    /**
+     * Provides a prefix to be appended to a ZNode name when protected. The method assumes that the provided string
+     * adjusts to the required format
+     * 
+     * @param protectedId canonical text representation of a UUID
+     * @return string that concatenates {@value #PROTECTED_PREFIX}, the given id and {@value #PROTECTED_SEPARATOR}
+     */
+    static String getProtectedPrefix(final String protectedId)
```
Randgalt, line 60: should be `public`

curator-framework/src/main/java/org/apache/curator/framework/imps/CreateBuilderImpl.java
```diff
@@ -1224,7 +1216,7 @@ String adjustPath(String path) throws Exception
         if ( protectedMode.doProtected() )
         {
             ZKPaths.PathAndNode pathAndNode = ZKPaths.getPathAndNode(path);
-            String name = getProtectedPrefix(protectedMode.protectedId()) + pathAndNode.getNode();
+            String name = ProtectedUtils.getProtectedPrefix(protectedMode.protectedId()) + pathAndNode.getNode();
```
Randgalt, line 1219: Let's move this into `ProtectedUtils`. Something like `ProtectedUtils.toProtectedZNode(String znodeName)`

curator-framework/src/main/java/org/apache/curator/framework/imps/ProtectedUtils.java
```diff
@@ -0,0 +1,95 @@
...
+public final class ProtectedUtils {
+
+    private ProtectedUtils() {
+        throw new RuntimeException("Protected Utils is a helper class");
+    }
+
+    /**
+     * First 3 characters in the prefix on ZNode names when using {@link CreateBuilderMain#withProtection()}
+     */
+    public static final String PROTECTED_PREFIX = "_c_";
```
Randgalt, line 38: Let's make these constants private and, if need, add additional public methods to do whatever tasks might be needed.

curator-framework/src/main/java/org/apache/curator/framework/imps/ProtectedUtils.java
chevaris: Any particular reason?
This method is only interesting for implementations of the CreatorBuilderMain...

Actually I think it is a better approach to put as a default method of CreatorBuilderMain

curator-framework/src/main/java/org/apache/curator/framework/imps/ProtectedUtils.java
chevaris: Fine

curator-framework/src/main/java/org/apache/curator/framework/imps/CreateBuilderImpl.java
chevaris: Fine for me, I will move it there.
Public also? My opinion is same as before (protected), even better as default method of the interface

curator-framework/src/main/java/org/apache/curator/framework/imps/ProtectedUtils.java
Randgalt: > Any particular reason?

I get frustrated with other libraries when I want to use something (for debugging, etc.) and it's not public. Any reason not to make this public?

chevaris: Thx for the explanation. Let me explain my reasoning.

When building libraries, I like to think in this way:
End user interest -> public APIs
Library developers Interest -> reduce visibility as much as possible (protected/private APIs)

This method is only interesting for interface implementations, because end users will limit to use withProtection() method. Other methods (normalize...) are interesting for end users to potentially remove non semantic part of the name when watching a ZNode.
I think that there is no way (maybe I am wrong) for the end user to provide an implementation of the CreateBuilderMain interface.

I will make the methods public, unless I have convinced you :) and you say the opposite

Randgalt: I think you're making a huge assumption there. We really don't know how clients will use the library - I've been very surprised over the years. Further, I've had to go to great lengths to get around stuff that's hidden in ZooKeeper. Remember, this is Java, nothing is truly hidden. You can access anything via reflection. 

tbh - I'd also like to see a method: `public static String extractProtectId(String znodeName)`. Let's give people as many tools as possible.

chevaris:
Thanks for your previous comments.
New version of the code that should refelect your previous feedback

curator-framework/src/main/java/org/apache/curator/framework/imps/ProtectedUtils.java
```diff
@@ -0,0 +1,157 @@
...
+     * Prefix length on ZNode name when using {@link #withProtection()}
+     */
+    @VisibleForTesting
+    static final int PROTECTED_PREFIX_WITH_UUID_LENGTH = 
+            PROTECTED_PREFIX.length() 
+            + 36 // UUID canonical text representation produced by {@link UUID#toString()}
+            + 1; // Separator length
+
+    /**
+     * Provides a prefix to be appended to a ZNode name when protected. The method assumes that the provided string
```
Randgalt, line 61: nit - should be "prepended". I'll fix it when I merge.

Randgalt:
merged

PR #367, opened 2020-06-15
CURATOR-574 DiscoveryService fatal error on deserializing an empty byte[] as JSON

eolivelli:
see https://issues.apache.org/jira/browse/CURATOR-574

### DIFF
--- a/curator-x-discovery/src/main/java/org/apache/curator/x/discovery/details/ServiceCacheImpl.java
+++ b/curator-x-discovery/src/main/java/org/apache/curator/x/discovery/details/ServiceCacheImpl.java
@@ -84,7 +84,7 @@ public class ServiceCacheImpl<T> implements ServiceCache<T>, PathChildrenCacheLi
             .withDataNotCached()
             .build();
         CuratorCacheListener listener = CuratorCacheListener.builder()
-            .forPathChildrenCache(discovery.getClient(), this)
+            .forPathChildrenCache(path, discovery.getClient(), this)
             .forInitialized(this::initialized)
             .build();
         cache.listenable().addListener(listener);
--- a/curator-recipes/src/main/java/org/apache/curator/framework/recipes/cache/CuratorCacheListenerBuilder.java
+++ b/curator-recipes/src/main/java/org/apache/curator/framework/recipes/cache/CuratorCacheListenerBuilder.java
@@ -86,11 +86,12 @@ public interface CuratorCacheListenerBuilder
      * behaves differently than {@link org.apache.curator.framework.recipes.cache.PathChildrenCache} so
      * things such as event ordering will likely be different.
      *
+     * @param rootPath the root path. The listener needs this information in order to bridge only events for children of this path and not the path itself
      * @param client the curator client
      * @param listener the listener to wrap
      * @return a CuratorCacheListener that forwards to the given listener
      */
-    CuratorCacheListenerBuilder forPathChildrenCache(CuratorFramework client, PathChildrenCacheListener listener);
+    CuratorCacheListenerBuilder forPathChildrenCache(String rootPath, CuratorFramework client, PathChildrenCacheListener listener);
 
     /**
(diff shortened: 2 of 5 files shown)


curator-x-discovery/src/main/java/org/apache/curator/x/discovery/details/ServiceCacheImpl.java
```diff
@@ -78,7 +79,7 @@ private static ExecutorService convertThreadFactory(ThreadFactory threadFactory)
 
         this.discovery = discovery;
 
-        String path = discovery.pathForName(name);
+        path = discovery.pathForName(name);
```
Randgalt, line 82: The bug might actually be in `forPathChildrenCache()` of `CuratorCacheListener`. We shouldn't have to filter the root out. I believe PathChildrenCache didn't report on the root so `forPathChildrenCache` should emulate that behavior.

eolivelli: @Randgalt do you think this is a regression ?

I will push down the fix to CuratorCacheListener, thanks for the suggestion.
I didn't go that way because I wasn't sure.

It is important to understand if this is a regression in 5.0 and then we have to fix it asap, otherwise applications will break

Randgalt: If pushing it down fixes it then it is a regression. I'm surprised we don't have a test for this. We can get a new version out if needed.

eolivelli: I see this problem with ZK 3.5.7 not with 3.6.

btw @Randgalt  I see that CuratorCacheListener does not hold a reference to the root path, and it does not hold any reference to the CuratorCache.
I cannot introduce such reference, because it will be an incompatible change.

itarakanov: We can see the problem even in ZK 3.6.1

Randgalt: > I see this problem with ZK 3.5.7 not with 3.6.

Yeah - I'm mistaken. I thought it use PathChildrenCache in pre-3.6 but it uses TreeCache which will have the same problem.

Randgalt: @eolivelli I think it could be handled by adding a new `CuratorCache.Options` - something like `IGNORE_ROOT` and then `CuratorCacheBridgeBuilder` would always set that option.

curator-x-discovery/src/main/java/org/apache/curator/x/discovery/details/ServiceCacheImpl.java
```diff
@@ -163,9 +164,19 @@ public void removeListener(ServiceCacheListener listener)
         discovery.getClient().getConnectionStateListenable().removeListener(listener);
     }
 
+    private boolean isRootPathEvent(PathChildrenCacheEvent event) {
```
itarakanov, line 167: I would rather call it something like **isEmptyPathEvent**, because "root" is a relative thing.

eolivelli:
@Randgalt 
the patch is finally ready for review.
I have added a few assertions on an existing test.

With this fix the errors on CURATOR-574 are not present anymore.
I wonder if it is better to rename the issue, actually this was a problem on PathChildrenCacheListenerWrapper.java and not in ServiceDiscovery.

The change is not backward compatibile, but it is on a fresh new API of Curator 5.0, so if you release 5.0.1 (or 5.1 if you want to strictly observe SemVer) it won't be a big deal

I can create the 5.0.1 release

Randgalt:
> I can create the 5.0.1 release

Please do

eolivelli:
@Randgalt I used the GitHub button and this resulted with a merge commit on git history.
I apologize.

PR #370, opened 2020-07-09
CURATOR-577 - Add a dedicated downloads page to the website

Randgalt:
Add a dedicated downloads page to the website.

The new version of the website can be seen here: http://curator.apache.org/staging/

I tried using the Build Helper plugin to get the current release version but it picked an old version on my machine and left me not trusting it. So, we now have a property specifically for the download page in the root POM: `currentStableVersion`. Once this new website is released I'll update the instructions on the wiki page for building the website.

### DIFF
--- a/src/site/site.xml
+++ b/src/site/site.xml
@@ -71,7 +71,7 @@
 
         <menu name="Apache Curator" inherit="top">
             <item name="About" href="index.html"/>
-            <item name="Download" href="https://cwiki.apache.org/confluence/display/CURATOR/Releases"/>
+            <item name="Download" href="releases.html"/>
             <item name="Getting Started" href="getting-started.html"/>
             <item name="Examples" href="curator-examples/index.html"/>
             <item name="Recipes" href="curator-recipes/index.html"/>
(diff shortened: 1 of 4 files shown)


src/site/site.xml
```diff
@@ -71,7 +71,7 @@
 
         <menu name="Apache Curator" inherit="top">
             <item name="About" href="index.html"/>
-            <item name="Download" href="https://cwiki.apache.org/confluence/display/CURATOR/Releases"/>
+            <item name="Download" href="releases.html"/>
```
TisonKun, line 74: Shall we name this tag as "Releases"?

eolivelli: Every website has a "Download" page 
:-)
In my opinion it is better to name it "Download"

TisonKun:
It's ok to name it as "Download" since people might be more interested in downloading. I get a bit confused because it also contains releases infos. Given other Apache projects use this pattern it should be fine :-)

eolivelli:
@TisonKun would you mind to "Approve" formally this PR ? Your review is still in status "left review comments"

eolivelli:
@TisonKun probably the "approve" button disappeared, sorry

TisonKun:
@eolivelli never mind. Will approve formally when discussion resolved the next time.

PR #371, opened 2020-08-14
Add tison as Committer and PMC

TisonKun:
cc @Randgalt @eolivelli I can see some of our commits merge without merge commits while others with. Thus I'm not sure what it the method I should put this commit onto master branch. Open a pull request for your advice.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -323,6 +323,18 @@
             </roles>
             <url>https://people.apache.org/~eolivelli</url>
         </developer>
+
+        <developer>
+            <id>tison</id>
+            <name>Zili Chen</name>
+            <email>tison@apache.org</email>
+            <timezone>+8</timezone>
+            <roles>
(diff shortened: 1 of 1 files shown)


TisonKun:
BTW, I've build the website locally and verify the change manually. From the guide it requires a version bump but it is necessary for website rebuild?

eolivelli:
@TisonKun  I have merged this patch using gitHub button.
We do not have a "standard" rule about commits, it is better IMO to squash the commits into one single commit in order to have a good history.

Regarding the website: 
I got into the same problem while adding myself.
Probably the best thing is to push the website with the current -SNAPSHOT version.
IIRC I have cherry picked the commit to a temporary local branch started from the latest released tag and then I rebuild the website,
then I dropped the temporary branch

TisonKun:
@eolivelli thanks for your advice! I think I've made the change.

PR #373, opened 2020-08-21
CURATOR-581. fix OSGi export pattern

TisonKun:
According to https://access.redhat.com/documentation/en-us/red_hat_fuse/7.2/html/deploying_into_apache_karaf/buildbundle this is the correct pattern for wildcard in export package config. Also it fixes the problem that IDEA complains with `The package 'org.apache.curator.framework.api' is not exported by the bundle dependencies`

cc @Randgalt @eolivelli @cammckenzie


TisonKun:
Thanks for your reviews!

PR #374, opened 2020-09-19
CURATOR-583: Fix ArrayIndexOutOfBoundsException when passing empty li…

hjyun328:
Whenever I add zookeeper servers by using reconfig API of the curator client, it always throw ArrayIndexOutOfBoundsException.

If there are no servers to add or remove when using reconfigure API of ZooKeeperAdmin, I think it must pass not empty list parameter but null to reconfigure API. Because ZooKeeperAdmin tries to join strings by accessing first index of the server list.

Please refer to the link below. 

https://github.com/apache/zookeeper/blob/release-3.6.2/zookeeper-server/src/main/java/org/apache/zookeeper/admin/ZooKeeperAdmin.java#L267-L269

https://github.com/apache/zookeeper/blob/release-3.6.2/zookeeper-server/src/main/java/org/apache/zookeeper/common/StringUtils.java#L57

### DIFF
--- a/curator-framework/src/main/java/org/apache/curator/framework/imps/ReconfigBuilderImpl.java
+++ b/curator-framework/src/main/java/org/apache/curator/framework/imps/ReconfigBuilderImpl.java
@@ -130,7 +130,7 @@ public class ReconfigBuilderImpl implements ReconfigBuilder, BackgroundOperation
     @Override
     public StatConfigureEnsembleable withNewMembers(List<String> servers)
     {
-        newMembers = (servers != null) ? ImmutableList.copyOf(servers) : ImmutableList.<String>of();
+        newMembers = (servers != null && !servers.isEmpty()) ? ImmutableList.copyOf(servers) : null;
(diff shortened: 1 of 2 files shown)


eolivelli (changes requested):
Thanks for your patch.

Can you please add unit tests that cover your change in order to prevent future regressions

eolivelli (approved):
Looks good to me.

Waiting for another reviewer to check this change before committing it

TisonKun (changes requested):
Changes make sense. Possibly it is also something ZK should handle properly.

For this change, we can add one more test for `withNewMembers` which is omitted in testAddAndRemoveWithEmptyList.

[0001-path.txt](https://github.com/apache/curator/files/5251113/0001-path.txt)

hjyun328:
@eolivelli, @TisonKun
Thanks for your review.
I just added a test case for `withNewMembers`.

TisonKun (approved):
LGTM. Merging...

PR #375, opened 2020-12-12
CURATOR-585: fix DiscoveryExample did not check for exceptions

iwangjie:


curator-examples/src/main/java/discovery/DiscoveryExample.java
```diff
@@ -192,8 +194,13 @@ private static void listInstances(ServiceDiscovery<InstanceDetails> serviceDisco
                 }
             }
         }
-        finally
```
eolivelli, line 195: Please keep the closing of resources into the finally block

eolivelli (changes requested):

curator-examples/src/main/java/discovery/DiscoveryExample.java
iwangjie: OK, thank you. I'll put the resource shutdown in the finally block.

curator-examples/src/main/java/discovery/DiscoveryExample.java
```diff
@@ -192,6 +194,14 @@ private static void listInstances(ServiceDiscovery<InstanceDetails> serviceDisco
                 }
             }
         }
+        catch ( Exception e )
+        {
+            if ( e instanceof KeeperException.NoNodeException )
```
eolivelli, line 199: which instruction can throw this exception ? is it possible that it is throws inside the inner loop ? 
in that case we could end the loop prematurely .
that do you think ?

iwangjie: When there is no one service in the registration call list will directly exit the process, and in fact should be prompted without any services, unless it is other exceptions, do you think?

iwangjie: Maybe, if it's not a NoNodeException, I should throw it to stop the execution of the program, right?

eolivelli: let's make it simple.
if you found this problem with NoNode then let's fix only it.

if you have ever saw other errors, then it is worth to do more

what do you think ?

curator-examples/src/main/java/discovery/DiscoveryExample.java
```diff
@@ -192,6 +194,14 @@ private static void listInstances(ServiceDiscovery<InstanceDetails> serviceDisco
...
+            {
+                System.err.println("There are no registered instances.");
+                return;
+            }
```
eolivelli, line 203: any other exception should be rethrown

iwangjie: Yes, that's what I thought, and I hope what I'm doing now is the same as what we discussed.

curator-examples/src/main/java/discovery/DiscoveryExample.java
iwangjie: OK, it has been modified

iwangjie:
Do I have anything to deal with?

eolivelli (approved):
sorry for late reply.
I am merging this patch now

eolivelli:
committed to master branch
thanks

PR #377, opened 2021-01-17
CURATOR-586: Check if namespace is not null

nryanov:
https://issues.apache.org/jira/browse/CURATOR-586

### DIFF
--- a/curator-framework/src/main/java/org/apache/curator/framework/imps/NamespaceFacade.java
+++ b/curator-framework/src/main/java/org/apache/curator/framework/imps/NamespaceFacade.java
@@ -57,7 +57,8 @@ class NamespaceFacade extends CuratorFrameworkImpl
     @Override
     public String getNamespace()
     {
-        return namespace.getNamespace();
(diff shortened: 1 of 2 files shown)


eolivelli (changes requested):
thanks for your fix.
can you please add a minimal test case in order to prevent regressions in the future ?

nryanov:
Done 👍

Randgalt:
I recall that every time we've touched this in the past there were problems. I'd feel more comfortable if some users tested this with their code.

TisonKun:
> I recall that every time we've touched this in the past there were problems. I'd feel more comfortable if some users tested this with their code.

Is there an exact example? What is the problems' kind?

tisonkun:
@nryanov as @cammckenzie mentioned in the JIRA ticket, Curator handles nullity correctly internally. You may provide a real-world use case to make this change. Otherwise, it can be closed as unnecessary.

tisonkun:
Closed as inactive.

PR #379, opened 2021-02-18
CURATOR-587 zookeeper embedded - preview

eolivelli:
This is a preview patch for when we will have ZK 3.7.0 and the official ZooKeeperServerEmbedded API.

At that time we won't need to hack the internals of ZK

### DIFF
--- a/curator-client/src/test/java/org/apache/curator/TestRetryLoop.java
+++ b/curator-client/src/test/java/org/apache/curator/TestRetryLoop.java
@@ -108,7 +108,7 @@ public class TestRetryLoop extends BaseClassForTests
                 }
             }
 
-            assertTrue(loopCount >= 2);
+            assertTrue(loopCount >= 2, "Unexpected loop count " + loopCount);
         }
         finally
         {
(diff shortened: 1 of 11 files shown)


eolivelli:
my IDE reformatted part of the code, I will revert

pom.xml
```diff
@@ -69,7 +69,7 @@
         <redirectTestOutputToFile>true</redirectTestOutputToFile>
 
         <!-- versions -->
-        <zookeeper-version>3.6.3</zookeeper-version>
+        <zookeeper-version>3.7.1-SNAPSHOT</zookeeper-version>
```
eolivelli, line 72: we have to move this to 3.7.1

in 3.7.1 we have fixed a few problems in ZK Embedded, and 3.8.0 is also better

@tisonkun

tisonkun: Since @kezhuw resolves CURATOR-596 by #426 I'd cc you here if you're interested in taking over this patch also.

eolivelli: I have rebased on current master

PR #384, opened 2021-04-24
CURATOR-592 Initial commit, no enough tests

MGelbana-Incorta:
Please give your feedback if possible and note that `InterProcessMultiLock` is not ready yet.

### DIFF
--- a/curator-recipes/src/main/java/org/apache/curator/framework/recipes/locks/InterProcessLock.java
+++ b/curator-recipes/src/main/java/org/apache/curator/framework/recipes/locks/InterProcessLock.java
@@ -34,6 +34,14 @@ public interface InterProcessLock
     public void acquire() throws Exception;
 
     /**
+     * Same as {@link #acquire()}, but also invokes the input runnable to confirm the creation of the lock's
+     * ephemeral node.
+     *
+     * @throws Exception ZK errors, connection interruptions
+     */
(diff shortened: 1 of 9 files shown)


Randgalt:
While I understand why this might be useful for you I'm concerned about adding such a bespoke feature to Curator. The lock code is already complicated and introducing this shim makes it much harder to reason about. We can't predict what this proc will do in the context of the lock mechanics. I'm sorry but I'm -1 on this change.

MGelbana-Incorta:
It's alright!

At least now I know I'll have to go with plan B (i.e. fork the project and maintain this change myself).

Thanks a lot for your time 👍🏾

eolivelli:
@MGelbana-Incorta 

I found the description of the purpose of this change in https://issues.apache.org/jira/browse/CURATOR-592
probably for the future it is better to make a brief summary in the description of the PR and set the Title accordingly.

Apart from that....
I see that you say that this change is still not ready to be useful.

As @Randgalt comments this code is much complex and it is hard to add this feature in a way that we can provide a consistent behaviour (at least without introducing many changes)

Did you consider to create new classes ? not a "fork" maintained by yourself alone but new enhanced classes to be added here next to the existing recipes.

MGelbana-Incorta:
Creating a completely new class is a great idea!

I'll close this PR for now. Thanks a lot @eolivelli.

PR #386, opened 2021-06-28
CURATOR-588  Upgrade ZooKeeper to 3.6.3

eolivelli:
It looks like upgrading to ZK 3.6.3 is not that easy as for 3.6.2

### DIFF
--- a/curator-framework/src/test/java/org/apache/curator/framework/imps/TestFrameworkEdges.java
+++ b/curator-framework/src/test/java/org/apache/curator/framework/imps/TestFrameworkEdges.java
@@ -670,7 +670,7 @@ public class TestFrameworkEdges extends BaseClassForTests
             client.checkExists().forPath("/hey");
             fail();
         }
-        catch ( KeeperException.ConnectionLossException e )
+        catch ( KeeperException.SessionExpiredException e )
         {
             // correct
         }
--- a/curator-framework/src/test/java/org/apache/curator/framework/imps/TestFrameworkEdges.java
+++ b/curator-framework/src/test/java/org/apache/curator/framework/imps/TestFrameworkEdges.java
@@ -672,7 +672,11 @@ public class TestFrameworkEdges extends BaseClassForTests
         }
         catch ( KeeperException.SessionExpiredException e )
         {
-            // correct
+            // correct, this happens on ZK 3.6.3+
+        }
+        catch ( KeeperException.ConnectionLossException e )
+        {
+            // correct, this happens on ZK 3.5.x, 3.6.0 -> 3.6.2
         }
         finally
         {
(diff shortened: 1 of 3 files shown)


pom.xml
```diff
@@ -1,4 +1,4 @@
-<?xml version="1.0" encoding="UTF-8"?><!--
+e?xml version="1.0" encoding="UTF-8"?><!--
```
madrob, line 1: ?

virajjasani:
Thanks @eolivelli.

Build failure:
```
The command "eval mvn install -DskipTests=true -Dmaven.javadoc.skip=true -B -V " failed
```
Is this relevant or some other known issue?

pom.xml
eolivelli: it was not intended, thanks @madrob

eolivelli:
This patch is still a draft, because some tests are still failing

eolivelli:
@cammckenzie I have fixed the problem for zk35 compatibility test.

The weird behaviour is about a ConnectionLoss vs SessionExpired that happens in a test from 3.6.3 onwards

curator-framework/src/test/java/org/apache/curator/framework/imps/TestFrameworkEdges.java
```diff
@@ -670,9 +670,13 @@ public void testFailure() throws Exception
             client.checkExists().forPath("/hey");
             fail();
         }
+        catch ( KeeperException.SessionExpiredException e )
+        {
+            // correct, this happens on ZK 3.6.3+
```
tisonkun, line 675: Why? Shall we setup a link to the reason?

It is a bit weird a patch version bump cause behavior changes.

tisonkun (commented):
Thanks for starting this upgrade! Comment inline.

curator-framework/src/test/java/org/apache/curator/framework/imps/TestFrameworkEdges.java
eolivelli: I really do not know why this is happening.
I have looked thru ZK 3.6.3 release notes and I haven't found anything that may be releated.

tisonkun: Fine. It is not a blocker for client library such as Curator, but is interesting...

eolivelli: probably Session Expiration is happening earlier with 3.6.3
with 3.6.2 we still see a ConnectionLoss, that usually becomes a SessionExpired when the client is able to connect again and the ZK server already expired it on the server side

cammckenzie (approved):
Looks good to me, all tests pass.

PR #388, opened 2021-06-29
CURATOR-599 Configurable ZookeeperFactory by ZKClientConfig

liran2000:
Option to use ZooKeeper client config.

This seems mandatory for using zookeeper.request.timeout for preventing the potential race condition of hanging indefinitely, as described at the ticket.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -67,7 +67,7 @@
         <jdk-version>1.${short-jdk-version}</jdk-version>
 
         <!-- versions -->
-        <zookeeper-version>3.6.0</zookeeper-version>
+        <zookeeper-version>3.6.1</zookeeper-version>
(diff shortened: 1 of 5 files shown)


pom.xml
```diff
@@ -67,7 +67,7 @@
         <jdk-version>1.${short-jdk-version}</jdk-version>
 
         <!-- versions -->
-        <zookeeper-version>3.6.0</zookeeper-version>
+        <zookeeper-version>3.6.1</zookeeper-version>
```
liran2000, line 70: Mandatory in order to use:

`new ZooKeeperAdmin(connectString, sessionTimeout, watcher, canBeReadOnly, zkClientConfig)`

PR #389, opened 2021-07-02
Use Awaitility to instead of Thread sheep method.

mattisonchao:
- Use Awaitility to avoid busy waiting.
- Refactor some code to Java8

PR #392, opened 2021-07-13
CURATOR-603 Use Awaitility to instead of Thread sleep method in recipes module.

mattisonchao:
Use Awaitility to instead of Thread sleep method in recipes module.

### DIFF
--- a/curator-recipes/pom.xml
+++ b/curator-recipes/pom.xml
@@ -86,6 +86,12 @@
             <artifactId>commons-math</artifactId>
             <scope>test</scope>
         </dependency>
+
+        <dependency>
+            <groupId>org.awaitility</groupId>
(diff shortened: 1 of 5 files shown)


tisonkun:
I suggest you start a discussion on dev@curator.apache.org first. This is not a trivial improvement but which suite we'd prefer.

I'll close this pull request before we reach a consensus.

mattisonchao:
Sure, and thanks for your reply.

tisonkun:
@mattisonchao I just notice we already use awaitility in curator-client https://github.com/apache/curator/pull/389.

Then I'm OK to continue the patch.

tisonkun:
@mattisonchao if you'd like to continue it, please merge master and ping me for review.

PR #394, opened 2021-08-04
CURATOR-607: InterProcessReadWriteLock should expose exposing getLockPath

faucct:
Another and simpler alternative would be exposing it directly in `InterProcessReadWriteLock`:
```
    protected String getLockPath()
    {
        final var lockPath = readMutex.getLockPath();
        if (lockPath != null) {
            return lockPath;
        }
        return writeMutex.getLockPath();
    }
```
https://issues.apache.org/jira/browse/CURATOR-607

### DIFF
--- a/curator-recipes/src/test/java/org/apache/curator/framework/recipes/locks/TestInterProcessReadWriteLock.java
+++ b/curator-recipes/src/test/java/org/apache/curator/framework/recipes/locks/TestInterProcessReadWriteLock.java
@@ -24,12 +24,15 @@ import static org.junit.jupiter.api.Assertions.assertEquals;
 import static org.junit.jupiter.api.Assertions.assertFalse;
 import static org.junit.jupiter.api.Assertions.assertNotNull;
 import static org.junit.jupiter.api.Assertions.assertTrue;
+import static org.junit.jupiter.api.Assertions.fail;
 import com.google.common.collect.Lists;
 import org.apache.curator.framework.CuratorFramework;
 import org.apache.curator.framework.CuratorFrameworkFactory;
 import org.apache.curator.framework.imps.TestCleanState;
 import org.apache.curator.retry.RetryOneTime;
 import org.apache.curator.test.BaseClassForTests;
+import org.apache.curator.test.KillSession;
+import org.apache.zookeeper.KeeperException;
 import org.junit.jupiter.api.Test;
 
 import java.util.Collection;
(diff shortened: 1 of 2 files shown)


curator-recipes/src/test/java/org/apache/curator/framework/recipes/locks/TestInterProcessReadWriteLock.java
```diff
@@ -362,4 +362,77 @@ private void doLocking(InterProcessLock lock, AtomicInteger concurrentCount, Ato
             }
         }
     }
+
+    public static class LockPathInterProcessReadWriteLock extends InterProcessReadWriteLock
```
eolivelli, line 366: This class seems not used.

Are you missing a commit?
Is there any test that covers these new APIs?

eolivelli (changes requested):
Great work.
I did one first pass.
IiUC It looks like we are missing some test coverage for this refactor, can you please add it?

curator-recipes/src/test/java/org/apache/curator/framework/recipes/locks/TestInterProcessReadWriteLock.java
faucct: Yes, it is not used, as I have only added it to demonstrate that there is a way to subclass `InterProcessReadWriteLock` the desired way. I will add a test today.

eolivelli (approved):
LGTM

@Randgalt @cammckenzie PTAL

cammckenzie (approved):

faucct:
bump

faucct:
Also, I think that there would be no need to expose lock-paths and use transactions if Curator exposed APIs with no implicit retries, where everything would happen in same session:

```
var lock = new Lock(path);
while (true) {
  try {
    var zookeeper = curatorFramework.getZookeeperClient().getZooKeeper();
    lock.lock(zookeeper);
    try {
      zookeeper.create("");
      zookeeper.delete("");
      ...
    } finally {
      lock.unlock(zookeeper);
    }
  } catch (Exception e) {
    if (curatorFramework.retry(e)) {
      continue;
    }
    throw e;
  }
}
```

or even

```
var lock = new Lock(path);
curatorFramework.zookeeper(zookeeper -> {
  lock.lock(zookeeper);
  try {
    zookeeper.create("");
    zookeeper.delete("");
    ...
  } finally {
    lock.unlock(zookeeper);
  }
});
```

Would you mind me pull-requesting such APIs?

eolivelli:
@faucct can you start a new discussion foe your proposal on dev@curator? 
Your ideas sound interesting

PR #396, opened 2021-08-05
CURATOR-609 - ModeledCache attempts to deserialize empty ZNodes on deletion, resulting in exceptions

Ryan0751:
On event type NODE_REMOVED, an additional check was added to ensure that the ZNode actually contains data before attempting to deserialize it into a model.

If no model is present for the ZNode, the accept callback is not called (as this is not a valid ModeledFramework object).

This is already accounted for with events NODE_ADDED/NODE_UPDATED, as there was already a check for null data in the node.

Note:  Triggering this bug via a unit test is quite easy, however detecting it is difficult as the exception is thrown by a Curator sub-thread.  Therefore I did not add a matching unit test (though ensured all existing unit tests pass).


curator-x-async/src/main/java/org/apache/curator/x/async/modeled/details/ModeledCacheImpl.java
```diff
@@ -188,9 +188,18 @@ private void internalChildEvent(TreeCacheEvent event)
...
+            T model = null;
+            if (entry != null) {
+                model = entry.model;
+            }
+            else if (event.getData().getData() != null) {
+                model = serializer.deserialize(event.getData().getData());
+            }
+            if (model != null) {
+                Stat stat = (entry != null) ? entry.stat : event.getData().getStat();
+                accept(ModeledCacheListener.Type.NODE_REMOVED, path, stat, model);
```
Randgalt, line 200: This changes the behavior though. Is this correct? Previously, `accept(ModeledCacheListener.Type.NODE_REMOVED...` would always be called.

Ryan0751: You are correct, this change would alter the behavior in that `accept(ModeledCacheListener.NODE_REMOVED... ` would not be called if the node in question doesn't contain any data (ie, a valid Model).

However, I'm not sure this is a bad thing.  If you look above in the code at the implementation for the NODE_ADDED and NODE_UPDATE case, the `accept()` method is also not called if the ZNode in question does not contain any data.

Unit testing doesn't seem to cover this use case either way (the tests all passed without any modification).

eolivelli:
This issue was targeted 2.5.1
I am postponing to 2.5.2, I am sorry

curator-x-async/src/main/java/org/apache/curator/x/async/modeled/details/ModeledCacheImpl.java
kezhuw: > Previously, `accept(ModeledCacheListener.Type.NODE_REMOVED...` would always be called.

It is only true for `ModelSerializer.raw` which I think is a unusual usage of modeled framework. With `JacksonModelSerializer`, the `deserialize` will fail hence `ModeledCacheListener.accept` was not called.

I think it is a chance to unify the two. I prefer the second option as there is no sense to accept a `null` model (`raw` or `json`) for intermediate nodes in all event types. Consistency  should play more importance here than potential breaking of `ModelSerializer.raw`.

@Randgalt What do you think ?

kezhuw:
@Ryan0751 Any interest  to continue this ? I think you can construct a test with `ModeledCacheListener.handleException`. `TestCachedModeledFramework` should be the right place. I can take care of it in case you are not available. We are evaluating this and plan to conclude (merge or not) this before next release.

Ryan0751:
I will take another look!

curator-x-async/src/main/java/org/apache/curator/x/async/modeled/details/ModeledCacheImpl.java
Ryan0751: I might help to mention the use case.

Imagine that we have a model called Widget.

In our application we group Widgets under sub-groups, and we can have many Widgets (100k+).  

As such, we structured ZooKeeper as follows:
/widget/<sub-group A>/<widget 1>
/widget/<sub-group B>/<widget 2>

We want to use ModeledFramework to deserialize widgets, however if we encounter a sub-group ZNode that is part of the ZNode, we'd like to just ignore that rather than cause serialization exceptions.

I will have a look at ModelSerializer.raw and where it fits in here.

Ryan0751: It appears that ModelSerializer.raw is package private an unused. Can you kindly describe its purpose?

Ryan0751:
I have gone ahead and added a unit test for this.

curator-x-async/src/main/java/org/apache/curator/x/async/modeled/details/ModeledCacheImpl.java
kezhuw: `ModelSerializer.raw` is a interface variable which means that it is `public`. Before this pr, the behavior is that:
* With `ModelSerializer.raw` or maybe other serializers that tolerate `null` input, `ModeledCacheListener.accept` will be called with `null` model.
* With `JacksonModelSerializer` or maybe other serializers that do not tolerate `null` input, `ModeledCacheListener.handleException` will be called with exception from `ModelSerializer.deserialize`.

To keep consistent with `NODE_ADDED` and `NODE_UPDATED`, I think all these methods `ModelSerializer.deserialize` and `ModeledCacheListener.accept/handleException` should not be called for intermediate nodes in case of `NODE_REMOVED`. `ModelSerializer` is an open interface, we should provide consistent behavior among `ModelSerializer.raw`, `JacksonModelSerializer` and possible others. It would be helpful if we can assert this in test.

cc @Randgalt @Ryan0751

curator-x-async/src/test/java/org/apache/curator/x/async/modeled/TestCachedModeledFramework.java
```diff
@@ -181,6 +184,72 @@ public void testAccessCacheDirectly()
...
+            @Override
+            public void accept(Type t, ZPath p, Stat s, TestModel m)
+            {
+                if (t == ModeledCacheListener.Type.NODE_ADDED && p.toString().equals(secondModelPath)) {
+                    latch.countDown();
+                }
+            }
+
+            public void handleException(Exception e) {
+                System.err.printf("Caught unexpected exception %s%n", e);
```
kezhuw, line 214: This is unnecessary. Or you could try `AtomicReference<Exception>` and `assertNull` ? This will give us cue about failure reason without resort to `printf`.

kezhuw (approved):
Generally looks good to me. Is it possible to alter the test a bit to accept generic model so we can test both `ModelSerializer.raw` and `JacksonModelSerializer` ? That would be nice for us to ensure that we changed the old behavior and gain consistence.

curator-x-async/src/test/java/org/apache/curator/x/async/modeled/TestCachedModeledFramework.java
Ryan0751: I like storing the Exception.  I tried to assertFail right there, but the exception is swallowed somewhere in the implementation and didn't cause the test to fail.

curator-x-async/src/main/java/org/apache/curator/x/async/modeled/details/ModeledCacheImpl.java
Ryan0751: Thank you for the explanation of ModelSerializer.raw (I totally missed that it was defined in an interface).

As for your second suggestion (not calling deserialized and accept/handleException for intermediate nodes in the case of NODE_REMOVED), does that present a bigger risk of breaking existing users if they expect those nodes to contain valid models?

kezhuw: "intermediate nodes" I means nodes with `null` bytes, e.g. `sub` in your test. The current fix is acceptable to me.

curator-x-async/src/test/java/org/apache/curator/x/async/modeled/TestCachedModeledFramework.java
kezhuw: We are failing what others failed. I experienced this too.

curator-x-async/src/test/java/org/apache/curator/x/async/modeled/TestCachedModeledFramework.java
```diff
@@ -181,6 +184,88 @@ public void testAccessCacheDirectly()
...
+        final String signalModelPath = subPath + "/signal";
+
+        final CountDownLatch latch = new CountDownLatch(1);
+
+        final AtomicReference<Exception> caughtException = new AtomicReference<>(null);
+
+        // Create a custom listener to signal the end of the test and ensure that nothing is thrown.
+        final ModeledCacheListener<T> listener = new ModeledCacheListener<T>() {
+            @Override
+            public void accept(Type t, ZPath p, Stat s, T m)
```
kezhuw, line 224: How about add a assert for non null model ? This way both the two tests will fail without this patch.
* `testEmptyNodeRawDeserialization` fails as it receive `null` input.
* `testEmptyNodeJacksonDeserialization` fails as deserialization failure due to `null` data.

curator-x-async/src/test/java/org/apache/curator/x/async/modeled/TestModeledFrameworkBase.java
```diff
@@ -48,8 +50,8 @@ public void setup() throws Exception
         rawClient.start();
         async = AsyncCuratorFramework.wrap(rawClient);
 
-        JacksonModelSerializer<TestModel> serializer = JacksonModelSerializer.build(TestModel.class);
-        JacksonModelSerializer<TestNewerModel> newSerializer = JacksonModelSerializer.build(TestNewerModel.class);
+        serializer = JacksonModelSerializer.build(TestModel.class);
```
kezhuw, line 53: Unreleated change ?

curator-x-async/src/test/java/org/apache/curator/x/async/modeled/TestCachedModeledFramework.java
Ryan0751: Done.  I had to do this as an AtomicBoolean again, as again the exception doesn't otherwise bubble up and fail the test.

tisonkun:
Merging...

Thank you @Ryan0751!

PR #397, opened 2021-08-08
CURATOR-610 Refactor CountCuratorWatcher in TestWatcherIdentity.java …

wx930910:
Fixes [CURATOR-610](https://issues.apache.org/jira/browse/CURATOR-610)

### Description
Refactor test class [CountCuratorWatcher](https://github.com/apache/curator/blob/4a11aaef8b190dc220d35b7a91df294bfa06250e/curator-framework/src/test/java/org/apache/curator/framework/imps/TestWatcherIdentity.java#L42) by using mocking object created by Mockito.

<hr>

##### Key changed/added classes in this PR
- Create mocking object to replace test subclass `CountCuratorWatcher`, decouple test from production code.
- Make test logic more clear by using method stub instead of method overriding.
- Extract AtomicInteger variable out of the test subclass. Use the extracted variable in assertation statement to check `process(WatchedEvent)` method invocation status.

<hr>

### DIFF
--- a/curator-framework/src/test/java/org/apache/curator/framework/imps/TestWatcherIdentity.java
+++ b/curator-framework/src/test/java/org/apache/curator/framework/imps/TestWatcherIdentity.java
@@ -20,7 +20,13 @@ package org.apache.curator.framework.imps;
 
 import static org.junit.jupiter.api.Assertions.assertEquals;
 import static org.junit.jupiter.api.Assertions.assertFalse;
-import com.google.common.collect.Sets;
+import static org.mockito.Matchers.any;
+import static org.mockito.Mockito.doAnswer;
+import static org.mockito.Mockito.mock;
+
+import java.util.Set;
+import java.util.concurrent.atomic.AtomicInteger;
+
 import org.apache.curator.framework.CuratorFramework;
 import org.apache.curator.framework.CuratorFrameworkFactory;
 import org.apache.curator.framework.api.CuratorWatcher;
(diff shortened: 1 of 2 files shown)


curator-framework/src/test/java/org/apache/curator/framework/imps/TestWatcherIdentity.java
```diff
@@ -32,167 +38,152 @@
...
+				timing.connection(), new RetryOneTime(1));
+		try {
+			client.start();
+			client.create().forPath(PATH);
+			// Add twice the same watcher on the same path
+			client.getData().usingWatcher(watcher).forPath(PATH);
+			client.getData().usingWatcher(watcher).forPath(PATH);
+			// Ok, let's test it
+			client.setData().forPath(PATH, new byte[] {});
+			timing.sleepABit();
```
eolivelli, line 183: What about using Awaiatility and wait for the condition to become 'true'

eolivelli (approved):
Overall ltgm
Thanks!

I left a  suggestion about removing Timing, but we can do it in a follow up task.

curator-framework/src/test/java/org/apache/curator/framework/imps/TestWatcherIdentity.java
wx930910: Make sense to me, thanks for the suggestion, I'll use Awaiatility to replace timeing.

curator-framework/src/test/java/org/apache/curator/framework/imps/TestWatcherIdentity.java
```diff
@@ -32,167 +38,152 @@
...
+	public void testSameCuratorWatcherPerZKDocs() throws Exception {
+		// Create variables for tracking behaviors of mock object
+		AtomicInteger actualWatcherCount = new AtomicInteger(0);
+		// Construct mock object
+		CuratorWatcher actualWatcher = mock(CuratorWatcher.class);
+		// Method Stubs
+		doAnswer((stubInvo) -> {
+			actualWatcherCount.incrementAndGet();
+			return null;
+		}).when(actualWatcher).process(any(WatchedEvent.class));
```
wx930910, line 94: We can further improve test by removing the method stub and the AtomicInteger variable. We can directly verify the behavior of `process(WacthedEvent)` by using Mockito.verify() with timeout function (so we can also remove `timing.sleepABit()`). This will make test condition more explict and elegant.

wx930910 (commented):
Further improve the test logic

curator-framework/src/test/java/org/apache/curator/framework/imps/TestWatcherIdentity.java
```diff
@@ -30,169 +36,141 @@
...
+			client.start();
+			client.create().forPath("/test");
+
+			// per ZK docs, this watcher should only trigger once
+			client.checkExists().usingWatcher(actualWatcher).forPath("/test");
+			client.getData().usingWatcher(actualWatcher).forPath("/test");
+
+			client.setData().forPath("/test", "foo".getBytes());
+			client.delete().forPath("/test");
+			Awaitility.waitAtMost(Duration.ofMillis(TIMEOUT_MS))
```
eolivelli, line 74: We can use the default time out

eolivelli (changes requested):

curator-framework/src/test/java/org/apache/curator/framework/imps/TestWatcherIdentity.java
wx930910: Changed `waitAtMost(Duration)` to `await()`

eolivelli (approved):
LGTM

it looks like there is an indentation issue in the pom.xml file

can you please fix it ?
then I will be happy to merge

eolivelli:
LGTM
I will merge as soon as CI is green

eolivelli:
committed to master branch.
thank you @wx930910

PR #398, opened 2021-10-29
CURATOR-653: fix potential double leader for LeaderLatch

woaishixiaoxiao:
When I use the LeaderLatch to select leader,  there is a double-leader phenomenon.
The timeline is as follows：
1. The zk cluster switch leader node bescause of zxid overflow. The cluster is unavailable to the outside world
2. A client(not leader befor zxid overflow) and B client(is leader before zxid overflow) enter the suspend state, B client set  its leader status to false
3. The zk cluster complete the leader node election and the cluster back to normal
4. A client enter the reconnect state  and  call the reset function, set its leader status to false. 
5. B client enter the reconnect state, call the reset function. set its leader status to false.  Delete its old path.
6. A client receive preNodeDeleteEvent.  Then getChildren from zkServer.  Find itself is the smallest number and set itself as a leader.
7. B client create a new temporary node  and then getChildren from zkServer.  Find itself not the node with the smallest serial number and listen to the previous node delete event.
8. A client delete its old path.
9. B client receive the preNodeDeleteEvent. then getchildren from zkServer. Find itself is the smallest sequence number and then set itself as a leader
10. A client create  a new temporary node  and then getChildren from zkServer.  Find itself not the node with the smallest serial number and listen to the previous node delete event. but it doesn't  set itself as a non-leader state. because of the sixth step operation, A still is leader state now.
11. now  A client and B client are  the leader at the same time


woaishixiaoxiao:
> @woaishixiaoxiao thanks for sharing your fix, do you think that we can add a test case to cover this change ?

HI i have added a unit test. please approval thanks

tisonkun (approved):

tisonkun:
LGTM. I also think of this change days before. cc @eolivelli @Randgalt can you also give a review?

tisonkun:
@woaishixiaoxiao can you create a JIRA ticket on https://issues.apache.org/jira/projects/CURATOR for this patch?

tisonkun (commented):
I think after #430 merged, RECONNECT no more causes `reset`. But it's reasonable to set leadership to false if `getChildren` finds itself not the leader.

cc @eolivelli @Randgalt

tisonkun:
Well. After #430 merged the test added in this patch failed. Need a closer look.

tisonkun:
I adjust the test to inject force `reset`s instead of depending on connection loss. Although this means it should be a non-real-world case now, I still agree on `setLeadership(false)` on `checkLeadership` find the latch isn't the leader. `setLeadership(false)` is idempotent.

eolivelli (approved):
+1

eolivelli:
@tisonkun thanks for fixing the test
@woaishixiaoxiao do you agree with @tisonkun 's fix ?

curator-recipes/src/test/java/org/apache/curator/framework/recipes/leader/TestLeaderLatch.java
```diff
@@ -220,6 +223,108 @@ public void testWatchedNodeDeletedOnReconnect() throws Exception
...
+                };
+                latch.addListener(listener);
+                latch.start();
+                latches.add(latch);
+                assertEquals(states.poll(timing.forWaiting().milliseconds(), TimeUnit.MILLISECONDS), ConnectionState.CONNECTED.name());
+                if (i == 0) {
+                    assertEquals(states.poll(timing.forWaiting().milliseconds(), TimeUnit.MILLISECONDS), "true");
+                }
+            }
+            catch (Exception e){
```
XComp, line 277: Why are we hiding thrown exceptions here? Shouldn't we expose it as part of the test run if something went wrong? :thinking: The test would succeed if the Exception is thrown in this block and caught here.

curator-recipes/src/test/java/org/apache/curator/framework/recipes/leader/TestLeaderLatch.java
```diff
@@ -220,6 +223,108 @@ public void testWatchedNodeDeletedOnReconnect() throws Exception
...
+                    public void notLeader() {
+                        states.add("false");
+                    }
+                };
+                latch.addListener(listener);
+                latch.start();
+                latches.add(latch);
+                assertEquals(states.poll(timing.forWaiting().milliseconds(), TimeUnit.MILLISECONDS), ConnectionState.CONNECTED.name());
+                if (i == 0) {
+                    assertEquals(states.poll(timing.forWaiting().milliseconds(), TimeUnit.MILLISECONDS), "true");
```
XComp, line 274:
```suggestion
                    assertEquals("The first LeaderLatch instance should acquire leadership.", states.poll(timing.forWaiting().milliseconds(), TimeUnit.MILLISECONDS), "true");
```
nit: maybe adding a bit more context to this polling here to describe the test case

curator-recipes/src/test/java/org/apache/curator/framework/recipes/leader/TestLeaderLatch.java
```diff
@@ -220,6 +223,108 @@ public void testWatchedNodeDeletedOnReconnect() throws Exception
...
+                }
+            }
+            catch (Exception e){
+                return;
+            }
+        }
+        timing.forWaiting().sleepABit();
+        // now latch1 is leader, latch2 is not leader. latch2 listens to the ephemeral node created by latch1
```
XComp, line 282: nit: Moving comments into assert messages improves the test output and still works as some kind of comment. This comment could be added to the `assertTrue` and `assertFalse` in line 284-285 below describing the currently expected state.

curator-recipes/src/test/java/org/apache/curator/framework/recipes/leader/TestLeaderLatch.java
```diff
@@ -220,6 +223,108 @@ public void testWatchedNodeDeletedOnReconnect() throws Exception
...
+        assertTrue(latch1.hasLeadership());
+        assertFalse(latch2.hasLeadership());
+        try {
+            latch2.debugRestWaitBeforeNodeDelete = new CountDownLatch(1);
+            latch2.debugResetWaitLatch = new CountDownLatch(1);
+            latch1.debugResetWaitLatch = new CountDownLatch(1);
+
+            // force latch1 and latch2 reset
+            latch1.reset();
+            ForkJoinPool.commonPool().submit(() -> {
```
XComp, line 294: Should we add a comment here on why we're calling `latch2.reset()` in a separate thread? AFAIU, it's done to not make the test's thread block due to `latch2.debugRestWaitBeforeNodeDelete`. It might help readers if this is reflected in a comment here. WDYT?

curator-recipes/src/test/java/org/apache/curator/framework/recipes/leader/TestLeaderLatch.java
```diff
@@ -220,6 +223,108 @@ public void testWatchedNodeDeletedOnReconnect() throws Exception
...
+                        .retryPolicy(new RetryOneTime(1))
+                        .connectionStateErrorPolicy(new StandardConnectionStateErrorPolicy())
+                        .build();
+                ConnectionStateListener stateListener = new ConnectionStateListener()
+                {
+                    @Override
+                    public void stateChanged(CuratorFramework client, ConnectionState newState)
+                    {
+                        if (newState == ConnectionState.CONNECTED) {
+                            states.add(newState.name());
```
XComp, line 250: nit: Adding the `LeaderLatch` ID (we could use for loop `i`) here might help further down in the test understanding the state of the queue.

curator-recipes/src/test/java/org/apache/curator/framework/recipes/leader/TestLeaderLatch.java
```diff
@@ -220,6 +223,108 @@ public void testWatchedNodeDeletedOnReconnect() throws Exception
...
+                    @Override
+                    public void isLeader() {
+                        states.add("true");
+                    }
+
+                    @Override
+                    public void notLeader() {
+                        states.add("false");
+                    }
+                };
```
XComp, line 268: Using the `LeaderLatch` ID in the event labels (as mentioned above) might help when evaluating the queue later on in the test. But to be fair: doing the asserts on `hasLeadership` like it's already done below (lines 303-304) serves the same purpose. I just mention it as another idea here. :shrug:

curator-recipes/src/test/java/org/apache/curator/framework/recipes/leader/TestLeaderLatch.java
```diff
@@ -220,6 +223,108 @@ public void testWatchedNodeDeletedOnReconnect() throws Exception
...
+                }
+            }
+            catch (Exception e){
+                return;
+            }
+        }
+        timing.forWaiting().sleepABit();
+        // now latch1 is leader, latch2 is not leader. latch2 listens to the ephemeral node created by latch1
+        LeaderLatch latch1 = latches.get(0);
+        LeaderLatch latch2 = latches.get(1);
```
XComp, line 284:
```suggestion
        LeaderLatch initialLeaderLatch = latches.get(0);
        LeaderLatch initialNonLeaderLatch = latches.get(1);
```
nit: maybe, making the variable more descriptive to avoid confusion. Especially because we're switching the order here in comparison to what is described in the PR description and the corresponding ticket.

curator-recipes/src/test/java/org/apache/curator/framework/recipes/leader/TestLeaderLatch.java
```diff
@@ -220,6 +223,108 @@ public void testWatchedNodeDeletedOnReconnect() throws Exception
...
+                assertEquals(states.poll(timing.forWaiting().milliseconds(), TimeUnit.MILLISECONDS), ConnectionState.CONNECTED.name());
+                if (i == 0) {
+                    assertEquals(states.poll(timing.forWaiting().milliseconds(), TimeUnit.MILLISECONDS), "true");
+                }
+            }
+            catch (Exception e){
+                return;
+            }
+        }
+        timing.forWaiting().sleepABit();
```
XComp, line 281: What's the purpose of waiting here? :thinking:

XComp (changes requested):
The change looks good. I went over the test and it does what it should do. There are only two things that require an actual change (typo in the variable and catch clause in the test implementation hiding any Exception).

About the other nit-picky comments: I'm aware that I'm nit picking here. Hence, see them as proposal rather than requests to change. :-) Nice proposal :+1:

tisonkun:
@XComp Thank you very much for the comments. Could you send a pull request based on the current patch? I don't know whether I can directly merge on the forked repository but at least I can submit the patch when you made it :)

XComp:
> @XComp Thank you very much for the comments. Could you send a pull request based on the current patch? I don't know whether I can directly merge on the forked repository but at least I can submit the patch when you made it :)

I created PR #436 but couldn't base it onto this PR. I handled my review comments individually to make it easier to select relevant changes. The last commit is a bigger refactoring of the test. You might want to leave that out if you think that it's too much of a change.

tisonkun:
> but couldn't base it onto this PR

Actually, you can send a pull request with base branch = https://github.com/woaishixiaoxiao/curator/tree/fixbug/LeaderLatch-double-master and head branch = your branch :)

XComp:
It was a bit trickier but I managed to base the PR based on his branch (his branch didn't show up when I was doing it the way I usually create PRs ¯\_(ツ)_/¯). https://github.com/woaishixiaoxiao/curator/pull/1. Although, I'm not sure whether that's of any help because the original author would have to merge it into his branch, wouldn't he?

tisonkun:
@XComp OK. After looking into the patch I think #436 is better to proceed (you made a significant diff :)). I'll try to take a look this wek.

XComp:
Yes, but the biggest part of the diff is the last commit ([adaee91](https://github.com/apache/curator/pull/436/commits/adaee91289014f06b60e512b8377de7f9fe4ebd6)): I'm ok with reverting that one if you think it's too much. The changes related to the comments of my review of this PR are included in the commits excluding [adaee91](https://github.com/apache/curator/pull/436/commits/adaee91289014f06b60e512b8377de7f9fe4ebd6) in #436 .

PR #399, opened 2021-11-03
[CURATOR-561] Reset connection after repeat expiry

scott-kirk:
If there is a problem posting the Expired KeeperState
during a session expiration, then the ZooKeeper event
thread will die without ever posting the Expired event.
This would then cause curator to keep trying to expire
the connection but it does nothing because the connection
is dead and no events will ever be posted.

This can be prevented by forcibly resetting the connection
if it's detected that the previous expiry had no effect

https://issues.apache.org/jira/browse/CURATOR-561


cammckenzie (approved):
LGTM, Thanks for the PR.

scott-kirk:
@cammckenzie I need to close this and re-open from a different fork, apologies for the inconvenience. New PR will be this commit cherry-picked at least.

PR #404, opened 2021-12-15
CURATOR-624 Fix GitHub actions

nicoloboschi:
We need to specify the JDK distribution. I set the 'adopt' one

### DIFF
--- a/.github/workflows/ci-unit-jdk11.yaml
+++ b/.github/workflows/ci-unit-jdk11.yaml
@@ -41,6 +41,7 @@ jobs:
         uses: actions/setup-java@v2
         with:
           java-version: 11
+          distribution: "adopt"
 
       - name: Build
         run: mvn -B clean install -DskipTests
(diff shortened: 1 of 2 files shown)


nicoloboschi:
closing to retrigger actions

martin-g:
I've just created https://github.com/apache/curator/pull/407 and now I noticed this PR.
IMO mine is a bit better because of two reasons:
* AdoptJDK has been donated to Eclipse Foundation and now its name is `Temurin`
* I've merged the two workflows into one and used `strategy.matrix` to test with the different versions. This way it would be easier to test with more JDK versions, e.g. `17`

eolivelli:
thank you @nicoloboschi and @martin-g 

I think @martin-g 's patch is better

eolivelli:
@nicoloboschi I am sorry this patch became stale

nicoloboschi:
no problem, we can merge the other one

PR #405, opened 2021-12-16
Fix Compatibility.getHostAddress NullPointerException

wangkaish:
Fix Compatibility.getHostAddress NullPointerException, use 'address.getHostName()' when 'address.getAddress()' is null.


tisonkun (changes requested):
Please create a JIRA ticket for this patch. Also, is there a real world workload trigger this code path? If the first call never returns null, we don't have to make this change.

tisonkun:
Closed. You may reopen this thread with a valid test case or real world issue.

Perhaps you can take a look at https://github.com/apache/curator/pull/425/files which shares several common intentions.

PR #406, opened 2021-12-17
[CURATOR-535] TestServer random port selection has a race condition

paul8263:
This PR is to mitigate the race condition issue. The random available port is now assigned right before starting Zookeeper testing server.

### DIFF
--- a/curator-test/src/main/java/org/apache/curator/test/InstanceSpec.java
+++ b/curator-test/src/main/java/org/apache/curator/test/InstanceSpec.java
@@ -64,9 +64,9 @@ public class InstanceSpec
     }
 
     private final File dataDirectory;
-    private final int port;
-    private final int electionPort;
-    private final int quorumPort;
+    private int port;
+    private int electionPort;
+    private int quorumPort;
     private final boolean deleteDataDirectoryOnClose;
     private final int serverId;
     private final int tickTime;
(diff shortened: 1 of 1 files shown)


tisonkun (commented):
It seems that with this patch it's still possible that a resolved port be in used by other process before it's used by the caller.

paul8263:
> It seems that with this patch it's still possible that a resolved port be in used by other process before it's used by the caller.

Yes. currently this PR updates InstanceSpec so that the random port will only be allocated when it is actually used.
But There are still some more steps before the port is actually used: wrap the configuration into QuorumPeerConfig and start the server. There could be a race condition in these steps.
```java    
QuorumPeerConfig config = configBuilder.buildConfig(thisInstanceIndex);
main.runFromConfig(config);
```
I suggest we could introduce a file lock in order to alleviate the race condition among processes. But the implementation might be cumbersome.

paul8263:
Hi @tisonkun @eolivelli and @Randgalt ,
Thank you for your reply. This problem is unusual. I got this problem when running unit tests in other project which relies on Zookeeper. The unit tests are running parallelly so that TestServer creating process might get a race condition when allocating unused ports. Currently my solution is the steps below:
1. Get a random unused port.
2. Implementing a file lock.
3. Allocating the port to TestServer.
4. Check if TestServer starts properly. If it starts successfully, release the file lock.
I would like to move those steps inside TestServer creation and start process. However I worry that introducing a file lock might not be the best solution as it only solves an unusual problem at the cost of performance degradation. 
Have you got any better ideas? I think using the file lock should be considered as the last resort. Correct me if I am wrong.

tisonkun:
Should be fixed by https://github.com/apache/curator/commit/7e7c2079357fbf10081906d7b783f3483bdecbe7 already.

PR #409, opened 2022-02-25
CURATOR-633: Fix backward compatibility in tests

martin-g:
Run TestConnectionStateManager#testConnectionStateRecoversFromUnexpectedExpiredConnection only for Zookeeper 3.6.x


curator-framework/src/test/java/org/apache/curator/framework/state/TestConnectionStateManager.java
```diff
@@ -92,6 +94,7 @@ public void stateChanged(CuratorFramework client, ConnectionState newState)
 
     @Test
     public void testConnectionStateRecoversFromUnexpectedExpiredConnection() throws Exception {
+        assumeTrue(() -> (getMajor() == 3 && getMinor() >= 6) || (getMajor() > 4), "Zookeeper version must be 3.6 or higher");
```
martin-g, line 97: This does not really work...
It seems javac inlines the values of the constants and the test still prints `3` and `6` even for `curator-test-zk35` modole.

eolivelli (commented):
So you mean that this tests didn't work with 3.5?

I can't remember about this problem

@Randgalt @tisonkun

martin-g:
Yes!
The build fails with JDK 1.8 and 11.

eolivelli:
Can you rebase in order to pick your changes?

martin-g:
CI is green! 🥳

curator-framework/src/test/java/org/apache/curator/framework/state/TestConnectionStateManager.java
```diff
@@ -91,7 +92,10 @@ public void stateChanged(CuratorFramework client, ConnectionState newState)
     }
 
     @Test
-    public void testConnectionStateRecoversFromUnexpectedExpiredConnection() throws Exception {
+    void testConnectionStateRecoversFromUnexpectedExpiredConnection() throws Exception {
```
tisonkun, line 95: I think it's not logically incompatible.

1. If you remove `@Test` above, this test doesn't actually run in any condition.
2. The original error report on JIRA looks because https://github.com/apache/zookeeper/pull/799 wasn't include in zk 3.5. And the usage of `queueEvent` here can be emulated by bringing back bridges of `queueEvent` like https://github.com/apache/curator/pull/344/ removed - it's not a logically incompatible.

tisonkun: Another perspective is that whether or not we won't to test that it's zk35compatible. If we don't intend to do that, we can move the test to another (new) class and assign it to zk36 group.

martin-g: > 1. If you remove `@Test` above, this test doesn't actually run in any condition.

I am not sure why do you explain this to me. I didn't touch `@Test` and the test is still executed in `curator-framework` Maven module, where Zookeeper is 3.6.3.
The problem is when this test is executed in `curator-test-zk35` module (Zookeeper 3.5.7) where `#queueEvent()` is not available. 
https://issues.apache.org/jira/browse/ZOOKEEPER-3269  is `Fixed" only for `3.6.0`+

eolivelli (approved):

curator-framework/src/test/java/org/apache/curator/framework/state/TestConnectionStateManager.java
tisonkun: FYI https://github.com/apache/curator/pull/411.

@martin-g you can integrate in this PR and close #411.

eolivelli (commented):
I have committed @tisonkun 's patch.
I didn't know we already have those tags.

Thank you very much @martin-g for your efforts

PR #407, opened 2022-02-25
CURATOR-624: Fix Github Actions by adding 'distribution' for JDK setup step

martin-g:
Merge the two workflows into one and use strategy.matrix for the Java
versions.

PR #410, opened 2022-02-25
Test with JDK 17. Update Mockito and JUnit 5.x

martin-g:
Continuation of #407 
I will rebase it to `master` once #407 is merged!


eolivelli (approved):
Lgtm

@nicoloboschi you are working on jdk17 in Pulsar and BookKeeper 
Do you want to take a look?

martin-g:
https://github.com/apache/curator/pull/410/commits/f6bac3372130882ee1991b087283842f542474ad "fixes" the problem in org.apache.curator.framework.imps.GzipCompressionProvider but it does not look nice.
I'd recommend a deeper look here!

curator-framework/src/main/java/org/apache/curator/framework/imps/GzipCompressionProvider.java
```diff
@@ -45,6 +45,23 @@
...
+            // e.g. 18-ea, take 18
+            JAVA_VERSION = Integer.parseInt(JAVA_VERSION_STR.split("-", 2)[0]);
+        } else if (JAVA_VERSION_STR.contains(".")) {
+            // e.g. 17.0.1, take 17
+            JAVA_VERSION = Integer.parseInt(JAVA_VERSION_STR.split("\\.", 2)[0]);
+        } else {
+            JAVA_VERSION = Integer.parseInt(JAVA_VERSION_STR, 10);
+        }
+    }
+    private static final byte OS_BIT = JAVA_VERSION >= 16 ? (byte) -1 : 0;
```
nicoloboschi, line 63: could you add a comment explaining why this is needed ?

nicoloboschi (changes requested):

curator-framework/src/main/java/org/apache/curator/framework/imps/GzipCompressionProvider.java
martin-g: I just reverse engineered it.
I don't know what exactly has changed in JDK 16 that caused the breakage here.
This is the reason I said `I'd recommend a deeper look here!`. Someone who knows this code better may have better idea what is going on.

nicoloboschi: I think this is the related issue 
https://bugs.openjdk.java.net/browse/JDK-8244706

martin-g: I will add it as a comment!

martin-g:
Now the build on JDK 17 fails due to Jersey.
https://github.com/apache/curator/blob/fe50da4904aa4ca4f51b473ad298ae664358ff3c/pom.xml#L84-L86 says that it is not worth it to upgrade.

eolivelli:
I am +1 to upgrading Jersey.
That comment (https://stackoverflow.com/questions/17098341#22033825) is very old.

Also I am not sure about why we are using Jersey in Curator...
do you want to try to upgrade please ?

martin-g:
> Also I am not sure about why we are using Jersey in Curator...
> do you want to try to upgrade please ?

It is used only in curator-x-discovery-server/src/test/java/org/apache/curator/x/discovery/server/jetty_jersey/TestMapsWithJersey.java

Is it valueable ? Should I try to upgrade it or just delete it ?

eolivelli:
> Is it valueable ? Should I try to upgrade it or just delete it ?

if we can keep the existing tests it is better, and it would be great to convert the test

that said, I am not sure about the real value of that test
maybe @Randgalt or @cammckenzie  know better than me.

martin-g:
I've updated Jersey to it latest 2.x version but there is still some problem - the MapDiscoveryContext is not seen as a `@Provider` and `org.apache.curator.x.discovery.server.jetty_jersey.MapDiscoveryResource`'s constructor fails with a NPE. I'll try to debug it later.

tisonkun:
@martin-g code conflict. If you'd like to move forward this effort, please rebase the commits so that I can give a review.

eolivelli (approved):
LGTM

tisonkun:
It seems tests failed steadily.

PR #412, opened 2022-02-28
CURATOR-634: Re-introduce TravisCI for testing on Linux ARM64

martin-g:


martin-g:
Any feedback on the proposal to (re-)use TravisCI for testing on Linux ARM64 ?

tisonkun:
I'm unfamiliar with how the old Travis CI set up. Thus cc @Randgalt @eolivelli please take a look.

.travis.yml
```diff
@@ -0,0 +1,59 @@
...
+# KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations
+# under the License.
+
+os: linux
+dist: focal
+arch: arm64-graviton2
+group: edge
+virt: vm
+language: generic
```
tisonkun, line 23:
```suggestion
language: java
```

Why not use a java env? In this approach you don't have to install maven manually.

tisonkun (commented):
@martin-g I learn more about running travis tasks on Apache projects recently and agree that this patch is good to merge. Comments inline.

martin-g:
Because this way it will use a rather old version of Maven. I don't
remember now whether Curator uses maven-enforcer-plugin or not.

tisonkun:
@martin-g Perhaps you can introduce a maven wrapper in curator repository first. It will be a good solution for such issues.

martin-g:
I'd rather just close the PR than play that limbo.

PR #413, opened 2022-03-02
CURATOR-630 Upgrade jetty to 9.4.latest

horvathdora:


madrob:
I believe the latest Jetty 9 is 9.4.45, but also jetty 9 is slated for end of life in April. I would look at 10 or 11 for the upgrade

horvathdora:
Good point @madrob, but Jetty 10 and 11 require Java 11, so we targeted the latest jetty version that is compatible with Java 8.

stoty:
There are quite a lot whitespace changes in this patch.
I am not familair with the curator coding conventions, but reverting the whitespace changes would make the patch easier to review, and possibly avoid breaking the code conventions.

curator-x-discovery-server/src/test/java/org/apache/curator/x/discovery/server/jetty_resteasy/TestStringsWithRestEasy.java
```diff
@@ -65,7 +65,11 @@ public void         setup() throws Exception
 
         port = InstanceSpec.getRandomPort();
         server = new Server(port);
-        Context root = new Context(server, "/", Context.SESSIONS);
+        ServletContextHandler root = new ServletContextHandler(ServletContextHandler.SESSIONS);
+        root.setContextPath("/");
+        server.setHandler(root);
```
stoty, line 70: nit:
Would be cleaner to set up the root object completely, and call setHandler after.

stoty (commented):
I don't see any more issues, but this should get approval from an actual Curator developer.

tisonkun (approved):
Thanks for your contribution! Merging...

PR #414, opened 2022-03-31
CURATOR-537: Fix effective path can be used as a fencing token of LeaderLatch

tisonkun:
This is a follow up to #324.

ourPath can be modified after it's retrived in checkLeadership and before isLeader saves it by ourPath.get() again - if the connection reset and node be recreated.

To avoid handling multiple concurrent cases, this patch simply saves the last node is leader as the localOurPath so that it's always the last valid fencing token - it can be verified as invalid later, but never false valid.

Signed-off-by: tison <wander4096@gmail.com>

cc @Randgalt @eolivelli @cammckenzie I'm unsure whether I get the data race possibility right. It seems that ourPath can be false valid only if it's `reset` concurrently. Other ops are synced by the node, but handle state on reconnect can be async.


tisonkun:
ping @Randgalt @eolivelli @cammckenzie

cammckenzie:
@tisonkun The change looks fine to me. My only concern is that we're modifying the name / function of a public API. Perhaps we should leave the existing implementation of getOurPath and add a new getLastPathIsLeader method?

cammckenzie (approved):

PR #416, opened 2022-04-12
Fix DistributedDoubleBarrier enter and leave methods because using the same watcher affects each other

nailsoul:
Scenario: QTY clients call the enter method, QTY-1 clients call the leave method, the number of clients is not enough so it blocks, and the working directory leaves the smallest node. At this time, the smallest node client receives the notification that the ready node is created to wake up the thread , it is checked that only one node deletes the node, and other clients are woken up
Solution: use different watchers for enter and leaver methods
Test case: TestDistributedDoubleBarrier.testEnterAffectLeaveByWatcher


tisonkun:
@nailsoul sounds reasonable. Could you file a JIRA ticket on https://issues.apache.org/jira/projects/CURATOR? You can create an ASF JIRA account via https://selfserve.apache.org/jira-account.html if you don't have one yet.

tisonkun:
Closing as inactive.

PR #417, opened 2022-04-23
Running JUnit Tests in Parallel with Maven

shisheng-1:
According to [Maven parallel test](https://www.baeldung.com/maven-junit-parallel-tests), we can run tests in parallel.

=====================
If there are any inappropriate modifications in this PR, please give me a reply and I will change them.


tisonkun:
@shisheng-1 could you create a corresponding JIRA ticket and share a report about the CI cost before and after running in parallel?

tisonkun:
@shisheng-1 it seems that before and after, time cost are always about 50 minutes.

tisonkun:
Closed as inactive.

PR #418, opened 2022-05-19
(test) Add test for InstanceSpec.<init> and .getQuorumPort

lacinoire:
Hey 😊
I want to contribute the following test:

Test that `spec.getQuorumPort()` is equal to `2`.
This tests the methods [`InstanceSpec.<init>`](https://github.com/apache/curator/blob/a5b24b8b5d954aea17fc0b1a50d23bb460482658/curator-test/src/main/java/org/apache/curator/test/InstanceSpec.java#L170) and [`InstanceSpec.getQuorumPort`](https://github.com/apache/curator/blob/a5b24b8b5d954aea17fc0b1a50d23bb460482658/curator-test/src/main/java/org/apache/curator/test/InstanceSpec.java#L204).
This test is based on the test [`setCustomTickTimeTest`](https://github.com/apache/curator/blob/a5b24b8b5d954aea17fc0b1a50d23bb460482658/curator-test/src/test/java/org/apache/curator/test/TestTestingServer.java#L37).

Curious to hear what you think!

(I wrote this test as part of a research study at TU Delft. [Find out more](https://github.com/lacinoire/lacinoire/blob/main/README.md))

### DIFF
--- a/curator-test/src/test/java/org/apache/curator/test/TestTestingServer.java
+++ b/curator-test/src/test/java/org/apache/curator/test/TestTestingServer.java
@@ -25,6 +25,7 @@ import static org.junit.jupiter.api.Assertions.assertEquals;
 import java.io.File;
 
 import org.apache.zookeeper.server.ZooKeeperServer;
+import org.junit.jupiter.api.Assertions;
 import org.junit.jupiter.api.Test;
 import org.junit.jupiter.api.io.TempDir;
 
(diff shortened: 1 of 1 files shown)


tisonkun:
This patch seems like an explicit test of getter/setter. Basically it's meaningless and we don't test such methods.

Could you elaborate why you'd like to add such a test?

lacinoire:
Good point regarding the getter method.
I still proposed the test case because it also covers new instructions in the constructor of [InstanceSpec](https://github.com/apache/curator/blob/a5b24b8b5d954aea17fc0b1a50d23bb460482658/curator-test/src/main/java/org/apache/curator/test/InstanceSpec.java#L175) (setting the quorum port to the passed value if it is larger than 0).

Do you find it meaningful to test (this kind of) constructor?

tisonkun:
Closed as we don't test for simple getter/setter.

If you're interested in setting port correctly, you may take a look at CURATOR-535, related patches:

* https://github.com/apache/curator/pull/406
* https://github.com/apache/curator/pull/421

PR #383, opened 2021-04-20
CURATOR-594: TestingZooKeeperMain isn't setting tickTime, if configured

franz1981:
https://issues.apache.org/jira/browse/CURATOR-594

PR #421, opened 2022-06-14
CURATOR-535: Fix client port conflict due to untrustworthy random port allocation

kezhuw:
This commit tries to solve port conflict for `TestingServer` if port is unspecified(aka. `port <= 0`):
* Set `InstanceSpec.port` to 0 if port is unspecified.
* Save OS chosen bind port after started to maintain same port across restart.

Ideally, it should be possible to bootstrap `TestingCluster`(with unspecified ports) too with help from `reconfig`. But there are difficulties since election port, quorum port were not designed to be bound to `0` in ZooKeeper. `TestingServer` should be enough for most cases.

Users should resort to other solutions(eg. container) if they got bored by port conflict due to usages of `TestingCluster` and `TestingServer.restart`.

### DIFF
--- a/curator-test/src/main/java/org/apache/curator/test/InstanceSpec.java
+++ b/curator-test/src/main/java/org/apache/curator/test/InstanceSpec.java
@@ -205,6 +205,10 @@ public class InstanceSpec
         return quorumPort;
     }
 
+    /**
+     * @deprecated use {@link TestingServer#getConnectString()} or {@link TestingCluster#getConnectString()} instead
+     */
+    @Deprecated
     public String getConnectString()
     {
         return hostname + ":" + port;
--- a/curator-test/src/main/java/org/apache/curator/test/QuorumConfigBuilder.java
+++ b/curator-test/src/main/java/org/apache/curator/test/QuorumConfigBuilder.java
@@ -97,7 +97,13 @@ public class QuorumConfigBuilder implements Closeable
 
     public QuorumPeerConfig buildConfig(int instanceIndex) throws Exception
     {
-        Properties properties = buildConfigProperties(instanceIndex);
+        InstanceSpec spec = instanceSpecs.get(instanceIndex);
+        return buildConfig(instanceIndex, spec.getPort());
+    }
+
+    public QuorumPeerConfig buildConfig(int instanceIndex, int instancePort) throws Exception
+    {
+        Properties properties = buildConfigProperties(instanceIndex, instancePort);
         QuorumPeerConfig config = new QuorumPeerConfig()
         {
             {
@@ -112,6 +118,12 @@ public class QuorumConfigBuilder implements Closeable
     }
 
     public Properties buildConfigProperties(int instanceIndex) throws Exception
+    {
+        InstanceSpec spec = instanceSpecs.get(instanceIndex);
+        return buildConfigProperties(instanceIndex, spec.getPort());
+    }
+
+    public Properties buildConfigProperties(int instanceIndex, int instancePort) throws Exception
     {
         boolean isCluster = (instanceSpecs.size() > 1);
         InstanceSpec spec = instanceSpecs.get(instanceIndex);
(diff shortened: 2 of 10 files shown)


kezhuw:
Ideally, it should be possible to bootstrap `TestingCluster`(with unspecified ports) too with help from `reconfig`. But there are difficulties since election port, quorum port were not designed to be bound to `0` in ZooKeeper. `TestingServer` should be enough for most cases, and users should resort to other solutions(eg. container) if they got bored by port conflict.

eolivelli (approved):
LGTM

tisonkun:
@kezhuw sorry for the late review. I'm going to merge this patch if the conflict gets resolved. Would you continue this patch?

eolivelli (commented):
As we need to use reflection to access the ZK server internals, what about sending a path to ZooKeeper in order to allow us to do this stuff without reflection?
Otherwise there is a good chance that we won't be compatible with future ZooKeeper releases if they do some refactoring

kezhuw:
@eolivelli I have created [ZOOKEEPER-4670](https://issues.apache.org/jira/browse/ZOOKEEPER-4670) for your point. It is would be really nice to code without reflection. But I noticed that CURATOR-587(#434) did not require bumping of zookeeper dependency. If we are going to depend on ZOOKEEPER-4670 then clients will have to sync their zookeeper dependencies with curator. Is it ok for this ?

kezhuw:
I think I could drop most of reflection usages but with one exception. The reflection in `ZooKeeperServerEmbeddedAdapter` is mandatory due to [ZOOKEEPER-4303](https://issues.apache.org/jira/browse/ZOOKEEPER-4303)(`ZooKeeperServerEmbeddedImpl` does not support bind to port 0 until apache/zookeeper#1868 which has not been released in any versions.)

kezhuw:
@eolivelli  I have dropped most reflection usages except `ZooKeeperServerEmbeddedAdapter` where I documented clearly its purpose(circumventing [ZOOKEEPER-4303](https://issues.apache.org/jira/browse/ZOOKEEPER-4303)).  

Alternative, I think we can wait until possible 3.7.2 with fix for ZOOKEEPER-4303.

eolivelli (approved):
This even looks better !


@tisonkun @Randgalt @cammckenzie would you mind taking a look ?

curator-test/src/main/java/org/apache/curator/test/InstanceSpec.java
```diff
@@ -205,6 +205,10 @@ public int getQuorumPort()
         return quorumPort;
     }
 
+    /**
+     * @deprecated use {@link TestingServer#getConnectString()} or {@link TestingCluster#getConnectString()} instead
+     */
+    @Deprecated
```
tisonkun, line 211: If we already remove `public QuorumPeerConfig getConfig() throws Exception {` which breaks compatibility, what if we remove this deprecated mark and change the visibility to default (pacakge)?

tisonkun: ... and why do we remove `ZooKeeperMainFace#getConfig`?

tisonkun:
The rest looks good. Comments above.

curator-test/src/main/java/org/apache/curator/test/InstanceSpec.java
kezhuw: Actually, this pr mixed two things: CURATOR-535 and bring back `TestTestingServer#setCustomTickTimeTest` for #383.

`ZooKeeperMainFace#getConfig` was introduced in #434, and its sole external usage in `TestTestingServer#setCustomTickTimeTest` undermines #383.

Comparing to public `InstanceSpec`, I think `ZooKeeperMainFace` is more like an internal concept. May be we can restrict visibility of `ZooKeeperMainFace` to package ? This should solve all above concerns if make sense to you.

tisonkun: @kezhuw That sounds reasonable. And we can add `getTickTime` or something so that we don't have to change `hasZooKeeperServerEmbedded`.

kezhuw: I have removed `public` from `ZooKeeperMainFace`. @tisonkun 

`getTickTime` needs reflection as `ZooKeeperServerEmbedded` exposes no such info([ZOOKEEPER-4670](https://issues.apache.org/jira/browse/ZOOKEEPER-4670)). In previous version(8eda7d0f73333a5e0781d38006065eac98a012a4), I exposed a get method in `ZooKeeperMainFace` to retrieve `ServerCnxnFactory` where both `getTickTime` and `getClientPort` could be easily constructed . But that approach depends heavily on reflection as `QuorumPeerMain` and `ZooKeeperServerEmbedded` exposes no such info.

tisonkun (approved):
LGTM. Further improvement can be invesitigated later.

eolivelli:
@kezhuw I can't assign the JIRA issue to you, but I have committed the patch
thank you very much for your contribution

PR #380, opened 2021-03-06
[CURATOR-590] Add option to disable parent creation for PersistentNode

pboutes:
Adds a `useParentCreation` boolean flag to control the parent creation.

If this flag is set to `false`, the `PersistentNode` won't create the underlying znodes with the `createParentContainersIfNeeded()`, meaning that the parent znodes will have to exist beforehand in order for the `PersistentNode` to succeed its creation.

The `useParentCreation` flag is set to `true` by default.

https://issues.apache.org/jira/browse/CURATOR-590

PR #422, opened 2022-06-15
CURATOR-643: Add option to disable parent creation for PersistentTtlNode

pboutes:
The PR makes the `PersistentTtlNode` able to configure the PersistentNode's `useParentCreation` boolean flag introduced in https://github.com/apache/curator/pull/380.

By default, the flag is set to `true`.

https://issues.apache.org/jira/browse/CURATOR-643


tisonkun (approved):
LGTM.

@eolivelli @cammckenzie could you give a review for this patch? It should be trivial to be included :)

cammckenzie (approved):

PR #428, opened 2022-07-04
CURATOR-646: Fix RAT failure reports and add a CI task to prevent regression

tisonkun:


eolivelli (approved):
LGTM

tisonkun:
@eolivelli I noticed that `apache-rat:check` is bind to `validate` phase by default and show be triggered when we run `mvn -B verify`. However it's not triggered before.

It's interesting. Do you have an idea on this case? Or I may send a mail to RAT's list.

PR #429, opened 2022-07-08
CURATOR-562. Follow up to cleanup ConnectionState#checkTimeouts references

tisonkun:
Signed-off-by: tison <wander4096@gmail.com>

### DIFF
--- a/curator-client/src/main/java/org/apache/curator/CuratorConnectionLossException.java
+++ b/curator-client/src/main/java/org/apache/curator/CuratorConnectionLossException.java
@@ -20,10 +20,6 @@ package org.apache.curator;
 
 import org.apache.zookeeper.KeeperException;
 
-/**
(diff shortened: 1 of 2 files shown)


tisonkun:
Previous PR #348.

cc @eolivelli @Randgalt @cammckenzie

cammckenzie:
hey @tisonkun , changes look fine to me and the code is clearer than before. This is functionally identical though? Or am I missing something?

tisonkun:
@cammckenzie Yes the semantic should be identical. It's exactly a followup to this comment (https://github.com/apache/curator/pull/348#discussion_r389262170) where `checkTimeout` has been removed but leaving code in this patch unchanged.

Actually we don't need `CuratorConnectionLossException` anymore, but `logError` needs a placeholder exception. It'd be better if you can find an existing exception as a good replacement.

cammckenzie (approved):

PR #430, opened 2022-07-13
CURATOR-644. CURATOR-645. Fix livelock in LeaderLatch

tisonkun:
See also:

* https://issues.apache.org/jira/browse/CURATOR-644
* https://issues.apache.org/jira/browse/CURATOR-645

## Livelock in details

Here we have two race conditions to cause livelock:

Case 1. Suppose there are two participants, p0 and p1:

T0. p1 is going to watch on preceding node belongs to p0.
T1. p0 gets reconnected, and thus reset its node, and create a new node to prepare watch on p1's node.
T2. p1 find preceding node has gone, and reset itself.

At the moment, p0 and p1 can be in the livelock that never see each other's node and infinitely reset themselves. This is the case reported by CURATOR-645.

Case 2. The similar case can happen even if there is only one participant:

image

~~If we still call `reset` when preceding node deleted by latter set new node, it's a live lock.~~

I cannot find a live lock here. If only the background in the same client are executed in serial, there are always three nodes to create, while with this patch, there are two nodes to create. But it should not create millions of nodes. If it's in case 1, it's possible since there's no guarantee between different clients.

This is the case reported by CURATOR-644.

## Solution

I make two significant changes to resolve these livelock cases:

1. Call `getChildren` instead of `reset` when preceding node not found in callback. This is previously reported in https://github.com/apache/curator/commit/ff4ec29f5958cc9162f0302c02f4ec50c0e796cd#r31770630. I don't find a reason we perform different between callback and watcher for the same condition. And concurrent `reset`s are the trigger for these livelock.
2. Call `getChildren` instead of `reset` when recovered from connection loss. The reason is similar to 1, while if a connection loss or session expire cause our node to be deleted, when `checkLeadership` we can see the condition and call `reset`.

These changes should fix CURATOR-645 and CURATOR-644.

I'm trying to add test cases and such changes must involve more eyes.

### DIFF
--- a/curator-recipes/src/main/java/org/apache/curator/framework/recipes/leader/LeaderLatch.java
+++ b/curator-recipes/src/main/java/org/apache/curator/framework/recipes/leader/LeaderLatch.java
@@ -609,7 +609,7 @@ public class LeaderLatch implements Closeable
         List<String> sortedChildren = LockInternals.getSortedChildren(LOCK_NAME, sorter, children);
         int ourIndex = (localOurPath != null) ? sortedChildren.indexOf(ZKPaths.getNodeFromPath(localOurPath)) : -1;
 
-        log.debug("checkLeadership with id: {},ourPath: {}, children: {}", id, localOurPath, sortedChildren);
+        log.debug("checkLeadership with id: {}, ourPath: {}, children: {}", id, localOurPath, sortedChildren);
 
         if ( ourIndex < 0 )
         {
--- a/curator-recipes/src/main/java/org/apache/curator/framework/recipes/leader/LeaderLatch.java
+++ b/curator-recipes/src/main/java/org/apache/curator/framework/recipes/leader/LeaderLatch.java
@@ -76,8 +76,6 @@ public class LeaderLatch implements Closeable {
     private final AtomicReference<Future<?>> startTask = new AtomicReference<>();
     private final ConnectionStateListener listener = (client, newState) -> handleStateChange(newState);
 
-    private final Object leadershipLock = new Object();
-
     public enum State {
         LATENT,
         STARTED,
@@ -166,34 +164,32 @@ public class LeaderLatch implements Closeable {
      * @param closeMode allows the default close mode to be overridden at the time the latch is closed.
      * @throws IOException errors
      */
-    public void close(CloseMode closeMode) throws IOException {
-        synchronized (leadershipLock) {
-            Preconditions.checkState(state.compareAndSet(State.STARTED, State.CLOSED), "Already closed or has not been started");
-            Preconditions.checkNotNull(closeMode, "closeMode cannot be null");
+    public synchronized void close(CloseMode closeMode) throws IOException {
+        Preconditions.checkState(state.compareAndSet(State.STARTED, State.CLOSED), "Already closed or has not been started");
+        Preconditions.checkNotNull(closeMode, "closeMode cannot be null");
 
-            cancelStartTask();
+        cancelStartTask();
 
-            try {
-                setNode(null);
-                client.removeWatchers();
-            } catch (Exception e) {
-                ThreadUtils.checkInterrupted(e);
-                throw new IOException(e);
-            } finally {
-                client.getConnectionStateListenable().removeListener(listener);
+        try {
+            setNode(null);
+            client.removeWatchers();
+        } catch (Exception e) {
+            ThreadUtils.checkInterrupted(e);
+            throw new IOException(e);
+        } finally {
+            client.getConnectionStateListenable().removeListener(listener);
 
-                switch (closeMode) {
-                    case NOTIFY_LEADER: {
-                        setLeadership(false);
-                        listeners.clear();
-                        break;
-                    }
+            switch (closeMode) {
+                case NOTIFY_LEADER: {
+                    setLeadership(false);
+                    listeners.clear();
+                    break;
+                }
 
-                    case SILENT: {
-                        listeners.clear();
-                        setLeadership(false);
-                        break;
-                    }
+                case SILENT: {
+                    listeners.clear();
+                    setLeadership(false);
+                    break;
                 }
             }
         }
@@ -279,9 +275,9 @@ public class LeaderLatch implements Closeable {
      *                              while waiting
      */
     public void await() throws InterruptedException, EOFException {
-        synchronized (leadershipLock) {
+        synchronized (this) {
             while ((state.get() == State.STARTED) && !hasLeadership.get()) {
-                leadershipLock.wait();
+                wait();
             }
         }
         if (state.get() != State.STARTED) {
@@ -330,7 +326,7 @@ public class LeaderLatch implements Closeable {
     public boolean await(long timeout, TimeUnit unit) throws InterruptedException {
         long waitNanos = TimeUnit.NANOSECONDS.convert(timeout, unit);
 
-        synchronized (leadershipLock) {
+        synchronized (this) {
             while (true) {
                 if (state.get() != State.STARTED) {
                     return false;
@@ -345,7 +341,7 @@ public class LeaderLatch implements Closeable {
                 }
 
                 long startNanos = System.nanoTime();
-                TimeUnit.NANOSECONDS.timedWait(leadershipLock, waitNanos);
+                TimeUnit.NANOSECONDS.timedWait(this, waitNanos);
                 long elapsed = System.nanoTime() - startNanos;
                 waitNanos -= elapsed;
             }
@@ -482,16 +478,14 @@ public class LeaderLatch implements Closeable {
         client.create().creatingParentContainersIfNeeded().withProtection().withMode(CreateMode.EPHEMERAL_SEQUENTIAL).inBackground(callback).forPath(ZKPaths.makePath(latchPath, LOCK_NAME), LeaderSelector.getIdBytes(id));
     }
 
-    private void internalStart() {
-        synchronized (leadershipLock) {
-            if (state.get() == State.STARTED) {
-                client.getConnectionStateListenable().addListener(listener);
-                try {
-                    reset();
-                } catch (Exception e) {
-                    ThreadUtils.checkInterrupted(e);
-                    log.error("An error occurred checking resetting leadership.", e);
-                }
+    private synchronized void internalStart() {
+        if (state.get() == State.STARTED) {
+            client.getConnectionStateListenable().addListener(listener);
+            try {
+                reset();
+            } catch (Exception e) {
+                ThreadUtils.checkInterrupted(e);
+                log.error("An error occurred checking resetting leadership.", e);
             }
         }
     }
@@ -499,43 +493,41 @@ public class LeaderLatch implements Closeable {
     @VisibleForTesting
     volatile CountDownLatch debugCheckLeaderShipLatch = null;
 
-    private void checkLeadership(List<String> children) throws Exception {
+    private synchronized void checkLeadership(List<String> children) throws Exception {
         if (debugCheckLeaderShipLatch != null) {
             debugCheckLeaderShipLatch.await();
         }
 
-        synchronized (leadershipLock) {
-            final String localOurPath = ourPath.get();
-            List<String> sortedChildren = LockInternals.getSortedChildren(LOCK_NAME, sorter, children);
-            int ourIndex = (localOurPath != null) ? sortedChildren.indexOf(ZKPaths.getNodeFromPath(localOurPath)) : -1;
-            if (ourIndex < 0) {
-                log.error("Can't find our node. Resetting. Index: " + ourIndex);
-                reset();
-            } else if (ourIndex == 0) {
-                lastPathIsLeader.set(localOurPath);
-                setLeadership(true);
-            } else {
-                String watchPath = sortedChildren.get(ourIndex - 1);
-                Watcher watcher = event -> {
-                    if (state.get() == State.STARTED && event.getType() == Watcher.Event.EventType.NodeDeleted) {
-                        try {
-                            getChildren();
-                        } catch (Exception ex) {
-                            ThreadUtils.checkInterrupted(ex);
-                            log.error("An error occurred checking the leadership.", ex);
-                        }
-                    }
-                };
-
-                BackgroundCallback callback = (client, event) -> {
-                    if (event.getResultCode() == KeeperException.Code.NONODE.intValue()) {
-                        // previous node is gone - retry getChildren
+        final String localOurPath = ourPath.get();
+        List<String> sortedChildren = LockInternals.getSortedChildren(LOCK_NAME, sorter, children);
+        int ourIndex = (localOurPath != null) ? sortedChildren.indexOf(ZKPaths.getNodeFromPath(localOurPath)) : -1;
+        if (ourIndex < 0) {
+            log.error("Can't find our node. Resetting. Index: " + ourIndex);
+            reset();
+        } else if (ourIndex == 0) {
+            lastPathIsLeader.set(localOurPath);
+            setLeadership(true);
+        } else {
+            String watchPath = sortedChildren.get(ourIndex - 1);
+            Watcher watcher = event -> {
+                if (state.get() == State.STARTED && event.getType() == Watcher.Event.EventType.NodeDeleted) {
+                    try {
                         getChildren();
+                    } catch (Exception ex) {
+                        ThreadUtils.checkInterrupted(ex);
+                        log.error("An error occurred checking the leadership.", ex);
                     }
-                };
-                // use getData() instead of exists() to avoid leaving unneeded watchers which is a type of resource leak
-                client.getData().usingWatcher(watcher).inBackground(callback).forPath(ZKPaths.makePath(latchPath, watchPath));
-            }
+                }
+            };
+
+            BackgroundCallback callback = (client, event) -> {
+                if (event.getResultCode() == KeeperException.Code.NONODE.intValue()) {
+                    // previous node is gone - retry getChildren
+                    getChildren();
+                }
+            };
+            // use getData() instead of exists() to avoid leaving unneeded watchers which is a type of resource leak
+            client.getData().usingWatcher(watcher).inBackground(callback).forPath(ZKPaths.makePath(latchPath, watchPath));
         }
     }
 
@@ -583,18 +575,16 @@ public class LeaderLatch implements Closeable {
         }
     }
 
-    private void setLeadership(boolean newValue) {
-        synchronized (leadershipLock) {
-            boolean oldValue = hasLeadership.getAndSet(newValue);
+    private synchronized void setLeadership(boolean newValue) {
+        boolean oldValue = hasLeadership.getAndSet(newValue);
 
-            if (oldValue && !newValue) { // Lost leadership, was true, now false
-                listeners.forEach(LeaderLatchListener::notLeader);
-            } else if (!oldValue && newValue) { // Gained leadership, was false, now true
-                listeners.forEach(LeaderLatchListener::isLeader);
-            }
-
-            leadershipLock.notifyAll();
+        if (oldValue && !newValue) { // Lost leadership, was true, now false
+            listeners.forEach(LeaderLatchListener::notLeader);
+        } else if (!oldValue && newValue) { // Gained leadership, was false, now true
+            listeners.forEach(LeaderLatchListener::isLeader);
         }
+
+        notifyAll();
     }
 
     private void setNode(String newValue) throws Exception {
--- a/curator-recipes/src/main/java/org/apache/curator/framework/recipes/leader/LeaderLatch.java
+++ b/curator-recipes/src/main/java/org/apache/curator/framework/recipes/leader/LeaderLatch.java
@@ -576,7 +576,7 @@ public class LeaderLatch implements Closeable
     @VisibleForTesting
     volatile CountDownLatch debugCheckLeaderShipLatch = null;
 
-    private synchronized void checkLeadership(List<String> children) throws Exception
+    private void checkLeadership(List<String> children) throws Exception
     {
         if ( debugCheckLeaderShipLatch != null )
         {
--- a/curator-recipes/src/main/java/org/apache/curator/framework/recipes/leader/LeaderLatch.java
+++ b/curator-recipes/src/main/java/org/apache/curator/framework/recipes/leader/LeaderLatch.java
@@ -190,6 +190,12 @@ public class LeaderLatch implements Closeable
         close(closeMode);
     }
 
+    // for testing
+    void closeOnDemand() throws IOException
+    {
+        internalClose(closeMode, false);
+    }
+
     /**
      * Remove this instance from the leadership election. If this instance is the leader, leadership
      * is released. IMPORTANT: the only way to release leadership is by calling close(). All LeaderLatch
@@ -198,9 +204,25 @@ public class LeaderLatch implements Closeable
      * @param closeMode allows the default close mode to be overridden at the time the latch is closed.
      * @throws IOException errors
      */
-    public synchronized void close(CloseMode closeMode) throws IOException
+    public void close(CloseMode closeMode) throws IOException
     {
-        Preconditions.checkState(state.compareAndSet(State.STARTED, State.CLOSED), "Already closed or has not been started");
+        internalClose(closeMode, true);
+    }
+
+    private synchronized void internalClose(CloseMode closeMode, boolean failOnClosed) throws IOException
+    {
+        if (!state.compareAndSet(State.STARTED, State.CLOSED))
+        {
+            if (failOnClosed)
+            {
+                throw new IllegalStateException("Already closed or has not been started");
+            }
+            else
+            {
+                return;
+            }
+        }
+
         Preconditions.checkNotNull(closeMode, "closeMode cannot be null");
 
         cancelStartTask();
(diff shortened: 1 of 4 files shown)


curator-recipes/src/main/java/org/apache/curator/framework/recipes/leader/LeaderLatch.java
```diff
@@ -604,7 +604,7 @@ else if ( ourIndex == 0 )
                 @Override
                 public void process(WatchedEvent event)
                 {
-                    if ( (state.get() == State.STARTED) && (event.getType() == Event.EventType.NodeDeleted) && (localOurPath != null) )
```
tisonkun, line 607: In this situation `localOurPath` must be not null.

curator-recipes/src/main/java/org/apache/curator/framework/recipes/leader/LeaderLatch.java
```diff
@@ -667,9 +667,9 @@ protected void handleStateChange(ConnectionState newState)
             {
                 try
                 {
-                    if ( client.getConnectionStateErrorPolicy().isErrorState(ConnectionState.SUSPENDED) || !hasLeadership.get() )
```
tisonkun, line 670: `getChildren` does nothing if we still hold leadership. And actually even if `hasLeadership` is true, we can say nothing if we coming back from a connection error.

tisonkun:
I notice that background are running serially and the order is the same as request sending. Perhaps we can setup invariant based on this premise.

tisonkun:
cc @rikimberley @yuri-tceretian

curator-recipes/src/test/java/org/apache/curator/framework/recipes/leader/TestLeaderLatch.java
```diff
@@ -218,6 +218,30 @@ public void testWatchedNodeDeletedOnReconnect() throws Exception
         }
     }
 
+    @Test
+    public void testOurPathDeletedOnReconnect() throws Exception
```
tisonkun, line 222: This test can pass on master with 3 nodes created, while with this patch there are 2 nodes created.

XComp (commented):
What's the motivation behind putting the solution for both Jira issues into a single PR? Wouldn't it more reasonable to split it up into two PRs analgously to the Jira issues?

tisonkun:
@XComp they're logically resolved simultaneously. That is, if you resolve CURATOR-644, you resolve CURATOR-645 - they're the same issue.

curator-recipes/src/main/java/org/apache/curator/framework/recipes/leader/LeaderLatch.java
```diff
@@ -667,9 +670,9 @@ protected void handleStateChange(ConnectionState newState)
...
```
XComp, line 670: I'm wondering whether we should remove this `if` entirely: Let's assume the current instance doesn't have the leadership but watches the child node referring to the current leader. If the connection was temporarily suspended it could be that the actual leader lost its leadership in the meantime. In that case, we want the current instance to check the leadership. :thinking:

XComp: But thinking about it once more, this might not be an issue because the current LeaderLatch would have a watcher on the current leader's node which would have been triggered when the leader's child would have been deleted. :thinking:

curator-recipes/src/test/java/org/apache/curator/framework/recipes/leader/TestLeaderLatch.java
XComp: The test is not suitable for the change. I reverted the change and ran the test. The test still succeeded.

XComp (changes requested):
I guess, we have to reiterate over the test once more. It succeeded even with the fix in the production code being reverted.

XComp:
> they're logically resolved simultaneously. That is, if you resolve CURATOR-644, you resolve CURATOR-645 - they're the same sort.
> 
> In another word, you can check out the diff and tell me how to split it up into two PRs.

I was thinking about it once more. CURATOR-645 could be covered separately in my opinion. CURATOR-645 was identified in FLINK-27078 where we run almost no logic before revoking the leadership by calling `LeaderLatch#close`. That caused the current leader's `LeaderLatch` instance to trigger its child node deletion while other `LeaderLatch` instances were right within setting up the watcher for its child node's predecessor.

Hence, I see CURATOR-645 being not that tightly related with the reconnect issue covered in CURATOR-644. CURATOR-645 just needs to be resolved before CURATOR-644 can be resolved.

Anyway, the changes are not that big in the end that we couldn't resolve both in the same PR.  ¯\_(ツ)_/¯

curator-recipes/src/test/java/org/apache/curator/framework/recipes/leader/TestLeaderLatch.java
tisonkun: Yes. As described. I don't find a good test case yet :/

XComp: What do you think of the following test:
```
 @Test
    public void testLeadershipElectionWhenNodeDisappearsAfterChildrenAreRetrieved() throws Exception
    {
        final String latchPath = "/foo/bar";
        final Timing2 timing = new Timing2();
        try (CuratorFramework client = CuratorFrameworkFactory.newClient(server.getConnectString(), timing.session(), timing.connection(), new RetryOneTime(1)))
        {
            client.start();
            LeaderLatch latch0 = new LeaderLatch(client, latchPath, "0");
            latch0.start();

            waitForALeader(Collections.singletonList(latch0), new Timing());

            LeaderLatch latch1 = new LeaderLatch(client, latchPath, "1");
            LeaderLatch latch2 = new LeaderLatch(client, latchPath, "2");

            latch1.debugCheckLeaderShipLatch = new CountDownLatch(1);

            latch1.start();
            timing.sleepABit();

            latch2.start();
            timing.sleepABit();

            latch0.close();

            latch1.debugCheckLeaderShipLatch.countDown();

            waitForALeader(Arrays.asList(latch1, latch2), new Timing());

            latch1.close();
            latch2.close();
        }
    }
```

This test case describes the situation for CURATOR-645. I checked in a local test run, that it will run forever when still using `reset()` in [LeaderLatch:633](https://github.com/apache/curator/blob/425598cb6bf6a5b227a5fdd293fe46c7978d6578/curator-recipes/src/main/java/org/apache/curator/framework/recipes/leader/LeaderLatch.java#L633) but would succeed when using `getChildren()`, instead.

tisonkun:
Thanks for your inputs @XComp!

I'll try to integrate your comment this week or the next.

Since we merge several fixes into the master branch, and there're users asking for a new release, if we don't have a consensus on this patch, I'll push the changes on debug logging first so that reporters of CURATOR-644 and CURATOR-645 can use the new version to provide an exact error log output :)

curator-recipes/src/test/java/org/apache/curator/framework/recipes/leader/TestLeaderLatch.java
tisonkun: Thanks! Integrated.

curator-recipes/src/main/java/org/apache/curator/framework/recipes/leader/LeaderLatch.java
```diff
@@ -717,6 +720,7 @@ else if ( !oldValue && newValue )
     private void setNode(String newValue) throws Exception
     {
         String oldPath = ourPath.getAndSet(newValue);
+        log.debug("setNode with oldPath: {}, newValue: {}", oldPath, newValue);
```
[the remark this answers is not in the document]
tisonkun: Updated.

curator-recipes/src/main/java/org/apache/curator/framework/recipes/leader/LeaderLatch.java
tisonkun: It's for backward compatibility that once we treat ConnectionLoss as ErrorState. I agree with that since we don't give up the leadership (`reset`) actively, we can remove the `if` condition (but still keep `getChildren`, of course).

However, it can be a follow-up improvement that doesn't affect the correctness. You're welcome to make the patch.

tisonkun:
Updated. I believe this patch is ready to merge.

Please help with reviewing @eolivelli @Randgalt @cammckenzie

curator-recipes/src/main/java/org/apache/curator/framework/recipes/leader/LeaderLatch.java
tisonkun: If you take a look at FLINK-10052, the final solution is using a SessionErrorPolicy that will skip this `if` block. Since a ConnectionLoss may be only network unstable instead of the node lost its leadership (the ephemeral node). Before this patch it's `reset` to be called and actively give up the leadership, it will cause reelection, increase ZK workload and cause further inconsistency.

The thorough solution should be something like I proposed in https://github.com/apache/flink/pull/9878, but I failed to contribute it to the upstream (FLINK-10052 takes more than 2 years to be merged. It's not a good experience to me, lol). We run with the solution in Tencent for years and it works well :)

curator-recipes/src/test/java/org/apache/curator/framework/recipes/leader/TestLeaderLatch.java
```diff
@@ -218,6 +218,56 @@ public void testWatchedNodeDeletedOnReconnect() throws Exception
...
+                // we want to make sure that the leader gets leadership before other instances joining the party
+                waitForALeader(Collections.singletonList(latchInitialLeader), new Timing());
+
+                // candidate #0 will wait for the leader to go away - this should happen after the child nodes are retrieved by candidate #0
+                latchCandidate0.debugCheckLeaderShipLatch = new CountDownLatch(1);
+
+                latchCandidate0.start();
+                timing.sleepABit();
+
+                // no extract CountDownLatch needs to be set here because candidate #1 will rely on candidate #0
```
XComp, line 246:
```suggestion
                // no extra CountDownLatch needs to be set here because candidate #1 will rely on candidate #0
```
typo

curator-recipes/src/test/java/org/apache/curator/framework/recipes/leader/TestLeaderLatch.java
```diff
@@ -218,6 +218,56 @@ public void testWatchedNodeDeletedOnReconnect() throws Exception
...
+            client.start();
+            LeaderLatch latchInitialLeader = new LeaderLatch(client, latchPath, "initial-leader");
+            LeaderLatch latchCandidate0 = new LeaderLatch(client, latchPath, "candidate-0");
+            LeaderLatch latchCandidate1 = new LeaderLatch(client, latchPath, "candidate-1");
+
+            try
+            {
+                latchInitialLeader.start();
+
+                // we want to make sure that the leader gets leadership before other instances joining the party
```
XComp, line 237:
```suggestion
                // we want to make sure that the leader gets leadership before other instances are joining the party
```
and another typo

curator-recipes/src/test/java/org/apache/curator/framework/recipes/leader/TestLeaderLatch.java
```diff
@@ -218,6 +218,56 @@ public void testWatchedNodeDeletedOnReconnect() throws Exception
...
+                waitForALeader(Collections.singletonList(latchInitialLeader), new Timing());
+
+                // candidate #0 will wait for the leader to go away - this should happen after the child nodes are retrieved by candidate #0
+                latchCandidate0.debugCheckLeaderShipLatch = new CountDownLatch(1);
+
+                latchCandidate0.start();
+                timing.sleepABit();
```
XComp, line 244: tbh, I'm not really happy with the sleep here and in line 248 because they are a cause for instabilities: The `close` in line 251 has to happen after the child nodes for `candidate #0` and `candidate #1` are created. AFAIU, we cannot ensure that with the sleep calls due to the asynchronous nature of the `start` command.
I tried to add a `waitForCondition` instead, when coming up with this test first, that would wait for the child to be created. Unfortunately, this resulted in the test blocking forever because (I guess) the await on the `latchCandidate0.debugCheckLeaderShipLatch` is called in the main thread which makes any subsequent operation (including the check for children nodes) being blocked.

I hoped that somebody else could come up with a better approach here. :innocent:

XComp (commented):
I went over my proposed test once more and added a few comments. Please see them below.

curator-recipes/src/test/java/org/apache/curator/framework/recipes/leader/TestLeaderLatch.java
```diff
@@ -218,6 +218,56 @@ public void testWatchedNodeDeletedOnReconnect() throws Exception
...
+
+                // no extract CountDownLatch needs to be set here because candidate #1 will rely on candidate #0
+                latchCandidate1.start();
+                timing.sleepABit();
```
XComp, line 248:
```suggestion
                assertNull(latchCandidate0.getOurPath());
                assertNull(latchCandidate1.getOurPath());

                latchCandidate0.start();
                latchCandidate1.start();

                Awaitility.await("Wait for the candidates to have a path assigned to them.")
                    .pollInterval(Duration.ofMillis(100))
                    .pollInSameThread()
                    .until(() -> latchCandidate0.getOurPath() != null && latchCandidate1.getOurPath() != null);
```
Would that work? The `LeaderLatch#ourPath` is set after the child is created. WDYT?

curator-recipes/src/test/java/org/apache/curator/framework/recipes/leader/TestLeaderLatch.java
XComp: There's a proposal to work around that issue in [the comment below](https://github.com/apache/curator/pull/430/files#r969545913)

XComp: I found a better solution for that by introducing another ZK client that is used for monitoring the children. That way, we don't end up being blocked on concurrent calls in the actual test.

tisonkun: @XComp sounds interesting. You can directly make a PR against this patch branch and I'll review it. Sorry that I may not have too much time recently to cooperate on this patch :)

XComp (changes requested):
Going over the PR once more with the new test proposal for [CURATOR-645](https://issues.apache.org/jira/browse/CURATOR-645), I would vote for splitting up the two Jira issues instead of handling them in a single PR. `CURATOR-645` is now clearly defined by the change you proposed and the test case we came up with together.

~We still have to revisit [CURATOR-644](https://issues.apache.org/jira/browse/CURATOR-644) (see [my comment below](https://github.com/apache/curator/pull/430#discussion_r970595872) on the production code change). Additionally,~ we have to come up with a test case for that `CURATOR-644`, still. Anyway, blocking `CURATOR-645` on `CURATOR-644` just because the latter one is based on the former one to solve both in a single PR is not needed and also might cause confusion later on. WDYT?

curator-recipes/src/main/java/org/apache/curator/framework/recipes/leader/LeaderLatch.java
```diff
@@ -667,9 +670,9 @@ protected void handleStateChange(ConnectionState newState)
             {
                 try
                 {
-                    if ( client.getConnectionStateErrorPolicy().isErrorState(ConnectionState.SUSPENDED) || !hasLeadership.get() )
+                    if ( client.getConnectionStateErrorPolicy().isErrorState(ConnectionState.SUSPENDED) )
                     {
-                        reset();
+                        getChildren();
```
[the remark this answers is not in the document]
XComp: Ok, after revisiting your response in [that thread earlier in this PR](https://github.com/apache/curator/pull/430#discussion_r967852427), I notice again that `getChildren` is the proper way of doing it:
* either the corresponding child node is still around: Then we would pick up the leadership again.
* if it's not around, `reset()` will be called, anyway

Sorry for the confusion. ...to many threads -.-

curator-recipes/src/main/java/org/apache/curator/framework/recipes/leader/LeaderLatch.java
```diff
@@ -586,6 +586,9 @@ private void checkLeadership(List<String> children) throws Exception
         final String localOurPath = ourPath.get();
         List<String> sortedChildren = LockInternals.getSortedChildren(LOCK_NAME, sorter, children);
         int ourIndex = (localOurPath != null) ? sortedChildren.indexOf(ZKPaths.getNodeFromPath(localOurPath)) : -1;
+
+        log.debug("checkLeadership with ourPath: {}, children: {}", localOurPath, sortedChildren);
```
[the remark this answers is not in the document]
tisonkun: Updated in e6ab28b80f19ad4c86118cd99ff7a3bee6579889.

curator-recipes/src/test/java/org/apache/curator/framework/recipes/leader/TestLeaderLatch.java
```diff
@@ -218,6 +218,56 @@ public void testWatchedNodeDeletedOnReconnect() throws Exception
...
+
+                // candidate #0 will wait for the leader to go away - this should happen after the child nodes are retrieved by candidate #0
+                latchCandidate0.debugCheckLeaderShipLatch = new CountDownLatch(1);
+
+                latchCandidate0.start();
+                timing.sleepABit();
+
+                // no extra CountDownLatch needs to be set here because candidate #1 will rely on candidate #0
+                latchCandidate1.start();
+                timing.sleepABit();
```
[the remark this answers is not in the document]
tisonkun: Resolved.

curator-recipes/src/main/java/org/apache/curator/framework/recipes/leader/LeaderLatch.java
```diff
@@ -190,6 +190,12 @@ public void close() throws IOException
         close(closeMode);
     }
 
+    // for testing
```
XComp, line 193:
```suggestion
    @VisibleForTesting
```
Isn't this annotation the better way of documenting this here?

curator-recipes/src/main/java/org/apache/curator/framework/recipes/leader/LeaderLatch.java
```diff
@@ -190,6 +190,12 @@ public void close() throws IOException
...
+    void closeOnDemand() throws IOException
+    {
+        internalClose(closeMode, false);
+    }
```
XComp, line 197: For now, this is a test-only method. I'm a bit reluctant to add test-related code in production code rather than providing a utilily class instead (to be fair: we do that in curator code in other locations like the `debug*CountDownLatches` in the `LeaderLatch` implementation). I see a risk if convenient methods like this are available in production: They might make it easier to write less-consistent production code. :thinking:

XComp: ...but if you want to stick to it, I would vote for coming up with a more descriptive name for it: Something like `closeIfStillRunning`

curator-recipes/src/main/java/org/apache/curator/framework/recipes/leader/LeaderLatch.java
```diff
@@ -198,9 +204,25 @@ public void close() throws IOException
...
+        if (!state.compareAndSet(State.STARTED, State.CLOSED))
+        {
+            if (failOnClosed)
+            {
+                throw new IllegalStateException("Already closed or has not been started");
+            }
+            else
+            {
+                return;
+            }
```
XComp, line 223:
```suggestion
            return;
```
nit: Is the `else` branch required based on code guidelines? If not, I'd propose removing it to shorten the code. The code format (having opening brackets on a new line) makes the code long, anyway already.

curator-test-zk35/pom.xml
```diff
@@ -166,6 +166,12 @@
             <artifactId>slf4j-log4j12</artifactId>
             <scope>test</scope>
         </dependency>
+
+        <dependency>
+            <groupId>org.awaitility</groupId>
+            <artifactId>awaitility</artifactId>
+            <scope>test</scope>
+        </dependency>
```
XComp, line 174: Why is this change necessary?

XComp (commented):
Thanks, @tisonkun for applying my comments. I have a few minor things to add. ...mostly cosmetic comments. I'd be curious about your opinion.

curator-test-zk35/pom.xml
tisonkun: https://github.com/apache/curator/actions/runs/3071902247/jobs/4962990393

XComp: I see, sorry that I missed that :+1:

curator-recipes/src/main/java/org/apache/curator/framework/recipes/leader/LeaderLatch.java
tisonkun: I think it's a design issue we can resolve in another thread. Generally, I'd prefer `close` to be idempotent but the original code isn't. Your test case can fail if:

1. Checking state != CLOSE.
2. Internally closed.
3. Outside closed.

And it happens in the CI environment once.

Your suggestion can be reasonable but I don't want to spend too much time on this bug fix PR. A package-level visible method is for internal usage :)

curator-recipes/src/main/java/org/apache/curator/framework/recipes/leader/LeaderLatch.java
tisonkun: Make sense. Although, they're almost the same. So I just keep it as is now.

curator-recipes/src/main/java/org/apache/curator/framework/recipes/leader/LeaderLatch.java
tisonkun: For me, `VisibleForTesting` can mean that this method can be used in production but its visibility is wider than necessary for testing.

But it's fair enough to use `VisibleForTesting`. I don't want to make the change now :)

tisonkun:
@eolivelli @cammckenzie @Randgalt Perhaps we can release a 5.4.0 later this month and I'd ask for a review on this patch for a consensus whether we include it or only the debug logging part.

curator-recipes/src/main/java/org/apache/curator/framework/recipes/leader/LeaderLatch.java
```diff
@@ -586,6 +608,9 @@ private void checkLeadership(List<String> children) throws Exception
         final String localOurPath = ourPath.get();
         List<String> sortedChildren = LockInternals.getSortedChildren(LOCK_NAME, sorter, children);
         int ourIndex = (localOurPath != null) ? sortedChildren.indexOf(ZKPaths.getNodeFromPath(localOurPath)) : -1;
+
+        log.debug("checkLeadership with id: {},ourPath: {}, children: {}", id, localOurPath, sortedChildren);
```
XComp, line 612:
```suggestion
        log.debug("checkLeadership with id: {}, ourPath: {}, children: {}", id, localOurPath, sortedChildren);
```
nit

XComp (approved):
Looks good from my end. Both changes make sense in my opinion and are covered by tests now. :+1:

curator-recipes/src/main/java/org/apache/curator/framework/recipes/leader/LeaderLatch.java
Randgalt, line 193: Use `@VisibleForTesting`

Randgalt (approved):
I don't see any issues, however it's been a very long time since I've looked at this code.

curator-recipes/src/main/java/org/apache/curator/framework/recipes/leader/LeaderLatch.java
tisonkun: OK. cc @XComp as multiple reviewers ask for it, I'm making the change now :)

curator-recipes/src/main/java/org/apache/curator/framework/recipes/leader/LeaderLatch.java
tisonkun: I'm reverting this change because it's a OR logic but I misread it as an AND logic in the first place.

tisonkun: ... while I think we can remove the whole if condition (keep the `getChildren` call. You can give it a try.

tisonkun:
Merging...

I'm looking into a related change #398 and then prepare the next release.

ImagineBrain:
It caused a new issue.
[CURATOR-724](https://issues.apache.org/jira/browse/CURATOR-724)

tisonkun:
@ImagineBrain It may be fixed by https://github.com/apache/curator/pull/500. Would you upgrade to Curator 5.7.0 and recheck?

ImagineBrain:
> @ImagineBrain It may be fixed by #500. Would you upgrade to Curator 5.7.0 and recheck?

@tisonkun I've tried 5.7.1, it helped nothing if the leaderPath didn't come back after reconnected. But I found a workaround, adding a ConnectionStateListener to create the leaderPath node, then all things will be fine.

PR #431, opened 2022-07-17
CURATOR-524. Fix all URLs as https

tisonkun:
Keep `http://www.apache.org/licenses/LICENSE-2.0` as is.


tisonkun:
Trivial changes. Merging since CI passed...

PR #433, opened 2022-08-19
CURATOR-538: address from InetSocketAddress could be null as it means unresolved address

snuyanzin:
To be aligned with contract of `java.net.InetSocketAddress` it should take into account that `java.net.InetSocketAddress#getAddress` could return `null` which will mean unresolved address

### DIFF
--- a/curator-client/src/main/java/org/apache/curator/utils/Compatibility.java
+++ b/curator-client/src/main/java/org/apache/curator/utils/Compatibility.java
@@ -113,7 +113,7 @@ public class Compatibility
                 log.error("Could not call addrField.get({})", server, e);
             }
         }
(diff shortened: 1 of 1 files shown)


tisonkun (commented):
Thanks for your contribution @snuyanzin!

The change looks good. But I don't know how can this patch be relevant to CURATOR-538?

snuyanzin:
hi @tisonkun thanks for the response

in fact there are 3 jira issues for the same situation CURATOR-597, CURATOR-538, CURATOR-649
every fails with NPE since the address was marked as unresolved and as a result it tries to host from address...
but at the same time for unresolved addresses address is `null`.... The idea is return `unknown` in case of unresolved addresses as for other exceptions

tisonkun (approved):
Thanks for your explanation!

Merging...

PR #437, opened 2022-10-17
Add Support for TLS-enabled `TestingZooKeeperMain`

hyperxpro:
When `ServerConfig#getSecureClientPortAddress` is not `null` then user has indicated the use of TLS. In this case, we will configure the Server factory to use TLS.


tisonkun:
The code diff in `ZooKeeperServerMain#runFromConfig` is much larger than this patch.

For modern testing usage, I think we already use `ZooKeeperServerEmbeddedAdapter` internally. `TestingZooKeeperMain ` is somehow deprecated.

PR #446, opened 2023-03-01
CURATOR-518: Fix LeaderSelector requeue broken by interruptLeadership

kezhuw:
Currently, `LeaderSelector.interruptLeadership` clould cancel queued election task before its execution which will break election loop.

### DIFF
--- a/curator-recipes/src/main/java/org/apache/curator/framework/recipes/leader/LeaderSelector.java
+++ b/curator-recipes/src/main/java/org/apache/curator/framework/recipes/leader/LeaderSelector.java
@@ -72,7 +72,10 @@ public class LeaderSelector implements Closeable
     private final InterProcessMutex mutex;
     private final AtomicReference<State> state = new AtomicReference<State>(State.LATENT);
     private final AtomicBoolean autoRequeue = new AtomicBoolean(false);
-    private final AtomicReference<Future<?>> ourTask = new AtomicReference<Future<?>>(null);
+
+    // guarded by synchronization
+    private Future<?> ourTask = null;
+    private Thread ourThread = null;
 
     private volatile boolean hasLeadership;
     private volatile String id = "";
@@ -87,9 +90,6 @@ public class LeaderSelector implements Closeable
         STARTED,
         CLOSED
     }
-
-    // guarded by synchronization
-    private boolean isQueued = false;
 
     private static final ThreadFactory defaultThreadFactory = ThreadUtils.newThreadFactory("LeaderSelector");
 
@@ -224,6 +224,10 @@ public class LeaderSelector implements Closeable
      * happens and false is returned. If the instance was not queued, it is re-queued and true
      * is returned
      *
+     * <p>The attempt will finish after session error, leadership release. This method is inherently
+     * hard to use as there is no public API to guarantee successful requeue. Try {@link #autoRequeue()}
+     * if you are in doubt.</p>
+     *
      * @return true if re-queue is successful
      */
     public boolean requeue()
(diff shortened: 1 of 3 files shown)


eolivelli (approved):
Nice work

eolivelli:
some tests failed. I have restarted CI

curator-recipes/src/main/java/org/apache/curator/framework/recipes/leader/LeaderSelector.java
```diff
@@ -379,10 +423,31 @@ public boolean hasLeadership()
...
+
+    /**
+     * Cancel ongoing election and leadership.
+     *
+     * <p>Subsequent {@link #requeue()} will succeed if {@link #autoRequeue()} is not enabled.
+     */
+    private synchronized void cancelElection() {
+        if ( ourTask != null ) {
+            ourTask.cancel(true);
+            isQueued = false;
```
tisonkun, line 446:
```suggestion
            clearIsQueued();
```

Keep consistent.

curator-recipes/src/main/java/org/apache/curator/framework/recipes/leader/LeaderSelector.java
```diff
@@ -237,13 +242,15 @@ private synchronized boolean internalRequeue()
...
-            Future<Void> task = executorService.submit(new Callable<Void>()
+            Future<?> oldTask = ourTask;
+            ourTask = executorService.submit(new Callable<Void>()
             {
                 @Override
                 public Void call() throws Exception
                 {
                     try
                     {
+                        waitTaskDone(oldTask, Duration.ofMillis(500), Duration.ofSeconds(5));
```
tisonkun, line 253: It seems this is a best-effort waiting call and we don't have happens-before relation to ensure that the previous task must be done or cancelled.

Filter only `hasLeadership` will proceed `interruptLeadership` seems enough for fixing 518?

tisonkun (commented):
Thanks for your contribution! Comments inline.

curator-recipes/src/main/java/org/apache/curator/framework/recipes/leader/LeaderSelector.java
kezhuw: I thought about the guarantee `LeaderSelector` try to provide, and conclude to that:
* It is crucial for `LeaderSelectorListener.takeLeadership` to not swallow interruption silently without return. Otherwise, states(e.g. `hasLeadership`, `mutex`) could be changed by dated task.
* Interruption for `LeaderSelectorListener.takeLeadership` is a must regardless of leadership. Otherwise, `LeaderSelectorListener.takeLeadership` could run with dated leadership and no future interruption.
* Without `autoRequeue`, `requeue` is hard to use.
   * An attempt could be cancelled due to session state change.
   * `requeue` after `LeaderSelector.interruptLeadership` might be a noop.

For your questions:
1. Yes. It may be better to enforce only one `LeaderSelectorListener.takeLeadership` call simultaneously. But that could make this pr not concentrated.
2. Yes, but `LeaderSelector.interruptLeadership` is not enough as it could introduce dated leadership.

I think I messed up things here, it might be better to make this pr more concentrated on CURATOR-518 and leaves other concerns(concurrency of `LeaderSelector.doWork` and `requeue` usage)  in followups.

tisonkun: I agree that we can focus on the fix here. Hard-code timeout seems brittle also. We can give another round for improvement.

kezhuw:
[CURATOR-518](https://issues.apache.org/jira/browse/CURATOR-518) covers two things I think:
1. `interruptLeadership` did not test `hasLeadership` as what its javadoc says "if this instance has leadership".
2. `interruptLeadership` could break `autoRequeue`.

There are different since `interruptLeadership` has two different usages:
1. Public API: we should either change its javadoc or its implementation.
2. Cancel election regardless of leadership in case of session state change. Currently, auto-requeue is done by submit next election in its own task. But `FutureTask.cancel` chould cancel election task before it got a chance to run. Apparently, this is a bug in implementation.

Any thoughts ? @eolivelli @tisonkun

curator-recipes/src/main/java/org/apache/curator/framework/recipes/leader/LeaderSelector.java
kezhuw: It is caused by my fix in `cancelElection`. Old code won't cause simultaneous elections, so didn't need such wait.

kezhuw:
@eolivelli @tisonkun Would you mind take another look ? I reworked the fix:

1. Fix implementation of `interruptLeadership` to interrupt only if `hasLeadership`.
2. Use `ourThread` to cancel election, this avoids immature future task cancellation which break election loops. With help from synchronized `taskStarted`, `taskDone`, we can certainly know next election after `cancelElection` will see cancellation cause(e.g. session expired).

tisonkun (approved):
LGTM. @eolivelli may you do another review?

tisonkun:
Merging...

Thanks for your contribution @kezhuw!

PR #448, opened 2023-03-03
CURATOR-662: Export getLockPath for locks from InterProcessReadWriteLock

kezhuw:
With protected `getLockPath`, clients have to inherit them to use `getLockPath`, it is cumbersome and nonsense.


kezhuw:
@eolivelli @cammckenzie Does this make sense ? The protected version was introduced in #394.

PR #453, opened 2023-03-31
CURATOR-665: Convey mkdirs exception to create backgorund

tisonkun:
See https://issues.apache.org/jira/browse/CURATOR-665 for details.

### DIFF
--- a/curator-client/src/main/java/org/apache/curator/utils/ZKPaths.java
+++ b/curator-client/src/main/java/org/apache/curator/utils/ZKPaths.java
@@ -23,15 +23,11 @@ import com.google.common.base.Splitter;
 import com.google.common.collect.Lists;
 import java.util.Collections;
 import java.util.List;
-import java.util.concurrent.BlockingQueue;
-import java.util.concurrent.LinkedBlockingQueue;
-import org.apache.zookeeper.AsyncCallback;
 import org.apache.zookeeper.CreateMode;
 import org.apache.zookeeper.KeeperException;
 import org.apache.zookeeper.ZooDefs;
 import org.apache.zookeeper.ZooKeeper;
 import org.apache.zookeeper.data.ACL;
-import org.apache.zookeeper.data.Stat;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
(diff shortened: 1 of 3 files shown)


curator-client/src/main/java/org/apache/curator/utils/ZKPaths.java
```diff
@@ -333,26 +341,49 @@ public static void mkdirs(ZooKeeper zookeeper, String path, boolean makeLastNode
...
                 }
-                catch ( KeeperException.NodeExistsException e )
+                if ( acl == null )
+                {
+                    acl = ZooDefs.Ids.OPEN_ACL_UNSAFE;
+                }
+
+                CompletableFuture<KeeperException> f = new CompletableFuture<>();
+
+                zookeeper.create(subPath, new byte[0], acl, getCreateMode(asContainers), new AsyncCallback.Create2Callback()
```
tisonkun, line 360: The callback-completable-future-mixed style is not quite elegant, but this is the best solution I find now.

curator-framework/src/main/java/org/apache/curator/framework/imps/CreateBuilderImpl.java
```diff
@@ -770,14 +790,24 @@ static <T> void backgroundCreateParentsThenNode(final CuratorFrameworkImpl clien
...
                 try
                 {
-                    ZKPaths.mkdirs(client.getZooKeeper(), path, false, aclProvider, createParentsAsContainers);
+                    ZKPaths.mkdirs(client.getZooKeeper(), path, false, aclProvider, createParentsAsContainers, callback);
                 }
                 catch ( KeeperException e )
                 {
                     if ( !client.getZookeeperClient().getRetryPolicy().allowRetry(e) )
                     {
+                        client.processBackgroundOperation(mainOperationAndData, event.get());
```
kezhuw, line 810: Sounds like all we need in exceptional case is `rc` and `path`. `KeeperException` already covers these two `KeeperException::code` and `KeeperException::getPath`. So I think we could construct `CuratorEvent` without resorting to a callback.

tisonkun: @kezhuw Thanks for your insight!

I checked that on exception, `stat` should be `null`, `ctx` we passed is `null`, and since we don't create sequential node here, `path == name`. So you're right that I can tidy the code a lot.

tisonkun: Updated.

kezhuw: @tisonkun On exception, `name` should be `null` as no node is created. But this won't affect final result as `KeeperException` needs only `CuratorEvent::getResultCode` and `CuratorEvent::getPath` to reconstruct itself.

kezhuw (approved):
LGTM

kezhuw:
Nit: the commit title sounds outdated now.

tisonkun:
@kezhuw You're right. Updated.

eolivelli (approved):
+1

tisonkun:
Thanks for your review! Merging...

PR #454, opened 2023-04-03
CURATOR-666: Fix twice unfixForNamespace in background create

kezhuw:
Previously, `unfixForNamespace` was called twice in background create. This will accidentally unfix node path if it happens to contain namespace in its prefix. Say, creating "/zoo/a" in namespace "/zoo" will get path "/a".

### DIFF
--- a/curator-framework/src/main/java/org/apache/curator/framework/imps/CreateBuilderImpl.java
+++ b/curator-framework/src/main/java/org/apache/curator/framework/imps/CreateBuilderImpl.java
@@ -797,9 +797,8 @@ public class CreateBuilderImpl implements CreateBuilder, CreateBuilder2, Backgro
                 {
                     if ( !client.getZookeeperClient().getRetryPolicy().allowRetry(e) )
                     {
-                        final CuratorEvent event = makeCuratorEvent(client, e.code().intValue(), e.getPath(), null, null, null);
-                        client.processBackgroundOperation(mainOperationAndData, event);
-                        throw e;
+                        sendBackgroundResponse(client, e.code().intValue(), e.getPath(), null, null, null, mainOperationAndData);
+                        return;
                     }
                     // otherwise safe to ignore as it will get retried
                 }
@@ -895,12 +894,13 @@ public class CreateBuilderImpl implements CreateBuilder, CreateBuilder2, Backgro
 
     private void sendBackgroundResponse(int rc, String path, Object ctx, String name, Stat stat, OperationAndData<PathAndBytes> operationAndData)
     {
-        client.processBackgroundOperation(operationAndData, makeCuratorEvent(client, rc, path, ctx, name, stat));
+        sendBackgroundResponse(client, rc, path, ctx, name, stat, operationAndData);
     }
 
-    private static CuratorEvent makeCuratorEvent(CuratorFrameworkImpl client, int rc, String path, Object ctx, String name, Stat stat)
+    private static <T> void sendBackgroundResponse(CuratorFrameworkImpl client, int rc, String path, Object ctx, String name, Stat stat, OperationAndData<T> operationAndData)
     {
-        return new CuratorEventImpl(client, CuratorEventType.CREATE, rc, path, name, ctx, stat, null, null, null, null, null);
+        CuratorEvent event = new CuratorEventImpl(client, CuratorEventType.CREATE, rc, path, name, ctx, stat, null, null, null, null, null);
+        client.processBackgroundOperation(operationAndData, event);
     }
 
     private ACLCreateModePathAndBytesable<String> asACLCreateModePathAndBytesable()
--- a/curator-framework/src/main/java/org/apache/curator/framework/imps/CreateBuilderImpl.java
+++ b/curator-framework/src/main/java/org/apache/curator/framework/imps/CreateBuilderImpl.java
@@ -900,8 +900,6 @@ public class CreateBuilderImpl implements CreateBuilder, CreateBuilder2, Backgro
 
     private static CuratorEvent makeCuratorEvent(CuratorFrameworkImpl client, int rc, String path, Object ctx, String name, Stat stat)
     {
-        path = client.unfixForNamespace(path);
-        name = client.unfixForNamespace(name);
         return new CuratorEventImpl(client, CuratorEventType.CREATE, rc, path, name, ctx, stat, null, null, null, null, null);
     }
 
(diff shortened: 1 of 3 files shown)


curator-framework/src/main/java/org/apache/curator/framework/imps/CreateBuilderImpl.java
```diff
@@ -797,7 +797,7 @@ public void performBackgroundOperation(OperationAndData<T> dummy) throws Excepti
                 {
                     if ( !client.getZookeeperClient().getRetryPolicy().allowRetry(e) )
                     {
-                        final CuratorEvent event = makeCuratorEvent(client, e.code().intValue(), e.getPath(), null, e.getPath(), null);
```
kezhuw, line 800: I overlooked the callback case in reviewing #453, `CuratorEvent::getName` is useless in reconstruction of `KeeperException`. But in callback, it should be `null` as no node is created. This is also consistent with ZooKeeper world's callback. This is currently pushed as hotfix commit. I think it might not worth to create separated pr for this. Any thoughts ? @tisonkun @eolivelli

curator-framework/src/main/java/org/apache/curator/framework/imps/CuratorEventImpl.java
```diff
@@ -133,7 +133,7 @@ public String toString()
         this.resultCode = resultCode;
         this.opResults = (opResults != null) ? ImmutableList.copyOf(opResults) : null;
         this.path = client.unfixForNamespace(path);
-        this.name = name;
+        this.name = client.unfixForNamespace(name);
```
kezhuw, line 136: This should be safe as `unfixForNamespace` is tolerated to `null` input.

curator-framework/src/main/java/org/apache/curator/framework/imps/CreateBuilderImpl.java
```diff
@@ -900,8 +900,6 @@ private void sendBackgroundResponse(int rc, String path, Object ctx, String name
 
     private static CuratorEvent makeCuratorEvent(CuratorFrameworkImpl client, int rc, String path, Object ctx, String name, Stat stat)
     {
```
tisonkun, line 902: Let's inline this method since it's now only a delegate.

curator-framework/src/main/java/org/apache/curator/framework/imps/CreateBuilderImpl.java
tisonkun: I'm fine with this change.

tisonkun (approved):
Verify the behavior as expected.

curator-framework/src/main/java/org/apache/curator/framework/imps/CreateBuilderImpl.java
kezhuw: I have introduced a static version of `sendBackgroundResponse`, so that all response are go through this method finally. I saw no "beauty" code path to construct `CuratorEvent`, too many parameters ...

tisonkun (approved):

curator-framework/src/main/java/org/apache/curator/framework/imps/CreateBuilderImpl.java
kezhuw: I changed the `throw e` to `return` in latest push. I think it is unnecessary to report or log such exception as un-handled now.

tisonkun: I'd prefer keep it as is to minimize the changeset. Now the exception will be handled by `handleBackgroundOperationException` and keep the current logic doesn't hurt user experience.

tisonkun (changes requested):

curator-framework/src/main/java/org/apache/curator/framework/imps/CreateBuilderImpl.java
tisonkun: At least they are two different things. Let's focus them one by one if you insist that such a change benefits.

kezhuw: @tisonkun You are right. I have reverted such a change to keep it as is. The change will break `TestLeaderAcls::testAclErrorWithLeader` which depends on this un-handled behavior.

tisonkun (approved):
Thanks for your update! I forget to change the review to approval >_<

Randgalt:
This looks reasonable but I recall this is a very difficult are to play around with. The tests pass, so that's good. It would be nice to find a few users who can validate this against real world usages.

eolivelli (approved):
+1

I would prefer to see @Randgalt or @cammckenzie opinions before committing.
I haven't worked on this part of Curator and I don't have much time to carefully review all the cases

tisonkun:
> very difficult are to play around with. The tests pass, so that's good

@Randgalt The behavior changes for "/zoo/zoo/a" with a namespace "/zoo":

* Previously, the result is unfixed as "/a";
* Now, it's unfixed as "/zoo/a".

Previous tests do not cover this case, actually.

tisonkun:
Another change is we now trim `name` when constructing `CuratorEventImpl`, while I review all the references to `getName` from an async Create callback, and the `name` is previously trimmed in the `sendBackground...` method. Other async callbacks may behavior differently but it seems few users will use `name` there.

This patch implements what we should originally do IMO. Due to the error state is rare so we don't receive reports before.

kezhuw:
@tison Only `create` has `name`. In zookeeper [async callbacks](https://github.com/apache/zookeeper/blob/3eda8a3a00ef3230d6c16f6ddf9201a3c0f5ca07/zookeeper-server/src/main/java/org/apache/zookeeper/AsyncCallback.java#L209), only `Create2Callback` and `StringCallback`, which both are callback for `ZooKeeper::create`, have `name` parameter. And `unfixForNamespace` is tolerated to `null` input, so I moved it to where we trim `path`. This enable caller to treat them same.

@Randgalt No sure whether it is related, there is [CURATOR-379](https://issues.apache.org/jira/browse/CURATOR-379)(#194), but its test does not convince me either. Though, it is possible for namespace "/zoo" and `getConfig`'s path "/zookeeper/config", but it is another story [ZOOKEEPER-4601](https://issues.apache.org/jira/browse/ZOOKEEPER-4601), [CURATOR-667](https://issues.apache.org/jira/browse/CURATOR-667).

PR #459, opened 2023-04-12
CURATOR-669. Integrate with ge.apache.org Gradle Enterprise server

tisonkun:
https://issues.apache.org/jira/browse/CURATOR-669

See -

* https://github.com/apache/pulsar/pull/19133
* https://github.com/apache/pulsar/pull/19194
* https://the-asf.slack.com/archives/CBX4TSBQ8/p1672826514725099

I'd prefer to use this method for analyzing our flaky tests and time-consuming tests for further improving.

### DIFF
--- a/.mvn/ge-extensions.xml
+++ b/.mvn/ge-extensions.xml
@@ -1,0 +1,34 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!--
+
+    Licensed to the Apache Software Foundation (ASF) under one
+    or more contributor license agreements.  See the NOTICE file
+    distributed with this work for additional information
+    regarding copyright ownership.  The ASF licenses this file
+    to you under the Apache License, Version 2.0 (the
+    "License"); you may not use this file except in compliance
+    with the License.  You may obtain a copy of the License at
+
+      http://www.apache.org/licenses/LICENSE-2.0
+
+    Unless required by applicable law or agreed to in writing,
+    software distributed under the License is distributed on an
+    "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+    KIND, either express or implied.  See the License for the
+    specific language governing permissions and limitations
+    under the License.
+
+-->
+<extensions xmlns="http://maven.apache.org/EXTENSIONS/1.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+            xsi:schemaLocation="http://maven.apache.org/EXTENSIONS/1.0.0 http://maven.apache.org/xsd/core-extensions-1.0.0.xsd">
+    <extension>
+        <groupId>com.gradle</groupId>
+        <artifactId>gradle-enterprise-maven-extension</artifactId>
+        <version>1.16.3</version>
+    </extension>
+    <extension>
+        <groupId>com.gradle</groupId>
+        <artifactId>common-custom-user-data-maven-extension</artifactId>
+        <version>1.11.1</version>
+    </extension>
+</extensions>
--- a/.mvn/gradle-enterprise-custom-user-data.groovy
+++ b/.mvn/gradle-enterprise-custom-user-data.groovy
@@ -1,0 +1,46 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+// Add Maven command line arguments
+def mavenCommand = ''
+
+if (System.env.MAVEN_CMD_LINE_ARGS) {
+    mavenCommand = "mvn ${System.env.MAVEN_CMD_LINE_ARGS}".toString()
+    buildScan.value('Maven command line', mavenCommand)
+}
+
+if (System.env.GITHUB_ACTIONS) {
+    if (session.goals.size() == 1 && session.goals[0] == 'initialize') {
+        // omit publishing build scan when the only goal is "initialize"
+        buildScan.publishOnDemand()
+    } else {
+        def jobName = System.env.GITHUB_JOB
+        buildScan.value('GitHub Actions Job name', jobName)
+        buildScan.value('GitHub Actions Event name', System.env.GITHUB_EVENT_NAME)
+        buildScan.value('GitHub Ref name', System.env.GITHUB_REF_NAME)
+        buildScan.value('GitHub Actor', System.env.GITHUB_ACTOR)
+        buildScan.link('GitHub Repository', "https://github.com/" + System.env.GITHUB_REPOSITORY)
+        buildScan.link('GitHub Commit', "https://github.com/" + System.env.GITHUB_REPOSITORY + "/commits/" + System.env.GITHUB_SHA)
+        buildScan.buildScanPublished {  publishedBuildScan ->
+            new File(System.env.GITHUB_STEP_SUMMARY).withWriterAppend { out ->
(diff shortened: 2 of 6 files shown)


tisonkun:
Result - https://ge.apache.org/scans/tests?search.rootProjectNames=Apache%20Curator&search.timeZoneId=Asia/Shanghai&tests.sortField=MEAN_DURATION

We can review the flaky tests and test time cost for further improving.

cc @eolivelli @kezhuw @Randgalt @cammckenzie

.mvn/ge-extensions.xml
```diff
@@ -0,0 +1,34 @@
...
+    specific language governing permissions and limitations
+    under the License.
+
+-->
+<extensions xmlns="http://maven.apache.org/EXTENSIONS/1.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+            xsi:schemaLocation="http://maven.apache.org/EXTENSIONS/1.0.0 http://maven.apache.org/xsd/core-extensions-1.0.0.xsd">
+    <extension>
+        <groupId>com.gradle</groupId>
+        <artifactId>gradle-enterprise-maven-extension</artifactId>
+        <version>1.16.3</version>
```
clayburn, line 27: The most recent version of this extension is 1.16.6. Once https://ge.apache.org is update to 2023.1, you may update this to 1.17

.mvn/gradle-enterprise.xml
```diff
@@ -0,0 +1,49 @@
...
+        <allowUntrusted>false</allowUntrusted>
+    </server>
+    <buildScan>
+        <capture>
+            <goalInputFiles>true</goalInputFiles>
+            <buildLogging>true</buildLogging>
+            <testLogging>true</testLogging>
+        </capture>
+        <backgroundBuildScanUpload>#{isFalse(env['GITHUB_ACTIONS'])}</backgroundBuildScanUpload>
+        <publish>ALWAYS</publish>
```
clayburn, line 36: I would recommend adding `publishIfAuthenticated>true</publishIfAuthenticated>`. It is a hidden option, but will give unauthenticated users a better user experience, given that not every builder of this open source project will be an authenticated user to https://ge.apache.org.

.mvn/gradle-enterprise-custom-user-data.groovy
```diff
@@ -0,0 +1,46 @@
...
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+// Add Maven command line arguments
+def mavenCommand = ''
+
+if (System.env.MAVEN_CMD_LINE_ARGS) {
+    mavenCommand = "mvn ${System.env.MAVEN_CMD_LINE_ARGS}".toString()
+    buildScan.value('Maven command line', mavenCommand)
```
clayburn, line 25: I would be very careful capturing the entire Maven command line. You could imagine scenarios where a user passes in a secret as a system property, or possibly even typos that include secrets. Gradle Enterprise doesn't capture certain environment details for this reason.

Perhaps it would make more sense to capture specific details you may be interested in? For example, we already add a custom value for the `skipTests` setting. We also have samples to [capture profiles](https://github.com/gradle/gradle-enterprise-build-config-samples/blob/main/build-data-capturing-maven-samples/capture-profiles/maven-profiles.groovy) as tags. This may actually make it easier to queries that combine the different values, rather than trying to construct queries that are considering the command line as a whole. It may also be more accurate, as capturing the value of configured properties would not _only_ capture values passed in at the command line.

.mvn/ge-extensions.xml
clayburn, line 1: I would highly recommend renaming this to `extensions.xml` and removing the `Configure Gradle Enterprise integration` from `ci.yml`. Consider:

* This makes it much easier for a local developer to publish a local build scan without having to move the `ge-extensions.xml` file, whether to https://ge.apache.org or to https://scans.gradle.com.
* This simplifies your CI workflow, requiring no changes to the your `ci` workflow.
* This gives all builds access to the local build cache.
* While we strongly recommend enabling build scans for all builds, a better way to only publish build scans on CI would utilize [criteria-based publishing](https://docs.gradle.com/enterprise/maven-extension/#publishing_based_on_criteria).
* Applying the extension to all builds is quite safe and well tested, even if the build does not publish a build scan.

clayburn (commented):
Hi, I am a Solutions engineer with Gradle assisting with the Gradle Enterprise rollout. It is exciting to see your interest in sending build scans for the Curator build to https://ge.apache.org! I've added some best practice recommendations to the pull request.

If you have any questions regarding Gradle Enterprise, please feel free to reach out and we would be happy to help you. I see your primary use case is the Test analytics, but after this project is connected, we will run some experiment builds to check how well your build utilizes the build cache and submit back any fixes we discover.

.mvn/gradle-enterprise.xml
tisonkun: Thank you! It actually exists also in the template I copied from. But it seems this option is omitted from the schema..

Let me add it now.

.mvn/ge-extensions.xml
tisonkun: > This makes it much easier for a local developer to publish a local build scan without having to move the ge-extensions.xml file, whether to https://ge.apache.org/ or to https://scans.gradle.com/.

It seems local environment doesn't have the necessary TOKEN.

.mvn/gradle-enterprise.xml
clayburn: That is correct, it is not currently an option in the schema. It is an option that is only recommended for use in open source environments where it is not desirable for every single developer that clones the project to be able to authenticate and publish build scans.

You can see the [SCIMple project](https://github.com/apache/directory-scimple/blob/develop/.mvn/gradle-enterprise.xml#L37) for another example of usage of this option (as well as configuration of Gradle Enterprise generally).

.mvn/ge-extensions.xml
clayburn: https://ge.apache.org is on Gradle Enterprise 2023.1.1, so you may now set this to 1.17

clayburn:
Just wanted to follow up on this. I know you originally looked at the Gradle Enterprise configruation in the Pulsar project. We've filed a [PR](https://github.com/apache/pulsar/pull/20187) there with a couple of changes over there to follow some new recommendations that come with Gradle Enterprise Maven Extension 1.17.

tisonkun:
@clayburn Thanks for your updates. I'll catch up changes in weeks :)

.mvn/ge-extensions.xml
tisonkun: Updated. May I ask why gradle-enterprise-maven-extension use a `X.Y` pattern instead of `X.Y.Z` now?

.mvn/gradle-enterprise-custom-user-data.groovy
tisonkun: I'll leave it as is for now. If you have better config suggestions, feel free to patch the script based on this branch or after this patch gets merged.

.mvn/ge-extensions.xml
tisonkun: Sounds good to me. Let me update it.

tisonkun:
@clayburn Updated at a298df5c9296421766c74e7d4ac557aa710a0620. Please give another look :D

.mvn/ge-extensions.xml
clayburn: The convention across Gradle products is to omit the patch version when equal to zero. So the next version may be either 1.18 or 1.17.1

clayburn (approved):
LGTM!

kezhuw (approved):
Generally looks good to me. I think @clayburn raised a good point about the capture of maven command line especially after enabling local scans. We can enable limited maven command line capture for local scans as what @clayburn suggested in following up pr.

tisonkun:
OK. I'm trying to redact mvn command.

.mvn/gradle-enterprise-custom-user-data.groovy
clayburn: > OK. I'm trying to redact mvn command.

@tisonkun - should just be a matter of removing this `if (System.env.MAVEN_CMD_LINE_ARGS) { ... }` block

tisonkun:
After a closer look, I'm going to remove the if block for now.

clayburn:
This test is failing on the Java 8 test build: https://ge.apache.org/s/ovviiad2aixry/tests/overview?outcome=failed

I don't know much about your test suite, but I would assume this one is not related to the changes in this PR?

tisonkun:
@clayburn Yep. I have rerun the tests and such flaky tests are part of issue we want to located and resolve with the help of Gradle Enterprise integration.

kezhuw:
Reported to https://issues.apache.org/jira/browse/CURATOR-671. I saw this also in #460(parallel with this pr). cc @clayburn @tisonkun

tisonkun:
Pending to merge...

PR #460, opened 2023-05-05
CURATOR-593: Append chroot for EnsembleProvider::setConnectionString in EnsembleTracker

kezhuw:
Curator uses `EnsembleProvider::getConnectionString` as connection string to `ZooKeeper`. `EnsembleTracker` subscribes to config node to construct up to date connection string for `EnsembleProvider`. This is great.

But, currently, `EnsembleTracker` omits chroot part of connection string which could cause curator locating at ZooKeeper root after reconnection. This could damage clients' data hierarchy in unexpected manner.

### DIFF
--- a/curator-framework/src/main/java/org/apache/curator/framework/imps/EnsembleTracker.java
+++ b/curator-framework/src/main/java/org/apache/curator/framework/imps/EnsembleTracker.java
@@ -209,10 +209,17 @@ public class EnsembleTracker implements Closeable, CuratorWatcher
         if (!properties.isEmpty())
         {
             QuorumMaj newConfig = new QuorumMaj(properties);
-            String connectionString = configToConnectionString(newConfig);
-            if (connectionString.trim().length() > 0)
+            String connectionString = configToConnectionString(newConfig).trim();
+            if (!connectionString.isEmpty())
             {
                 currentConfig.set(newConfig);
+                String oldConnectionString = ensembleProvider.getConnectionString();
+                int i = oldConnectionString.indexOf('/');
+                if (i >= 0)
+                {
+                    String chroot = oldConnectionString.substring(i);
+                    connectionString += chroot;
+                }
                 ensembleProvider.setConnectionString(connectionString);
             }
             else
--- a/curator-framework/src/test/java/org/apache/curator/framework/imps/TestReconfiguration.java
+++ b/curator-framework/src/test/java/org/apache/curator/framework/imps/TestReconfiguration.java
@@ -51,6 +51,7 @@ import java.io.IOException;
 import java.net.InetAddress;
 import java.net.InetSocketAddress;
 import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.Collection;
 import java.util.Collections;
 import java.util.Iterator;
@@ -60,6 +61,7 @@ import java.util.concurrent.CountDownLatch;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicReference;
 
+import static org.junit.jupiter.api.Assertions.assertArrayEquals;
 import static org.junit.jupiter.api.Assertions.assertEquals;
 import static org.junit.jupiter.api.Assertions.assertNotEquals;
 import static org.junit.jupiter.api.Assertions.assertNotNull;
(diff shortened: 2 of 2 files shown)


curator-framework/src/test/java/org/apache/curator/framework/imps/TestReconfiguration.java
```diff
@@ -436,6 +439,51 @@ public void testNewMembers() throws Exception
...
+            assertConfig(newConfig, newInstances);
+
+            assertTrue(timing.awaitLatch(ensembleLatch));
+            String connectString = EnsembleTracker.configToConnectionString(newConfig) + chroot;
+            assertEquals(connectString, ensembleProvider.getConnectionString());
+
+            client.getZookeeperClient().reset();
+            client.sync().forPath("/");
+            byte[] data = client.getData().forPath("/");
+            assertArrayEquals("deadbeef".getBytes(), data, () -> "expected \"deedbeef\", got data: " + Arrays.toString(data));
```
tisonkun, line 483:
```suggestion
            assertThat(new String(data)).isEqualTo("deadbeef");
```

Simplify the assertion

curator-framework/src/main/java/org/apache/curator/framework/imps/EnsembleTracker.java
```diff
@@ -209,10 +209,17 @@ private void processConfigData(byte[] data) throws Exception
...
             QuorumMaj newConfig = new QuorumMaj(properties);
-            String connectionString = configToConnectionString(newConfig);
-            if (connectionString.trim().length() > 0)
+            String connectionString = configToConnectionString(newConfig).trim();
+            if (!connectionString.isEmpty())
             {
                 currentConfig.set(newConfig);
+                String oldConnectionString = ensembleProvider.getConnectionString();
+                int i = oldConnectionString.indexOf('/');
+                if (i >= 0)
```
tisonkun, line 218: Perhap `== 0`? I'm not sure what if `i > 0` the string constructing.

curator-framework/src/test/java/org/apache/curator/framework/imps/TestReconfiguration.java
```diff
@@ -558,6 +606,7 @@ private CuratorFramework newClient(String connectionString) {
     private CuratorFramework newClient(String connectionString, boolean withEnsembleProvider)
     {
         final AtomicReference<String> connectString = new AtomicReference<>(connectionString);
+        ensembleLatch = new CountDownLatch(1);
```
tisonkun, line 609: I suggest to add a `newClient(String connectionString, boolean withEnsembleProvider, CountDownLatch)` overload to avoid using field - this is somehow global variable that is brittle.

tisonkun (commented):
Thanks for taking care of this issue. One question for EnsembleTracker. Other comments are about style.

curator-framework/src/main/java/org/apache/curator/framework/imps/EnsembleTracker.java
kezhuw: Hmm, we are extracting `chroot` from connection string. Everything last "/" (including "/" itself") are `chroot`.

The `if` part is similar to zookeeper.

https://github.com/apache/zookeeper/blob/a64dbf5b06ca1a73dc2ad6c7d31e679e312082aa/zookeeper-server/src/main/java/org/apache/zookeeper/client/ConnectStringParser.java#L55-L67

curator-framework/src/test/java/org/apache/curator/framework/imps/TestReconfiguration.java
kezhuw: Fixed in fixup commit.

curator-framework/src/test/java/org/apache/curator/framework/imps/TestReconfiguration.java
kezhuw: Added `newClient(String connectionString, CountDownLatch ensembleLatch)`. The `withEnsembleProvider` parameter will be `true` if `ensembleLatch != null` in this overload.

tisonkun (approved):

curator-framework/src/main/java/org/apache/curator/framework/imps/EnsembleTracker.java
```diff
@@ -209,10 +209,17 @@ private void processConfigData(byte[] data) throws Exception
         if (!properties.isEmpty())
         {
             QuorumMaj newConfig = new QuorumMaj(properties);
-            String connectionString = configToConnectionString(newConfig);
-            if (connectionString.trim().length() > 0)
+            String connectionString = configToConnectionString(newConfig).trim();
```
tisonkun, line 212: Is it possible that the `connectionString` also has a `chroot` part and we append the old chroot last which is disordered?

kezhuw: No. This `connectionString` is constructed from `QuorumMaj` which is a server side concept and has no notion of `chroot`.

PR #465, opened 2023-06-04
CURATOR-675: Fix AsyncCuratorFramework's UnhandledErrorListener hang AsyncStage

kezhuw:
`AsyncCuratorFramework` binds its `UnhandledErrorListener` to operation level which could prevent `AsyncStage` from completion if exception is swallowed by that listener.

In any cases, a framework level error listener should not prevent an operation future from completion.


tisonkun (approved):

tisonkun:
Good catch!

PR #464, opened 2023-05-30
CURATOR-673: Complete BackgroundCallback if CuratorFramework got closed

kezhuw:
Previously, if curator framework is closed while there are background operations, these operations are dropped with no dedicated events.

It is hard for blocking clients to stop waiting after framework closed. `CuratorListener` seems like a circumvent. But apparently, it is hard and I don't think this is by design.

This pr introduces a `clockLock` to synchronize framework closing and operation queuing. So we are sure that no operations could be queued after framework marked as closing.

PR #467, opened 2023-06-12
CURATOR-677: Complete BackgroundCallback if sub operation failed or cancelled

kezhuw:
Currently, some background operations use auxiliary sub operations to
complete task in case of primary conditions are not satisfied. But most
of these sub operations count only success path, so they will hang
`BackgroundCallback` if they are failed or cancelled due to framework
closed.

This is the leftover of [CURATOR-673][](https://github.com/apache/curator/pull/464).

[CURATOR-673]: https://issues.apache.org/jira/browse/CURATOR-673


kezhuw (commented):
I deliberately split this pr into two commits, first for test case and second for code fix. Most newly added tests could not even resist one run in "repeat until failure" mode. All new tests resist 100 runs without failure in local.

tisonkun (approved):
Thank you!

+1 for merging.

PR #470, opened 2023-06-21
CURATOR-681: Fix flaky TestInterProcessMutex.testReentrantSingleLock

kezhuw:
Placing `Thread.sleep(550)` after `mutex.release()` and before `hasLock.set(false)` will fail this test more frequently.


tisonkun (approved):
Good catch!

PR #473, opened 2023-08-07
CURATOR-457: Custom service discovery path

kezhuw:
I append one commit bd7c333. See also #256.


curator-x-discovery/src/test/java/org/apache/curator/x/discovery/details/TestServiceDiscovery.java
```diff
@@ -261,6 +262,49 @@ public void testBasic() throws Exception {
...
+    private static class CustomPathConstructor implements DiscoveryPathConstructor {
+        @Override
+        public String getBasePath() {
+            return "/test";
+        }
+
+        @Override
+        public String getPathForInstances(String serviceName) {
+            return String.format("/test/%s/instances", serviceName);
+        }
```
kezhuw, line 274: In case of old version, there is one more method `getPathForInstance`(with no trailing `s` character). In this concrete test case, `getPathForInstance` can't return anything other than `String.format("/test/%s/instances/%s", serviceName, instanceId)`, otherwise it breaks `queryForInstances`.

`queryForInstances` relies on `getPathForInstances` which states "Return the (parent) path where all instances of the service registered".

tisonkun (approved):

PR #474, opened 2023-08-08
CURATOR-667: Fix abnormal event config path from getConfig

kezhuw:
`getConfig` will read data from path "/zookeeper/config" which belongs to no namespace nor chroot in ZooKeeper side. Path in background event or watch event should stick to "/zookeeper/config".

See also:
* https://issues.apache.org/jira/browse/ZOOKEEPER-4601
* https://lists.apache.org/thread/2tsg1hcopl80zot12tqrynrbg2h792jf


eolivelli (approved):
Lgtm

PR #475, opened 2023-08-08
CURATOR-626: Postpone null namespace facade creation to avoid partial initialization

kezhuw:
In construction of `CuratorFrameworkImpl`, it constructs a sub class `NamespaceFacade` for commonly used `nullNamespace`. Due to this recursive construction, `nullNamespace` has uninitialized fields which cause NPT in certain cases.

This commit solves this by postponing `nullNamespace` creation.


eolivelli (approved):
Lgtm

PR #476, opened 2023-08-15
CURATOR-685. Cleanup docs and setups for the old site deployment

tisonkun:
This refers to CURATOR-685.


curator-test/src/site/confluence/index.confluence
```diff
@@ -1,10 +0,0 @@
-h1. Testing Utilities
```
kezhuw, line 1: This is merged  with "utilities.confluence" to https://curator.apache.org/docs/utilities. That is good! It is a subset of that page.

src/site/confluence/compatibility.confluence
```diff
@@ -1,10 +0,0 @@
-h1. API Compatibility
```
kezhuw, line 1: These compatibility reports are left over. I think @tisonkun already has solution for javadoc. This sounds similar. It is a a good to have compatibility reports , but without ci in old site, these were dated anyway.

kezhuw (approved):
Great! Thank you for your hard work!

I guess I am 99.99 percent sure we are migrate all docs except generated one 😄 .

src/site/confluence/compatibility.confluence
tisonkun: Yeah. I think these reports should actually report for every PR and guard that we're in a healthy state instead of showing on the website - I highly suspect who will read these pages.

tisonkun:
Merging...

PR #477, opened 2023-08-16
CURATOR-686: Enable GitHub autolink jira issues

kezhuw:


tisonkun:
Oops. I add it directly as I saw ZOOKEEPER's similar PR. Sorry.

kezhuw:
I saw the effect, it is nice! I won't have to type linking jira address by myself from now on.

I guess we both are excited about these tiny productivity improvement.

tisonkun:
Exactly 🤣

PR #479, opened 2023-09-13
Curator 690. curator cache fails to load the cache if there are more than 64 k child z nodes

Ryan0751:
This PR addresses the 64K child ZNode scaling limitation of CuratorCache.

This is done by utilizing tiered Phasers to handle monitoring of additional outstanding operations during the initialization phase of the cache.

As each child Phaser reaches the 0xffff capacity of a single Phaser, an additional child Phaser is created and registered with the root Phaser.

After initialization completes, the "currentChildPhaser" member variable is set to rootPhaser such that all child Phasers (which will never be used again) can be GC'd.

A unit test has been provided which fails without this fix, but succeeds with.

### DIFF
--- a/curator-recipes/src/main/java/org/apache/curator/framework/recipes/cache/CuratorCacheImpl.java
+++ b/curator-recipes/src/main/java/org/apache/curator/framework/recipes/cache/CuratorCacheImpl.java
@@ -19,9 +19,19 @@
 
 package org.apache.curator.framework.recipes.cache;
 
+import static org.apache.curator.framework.recipes.cache.CuratorCacheListener.Type.*;
+import static org.apache.zookeeper.KeeperException.Code.NONODE;
+import static org.apache.zookeeper.KeeperException.Code.OK;
 import com.google.common.annotations.VisibleForTesting;
 import com.google.common.base.Preconditions;
 import com.google.common.collect.Sets;
+import java.util.Collections;
+import java.util.Optional;
+import java.util.Set;
+import java.util.concurrent.Phaser;
+import java.util.concurrent.atomic.AtomicReference;
+import java.util.function.Consumer;
+import java.util.stream.Stream;
 import org.apache.curator.framework.CuratorFramework;
 import org.apache.curator.framework.api.BackgroundCallback;
 import org.apache.curator.framework.api.CuratorEvent;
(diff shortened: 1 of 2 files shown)


curator-recipes/src/main/java/org/apache/curator/framework/recipes/cache/CuratorCacheImpl.java
```diff
@@ -59,14 +59,29 @@ class CuratorCacheImpl implements CuratorCache, CuratorCacheBridge {
...
+            } 
             return true;
         }
     };
 
+    private Phaser currentChildPhaser = new Phaser(rootPhaser);
+
+    private synchronized Phaser getPhaserAndRegister() {
+        if (currentChildPhaser.getRegisteredParties() >= 0xffff) {
+            currentChildPhaser = new Phaser(rootPhaser);
```
tisonkun, line 77:
```suggestion
            currentChildPhaser = new Phaser(currentChildPhaser);
```

?

Ryan0751: Is there a reason to prefer chain of child phasers versus a single layer deep tree off the single parent root phaser?

I suppose a tree has a limit of 64k children, but that should support 172x10^9 ZNodes.

tisonkun: > prefer chain of child phasers versus a single layer deep tree off the single parent root phaser

You are right that the returned currentChildPhaser are referred in the method and called `arriveAndDeregister` accordingly. Then I don't have issue for this implementation.

tisonkun (commented):
@Randgalt once build a well implemented `OutstandingOps` class for doing so without the `0xFFFF` limitation.

I propose to use `Phaser` because if JDK provides a battle tested one, we don't need to make our own.

But now it seems it's also complicated to make it work properly with JDK's Phaser. Then I suggest we simply revert to the original solution.

Thanks for your test case!

I'm going to push a commit directly on this branch https://github.com/apache/curator/commit/1038aa6d03f240fb8a8de9736adc51ccbcc9f5b6.

tisonkun:
Perhaps you don't allow maintainer edit. I make a patch based on yours at https://github.com/apache/curator/pull/480.

On merging your credit will retain.

PR #480, opened 2023-09-14
CURATOR-690. CuratorCache fails to load the cache if there are more than 64k child znodes

tisonkun:
This closes https://github.com/apache/curator/pull/479.

Ref - CURATOR-690


tisonkun:
The revert is trivial and the previous solution is reviewed.

Merging...

kezhuw:
Late, but +1 on this reverting.

Tiered phaser may be more suitable in high contention cases, `OutstandingOps` is simple enough for us to evaluate and the cache loading is one-shot. So we don't have to pay extra complexity coming from tiering.

Thank you for your contribution! @Ryan0751

PR #481, opened 2023-09-15
CURATOR-521. Check LeaderLatch#close(CloseMode) does not accept null

tisonkun:
Ref - CURATOR-521

### DIFF
--- a/curator-recipes/src/main/java/org/apache/curator/framework/recipes/leader/LeaderLatch.java
+++ b/curator-recipes/src/main/java/org/apache/curator/framework/recipes/leader/LeaderLatch.java
@@ -188,6 +188,7 @@ public class LeaderLatch implements Closeable {
      * @throws IOException errors
      */
     public void close(CloseMode closeMode) throws IOException {
+        Preconditions.checkNotNull(closeMode, "closeMode cannot be null");
         internalClose(closeMode, true);
     }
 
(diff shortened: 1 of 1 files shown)


curator-recipes/src/main/java/org/apache/curator/framework/recipes/leader/LeaderLatch.java
```diff
@@ -200,8 +200,6 @@ private synchronized void internalClose(CloseMode closeMode, boolean failOnClose
             }
         }
 
-        Preconditions.checkNotNull(closeMode, "closeMode cannot be null");
```
tisonkun, line 203: Ensure nonnulll and even we can fallback to the default behavior anyway.

curator-recipes/src/main/java/org/apache/curator/framework/recipes/leader/LeaderLatch.java
```diff
@@ -213,18 +211,12 @@ private synchronized void internalClose(CloseMode closeMode, boolean failOnClose
...
-                    listeners.clear();
-                    break;
-                }
-
-                default: {
-                    listeners.clear();
-                    setLeadership(false);
-                    break;
-                }
+            if (closeMode == CloseMode.NOTIFY_LEADER) {
```
kezhuw, line 214: I think it is better to choose `closeMode` in the first place(a.k.a. `close(CloseMode closeMode)`), but not here. I just googled to know that `switch(null)` will cause NPE.

PR #485, opened 2023-10-04
CURATOR-692. Get rid of mockito

tisonkun:
Ref - CURATOR-692


curator-framework/src/test/java/org/apache/curator/framework/imps/TestFramework.java
```diff
@@ -755,19 +754,6 @@ public void testNullNamespace() {
         CloseableUtils.closeQuietly(client);
     }
 
-    @Test
-    public void testNoPartialConstruction() {
```
tisonkun, line 759: The original behavior test is covered by testNullNamespace - https://github.com/apache/curator/pull/475

We don't need this detailed test.

curator-client/src/test/java/org/apache/curator/TestEnsurePath.java
tisonkun, line 1: 1. `EnsurePath` is well-implemented, battle-tested, and deprecated, which means it's stable and doesn't not subject to change.
2. This test checks how many `exist` calls happen - why? Although we can take some effort to port it, I don't think it is worthy.

curator-client/pom.xml
```diff
@@ -62,12 +62,6 @@
             <artifactId>slf4j-api</artifactId>
         </dependency>
 
-        <dependency>
-            <groupId>org.mockito</groupId>
-            <artifactId>mockito-core</artifactId>
-            <scope>test</scope>
-        </dependency>
```
kezhuw, line 69: I am positive to this. I prefer real classes than mocks in most cases if not all. I do believe mock could provide "clean" tests, but the "clean" part could lost effect in refactor comparing to stupid/verbose/explicit hooks. Also, I am fearing about tests full of mocks. Removing this will force us to think about testability. That is good for me.

kezhuw (approved):
+1

Thank you for your efforts!

PR #490, opened 2023-11-22
CURATOR-262. Fix sleepMs overflow

shenjianeng:
fix sleepMs overflow

```java
public class Main {
    private static final Random random = new Random();
    private static int baseSleepTimeMs = Integer.MAX_VALUE;

    public static void main(String[] args) {
        int retryCount = 2;
        long sleepMs1 = baseSleepTimeMs * Math.min(2, random.nextInt(1 << (retryCount + 1)));
        long sleepMs2 = (long) baseSleepTimeMs * Math.min(2, random.nextInt(1 << (retryCount + 1)));
        System.out.println(sleepMs1);
        System.out.println(sleepMs2);
    }
}
```


tisonkun (approved):
Thank you!

tisonkun:
This is actually a follow-up to https://github.com/apache/curator/pull/136 where we introduce the bug by missing integers casting.

tisonkun:
Merging ...

Thanks for your contribution!

PR #494, opened 2023-12-11
[CURATOR-695] Open Telemetry tracing

teabot:
This changes adds an Open Telemetry tracing driver. To make this more effective, it also decomposes `OperationTrace` events so that the start and end of the event can be handled distinctly.

Curator's `OperationTraces` are mapped to OTel `Spans`. When an `OperationTrace` is started (created), a corresponding OTel span is started and made current. The span is closed when the `OperationTrace` is committed. `OperationTrace` members are mapped to OTel span attributes, and result codes are mapped to appropriate OTel representations.

Any Curator `EventTraces` are attributed to the current OTel span.


curator-drivers/LICENSE
```diff
@@ -0,0 +1,202 @@
+
+                                 Apache License
```
eolivelli, line 2: We don't need this file

curator-drivers/NOTICE
```diff
@@ -0,0 +1,5 @@
+Apache Curator
```
eolivelli, line 1: We don't need this file

curator-drivers/open-telemetry/LICENSE
```diff
@@ -0,0 +1,202 @@
+
+                                 Apache License
```
eolivelli, line 2: No need for this file

curator-drivers/open-telemetry/NOTICE
```diff
@@ -0,0 +1,5 @@
+Apache Curator
```
eolivelli, line 1: No need

curator-drivers/open-telemetry/pom.xml
```diff
@@ -0,0 +1,78 @@
...
+
+    <properties>
+        <opentelemetry.version>1.31.0</opentelemetry.version>
+    </properties>
+
+    <dependencies>
+        <dependency>
+            <groupId>org.apache.curator</groupId>
+            <artifactId>curator-client</artifactId>
+            <version>${project.version}</version>
```
eolivelli, line 49: This should be at scope 'provided'

curator-drivers/open-telemetry/target/classes/META-INF/NOTICE
```diff
@@ -0,0 +1,5 @@
+Apache Curator
```
eolivelli, line 1: The target directory must not be committed

eolivelli (changes requested):
Nice work.
I left some comment about files that we don't need.

Apart from that I think it is better to also add a test case that uses some basic curator functions and verify that otel tracks them

curator-client/pom.xml
```diff
@@ -26,11 +26,11 @@
     <parent>
         <groupId>org.apache.curator</groupId>
         <artifactId>apache-curator</artifactId>
-        <version>5.5.1-SNAPSHOT</version>
+        <version>5.6.1-SNAPSHOT</version>
```
eolivelli, line 29: we should bump the version in a separate commit, can you please revert ?

eolivelli (approved):
Overall LGTM

but please revert the change to the pom versions

eolivelli:
I have asked for review on the mailing list

eolivelli:
also please check CI

curator-drivers/open-telemetry/src/main/java/org/apache/curator/drivers/opentelemetry/OpenTelemetryTracingDriver.java
```diff
@@ -0,0 +1,185 @@
...
+        }
+        if (trace.getSessionId() >= 0) {
+            span.setAttribute(SESSION_ID_ATTRIBUTE, trace.getSessionId());
+        }
+        if (trace.getRequestBytesLength() > 0) {
+            span.setAttribute(REQUEST_BYTES_LENGTH_ATTRIBUTE, trace.getRequestBytesLength());
+        }
+        // We will close this in endTrace
+        @SuppressWarnings("MustBeClosedChecker")
+        Scope scope = span.makeCurrent();
```
kezhuw, line 111: Curator invokes callbacks asynchronously.

The timing of callback invocation is indeterministic.  `Span::makeCurrent` in `startTrace` and `Scope::close` in `endTrace` probably will not form try-with-resources as suggested in javadoc of `Span::makeCurrent`.

Also, `endTrace` is probably not invoked in thread where `startTrace` called as far as I known. Given `CuratorFramework::getChildren` as an example, `startTrace` is invoked in Curator threads while `endTrace` is invoked in ZooKeeper thread.

Since `startTrace` and `endTrace` are the only interceptions, I think we can drop `scope`/`context` propagation here.

curator-drivers/open-telemetry/src/main/java/org/apache/curator/drivers/opentelemetry/OpenTelemetryTracingDriver.java
```diff
@@ -0,0 +1,185 @@
...
+            span.setStatus(StatusCode.OK);
+        } else {
+            span.setStatus(StatusCode.ERROR);
+            span.recordException(
+                    KeeperException.create(KeeperException.Code.get(trace.getReturnCode())),
+                    Attributes.of(PATH_ATTRIBUTE, trace.getPath())
+            );
+        }
+
+        span.setAttribute(LATENCY_MS_ATTRIBUTE, trace.getRequestBytesLength());
```
kezhuw, line 131: Should be `trace.getLatencyMs()` ?

PR #496, opened 2024-01-31
CURATOR-699. Upgrade ZooKeeper version to 3.9

tisonkun:


curator-recipes/src/test/java/org/apache/curator/framework/recipes/leader/ChaosMonkeyCnxnFactory.java
```diff
@@ -81,22 +79,19 @@ public void submitRequest(Request si) {
...
+            log.debug("Applied : {}", si);
             super.submitRequest(si);
             // Raise an error if a lock is created
             if ((si.type == ZooDefs.OpCode.create) || (si.type == ZooDefs.OpCode.create2)) {
-                CreateRequest createRequest = new CreateRequest();
                 try {
-                    ByteBuffer duplicate = si.request.duplicate();
-                    duplicate.rewind();
-                    ByteBufferInputStream.byteBuffer2Record(duplicate, createRequest);
+                    CreateRequest createRequest = si.readRequestRecord(CreateRequest::new);
```
tisonkun, line 94: This may cause failures on early ZK versions and we need a testing compatible layer.

Let the CI verify if it's the case.

tisonkun: OK. It fails.

Let me think if we should drop support <= 3.7 or add a test compatible switch.

cc @kezhuw @eolivelli @cammckenzie @Randgalt

eolivelli:
I think that it is fine to drop support for 3.7 in those tests

tisonkun:
Make sense. I understand it as introducing a new test tag said `zk38OrLater` to turn on related tests only when zk 3.8+ as dep.

tisonkun:
```
[ERROR] Tests run: 3, Failures: 0, Errors: 3, Skipped: 0, Time elapsed: 0.44 s <<< FAILURE! - in org.apache.curator.framework.imps.TestTransactionsOld
[ERROR] org.apache.curator.framework.imps.TestTransactionsOld.testWithCompression  Time elapsed: 0.144 s  <<< ERROR!
java.lang.NoSuchMethodError: 'com.google.common.base.Predicate org.apache.curator.framework.api.transaction.CuratorTransactionResult.ofTypeAndPath(org.apache.curator.framework.api.transaction.OperationType, java.lang.String)'
	at org.apache.curator.framework.imps.TestTransactionsOld.testWithCompression(TestTransactionsOld.java:165)
```

It seems because we relocate these methods in curator-framework but not for the test-jar. I don't find a way to relocate the test jar now ..

tisonkun:
I create a monkey patch to work around it - e23d087285edeb45d4dd925379a6af76777cb493

eolivelli (approved):
+1

tisonkun:
@eolivelli Thanks for your review. I make this patch work now. Let me create the ticket and a few follow-ups to improve the case.

PR #500, opened 2024-05-10
CURATOR-696. Fix double leader for LeaderLatch

tisonkun:
A possible fix to CURATOR-696.

* Refer to https://github.com/apache/druid/issues/16411.
* Refer to #430.

I'm thinking whether this has more corner cases ...

cc @gianm @kezhuw @eolivelli

### DIFF
--- a/curator-recipes/src/main/java/org/apache/curator/framework/recipes/leader/LeaderLatch.java
+++ b/curator-recipes/src/main/java/org/apache/curator/framework/recipes/leader/LeaderLatch.java
@@ -509,7 +509,7 @@ public class LeaderLatch implements Closeable {
                         getChildren();
                     }
                 } else {
-                    log.error("getChildren() failed. rc = " + event.getResultCode());
+                    log.error("getChildren() failed. rc = {}", event.getResultCode());
                 }
             }
         };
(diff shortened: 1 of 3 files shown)


curator-recipes/src/main/java/org/apache/curator/framework/recipes/leader/LeaderLatch.java
```diff
@@ -548,43 +548,58 @@ private void checkLeadership(List<String> children) throws Exception {
...
+                                event.getStat() != null ? event.getStat().getEphemeralOwner() : -1;
+                        final long thisSessionId =
+                                client.getZookeeperClient().getZooKeeper().getSessionId();
+                        if (ephemeralOwner != thisSessionId) {
+                            // this node is gone - reset
+                            reset();
+                        } else {
+                            lastPathIsLeader.set(localOurPath);
+                            setLeadership(true);
                         }
```
tisonkun, line 569: This is the major fix.

asdf2014 (approved):
LGTM, thanks for the quick fix

tisonkun:
```
Fri, 10 May 2024 08:03:01 GMT
[INFO] Running org.apache.curator.framework.recipes.leader.TestLeaderLatch
Fri, 10 May 2024 10:33:49 GMT
Error: The operation was canceled.
```

Test timeout now. It seems we introduce new edge cases. Investigating ...

kezhuw (approved):
+1

I think we can also set up a watcher for `ourPath` to gain ability to "perceive" external deletion (from human ?). It is separated anyway.

tisonkun:
@kezhuw Thanks for your review.

> a watcher for ourPath to gain ability to "perceive" external deletion 

Curator has a basic assumption that we own the node. This patch fixes a bug that we don't manage the node (ephemeral owner, a.k.a, session id) properly. We may not handle "external deletion".

And such a watcher doesn't fix our issue here, because it can take some time the deletion pass to the client before it gives up the leadership.

eolivelli (commented):
Do you think that it is possible to add a test case?

tisonkun:
@eolivelli pushed. Ensure that it fails without this patch and passes with this patch.

tisonkun:
Although, the test is not quite accurate since it emulates the session expire manner in the client side, but it's hard to emulate the session expire (ephemeral deletions, exactly for latch0 but not for latch1) in the server side.

tisonkun:
image

Reproduced with a more handy tests. The test pushed above can prevent regression already.

curator-recipes/src/test/java/org/apache/curator/framework/recipes/leader/TestLeaderLatch.java
```diff
@@ -208,6 +213,59 @@ public void testWatchedNodeDeletedOnReconnect() throws Exception {
...
+            // this call is time-consuming but necessary because we don't have a handle to detect the end of the reset
+            // call
+            timing.forWaiting().sleepABit();
+
+            assertTrue(latch0.hasLeadership());
+            assertFalse(latch1.hasLeadership());
+
+            client0.getZookeeperClient().getZooKeeper().getTestable().injectSessionExpiration();
+
+            assertThat(events.poll(timing.forWaiting().milliseconds(), TimeUnit.MILLISECONDS))
```
kezhuw, line 256: Is it better to separate events for latch0 and latch1 and assert that the event (whatever it is) after `LOST_LEADERSHIP` is not equal to `new TestEvent(id0, TestEventType.GAINED_LEADERSHIP)` ? This way it might be clear that we are asserting that no `GRAINED_LEADERSHIP` for `id0` after session changed.

I was thinking this could pass without a run and realized that if failed in old code since there was a `GRAINED_LEADERSHIP` to `id0` between the two.

kezhuw (approved):
Thank you for the hard work! A minor suggestion left.

curator-recipes/src/test/java/org/apache/curator/framework/recipes/leader/TestLeaderLatch.java
tisonkun: This makes sense. I just reuse the test structure and didn't tune a lot.

Would you help in refactor this test case? I suggest you can directly push commit to this branch.

tisonkun:
I'll merge this patch later this week if no more inputs.

I'd still seek possibility to reduce the test consumption since it now takes 140+ minutes to finish ...

curator-recipes/src/test/java/org/apache/curator/framework/recipes/leader/TestLeaderLatch.java
tisonkun: Thanks for your help! I saw the commit.

razinbouzar:
@tisonkun do you know when the 5.7.0 release will be made available?

tisonkun:
@razinbouzar I'll start the discussion today and hopefully vote in two to three weeks. You can keep an eye on the dev mailing list (dev@curator.apache.org).

tisonkun:
@razinbouzar Curator 5.7.0 has released.

PR #501, opened 2024-05-15
CURATOR-705: Add additional test case to TestServiceCache

naude-r:
Highlight the potential issues as described in CURATOR-705.

### DIFF
--- a/curator-x-discovery/src/test/java/org/apache/curator/x/discovery/TestServiceCache.java
+++ b/curator-x-discovery/src/test/java/org/apache/curator/x/discovery/TestServiceCache.java
@@ -27,10 +27,13 @@ import java.io.Closeable;
 import java.util.Collection;
 import java.util.Collections;
 import java.util.List;
+import java.util.concurrent.ArrayBlockingQueue;
 import java.util.concurrent.CountDownLatch;
 import java.util.concurrent.Executors;
+import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Semaphore;
 import java.util.concurrent.TimeUnit;
+import java.util.concurrent.ThreadPoolExecutor;
 import org.apache.curator.framework.CuratorFramework;
 import org.apache.curator.framework.CuratorFrameworkFactory;
 import org.apache.curator.framework.state.ConnectionState;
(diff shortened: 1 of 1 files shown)


curator-x-discovery/src/test/java/org/apache/curator/x/discovery/TestServiceCache.java
```diff
@@ -114,6 +117,81 @@ public void stateChanged(CuratorFramework client, ConnectionState newState) {}
...
+                    .port(10065)
+                    .build();
+            ServiceInstance<String> instance3 = ServiceInstance.<String>builder()
+                    .payload("test")
+                    .name("test")
+                    .port(10066)
+                    .build();
+            discovery.registerService(instance2);
+            discovery.registerService(instance3);
+            assertEquals(3, cache.getInstances().size());
```
kezhuw, line 177: Adding `Thread.sleep(500)` before assertions pass them.

I think it is about freshness, and the javadoc says:

>  NOTE: there is no guarantee of freshness.

Similar, `testInitialLoad` asserts after `latch.await`.

kezhuw (commented):
I left comments inline related to concrete assertions.

Regarding to the `CuratorFramework::runSafe`, it is only used for notification purpose. No hard works are run on it. I changed the executor to `Executors.newSingleThreadExecutor()`, and all behave well.

curator-x-discovery/src/test/java/org/apache/curator/x/discovery/TestServiceCache.java
kezhuw: Swapping the order of assertion and `latch.await` is sufficient for this to pass. Then it is same as `testInitialLoad`.

naude-r: @kezhuw thank you for taking the time. in a real application we cannot afford to wait and we do not know if any instances are available. failing faster would be better.

you are right about the freshness. guess we were spoiled with ` PathChildrenCache::start(BUILD_INITIAL_CACHE)` and the synchronicity thereof.

kezhuw: `ServiceCache::start` is a blocking operation, immediate `getInstances` reflects initial cache after start. `ServiceCacheListener::cacheChanged` will be called every time there is a change after start.

kezhuw:
Close with same reason as jira.

But I think we can improve the javadoc a bit to state:
* `ServiceCache::start` blocks until fully initialized with all existing services for `getInstances`.
* `ServiceCacheListener::cacheChanged` will be called every time there is a change after `ServiceCache::start`.

PR #505, opened 2024-08-19
CURATOR-714: Assert that namespace paths are created using CuratorFramework's ACLProvider

kezhuw:
CURATOR-221 reported that namespace paths are created with no ACL. Though, I belive that it should be fixed by CURATOR-222. It should still be a good to write tests to assert that.


eolivelli (approved):
+1

PR #506, opened 2024-08-31
CURATOR-715/mkdirs/bottom up existence check

pfcoperez:
This PR tackles the problem reported in https://issues.apache.org/jira/browse/CURATOR-715.

Namely, that ZooKeeper trees where upper nodes have restricted READ ACLs, it is not possible to use `creatingParentsIfNeeded` without changing the ACLs. 

This is because the implementation prior to this patch iterated from root node to the newly created one, checking existence of all intermediate nodes.

This patch changes that approach from top-down to bottom-up and the first ancestor that exists is where the validation stops thus avoiding `exists` checks on the higher nodes with more restricted access.

This is only relevant since https://issues.apache.org/jira/browse/ZOOKEEPER-2590

### DIFF
--- a/curator-client/src/main/java/org/apache/curator/utils/ZKPaths.java
+++ b/curator-client/src/main/java/org/apache/curator/utils/ZKPaths.java
@@ -289,7 +289,20 @@ public class ZKPaths {
             throws InterruptedException, KeeperException {
         PathUtils.validatePath(path);
 
-        int pos = 1; // skip first slash, root is guaranteed to exist
+        int pos = path.length();
+        String subPath;
+
+        // This first loop locates the first parent that doesn't exist from leaf nodes towards root
+        // this way, it is not required to have read access on all parents.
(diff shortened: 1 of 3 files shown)


madhava-sridhar:
I would vote up the fix , this is real blocker to anybody upgrading zookeeper to version with fix for [ZOOKEEPER-2590](https://issues.apache.org/jira/browse/ZOOKEEPER-2590)

eolivelli:
Hopefully  we can include this in the next release, looking for a second reviewer
thanks

kezhuw (approved):

tisonkun (commented):
Thanks for your contribution!

Looks reasonable.

pfcoperez:
Thank you @eolivelli @kezhuw and @tisonkun for your reviews :bow: 

Do you have any plans of releasing before the end of the year?

eolivelli:
Please ask on dev@curator.apache.org
We can cut a release

PR #507, opened 2024-09-29
CURATOR-717: Fix typo in CuratorCache Javadoc

pdolif:
There is a typo in the CuratorCache Javadoc:

> **Image** another client that isn't partitioned, deletes "/foo/bar" and then a third client re-creates "/foo/bar"

"Image" should be "Imagine".

### DIFF
--- a/curator-recipes/src/main/java/org/apache/curator/framework/recipes/cache/CuratorCache.java
+++ b/curator-recipes/src/main/java/org/apache/curator/framework/recipes/cache/CuratorCache.java
@@ -39,7 +39,7 @@ import org.apache.curator.framework.listen.Listenable;
  *     <ul>
  *         <li>Pre-network partition the cache contains "/foo" and "/foo/bar"</li>
  *         <li>A network partition occurs and your Curator client loses connection to the server</li>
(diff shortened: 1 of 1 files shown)


tisonkun (approved):
Thanks for your contribution! LGTM.

kezhuw:
I am not a native english speaker. But https://www.dictionary.com/browse/image#american-image-verb-used-with-object says that they are interchange in case of "verb".
> verb (used with object), im·aged, im·ag·ing.    to picture or represent in the mind; [imagine](https://www.dictionary.com/browse/imagine); conceive.

I do search it.
* [image vs imagine both as verbs. What's the difference?](https://english.stackexchange.com/questions/273425/can-we-use-to-image-to-imagine-interchangeably)
* [Can we use "to image" & "to imagine" interchangeably?](https://english.stackexchange.com/questions/273425/can-we-use-to-image-to-imagine-interchangeably)

Both say "image" as a verb. is rare. So I think "imagine" is good to go.

kezhuw (approved):

kezhuw:
@pdolif Thank you! Merged.

PR #508, opened 2024-10-16
CURATOR-710: Fix leaking watch in EnsembleTracker

kezhuw:
CURATOR-667(#474) fixes asynchronous event path for `getConfig` to
"/zookeeper/config" by using `CuratorFramework::usingNamespace(null)` to
fetch data.

It causes watcher not registering to possible `WatcherRemovalManager`,
so leaking in `WatcherRemoveCuratorFramework::removeWatchers`.

### DIFF
--- a/curator-framework/src/main/java/org/apache/curator/framework/imps/Watching.java
+++ b/curator-framework/src/main/java/org/apache/curator/framework/imps/Watching.java
@@ -28,10 +28,12 @@ public class Watching {
     private final CuratorWatcher curatorWatcher;
     private final boolean watched;
     private final CuratorFrameworkImpl client;
+    private WatcherRemovalManager watcherRemovalManager;
     private NamespaceWatcher namespaceWatcher;
 
     public Watching(CuratorFrameworkImpl client, boolean watched) {
         this.client = client;
+        this.watcherRemovalManager = client.getWatcherRemovalManager();
         this.watcher = null;
         this.curatorWatcher = null;
         this.watched = watched;
@@ -39,6 +41,7 @@ public class Watching {
 
     public Watching(CuratorFrameworkImpl client, Watcher watcher) {
         this.client = client;
+        this.watcherRemovalManager = client.getWatcherRemovalManager();
         this.watcher = watcher;
         this.curatorWatcher = null;
         this.watched = false;
@@ -46,6 +49,7 @@ public class Watching {
 
     public Watching(CuratorFrameworkImpl client, CuratorWatcher watcher) {
         this.client = client;
+        this.watcherRemovalManager = client.getWatcherRemovalManager();
         this.watcher = null;
         this.curatorWatcher = watcher;
         this.watched = false;
@@ -53,9 +57,15 @@ public class Watching {
 
     public Watching(CuratorFrameworkImpl client) {
         this.client = client;
+        this.watcherRemovalManager = client.getWatcherRemovalManager();
         watcher = null;
         watched = false;
         curatorWatcher = null;
+    }
+
+    Watching setWatcherRemovalManager(WatcherRemovalManager watcherRemovalManager) {
+        this.watcherRemovalManager = watcherRemovalManager;
+        return this;
     }
 
     Watcher getWatcher(String unfixedPath) {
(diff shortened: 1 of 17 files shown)


curator-framework/src/main/java/org/apache/curator/framework/imps/Watching.java
```diff
@@ -28,36 +28,46 @@ public class Watching {
...
+        this.watcherRemovalManager = client.getWatcherRemovalManager();
         watcher = null;
         watched = false;
         curatorWatcher = null;
     }
 
+    Watching setWatcherRemovalManager(WatcherRemovalManager watcherRemovalManager) {
+        this.watcherRemovalManager = watcherRemovalManager;
+        return this;
+    }
```
tisonkun, line 69: It seems the `watcherRemovalManager` already set in the constructor by `this.watcherRemovalManager = client.getWatcherRemovalManager();`?

tisonkun: Or otherwise we can merge this setter into the constructor as a parameter.

curator-framework/src/main/java/org/apache/curator/framework/imps/Watching.java
```diff
@@ -85,10 +95,8 @@ void commitWatcher(int rc, boolean isExists) {
             doCommit = (rc == KeeperException.Code.OK.intValue());
         }
 
-        if (doCommit && (namespaceWatcher != null)) {
-            if (client.getWatcherRemovalManager() != null) {
-                client.getWatcherRemovalManager().add(namespaceWatcher);
-            }
+        if (doCommit && namespaceWatcher != null && watcherRemovalManager != null) {
```
tisonkun, line 98: Can you elaborate a bit when `watcherRemovalManager` can be different from `client.getWatcherRemovalManager()`.

eolivelli: I agree with the question: we can save some memory by not having the field. The performance impact is negligible

kezhuw: 1. Given a fresh new `CuratorFrameworkImpl`, say `framework0`.
2. Let's assume `framework1 = framework0.newWatcherRemoveCuratorFramework()`. `framework1.getWatcherRemovalManager` will be the `WatcherRemovalManager`.
3. Let's assume `framework2 = frame1.usingNamespace(null)`. `framework2` will loss `WatcherRemovalManager` from above.
5. `framework1.removeWatchers` will not drop watchers from `new Watching(framework2, ...)`.

Step.2 is what exactly `EnsembleTracker` does.
https://github.com/apache/curator/blob/6a27ff0a17df9ecee994ed06acfc2de6cbde58cf/curator-framework/src/main/java/org/apache/curator/framework/imps/EnsembleTracker.java#L87-L90

Step.3 is what exactly `GetConfigBuilderImpl`(CURATOR-667(#474)) does currently.
https://github.com/apache/curator/blob/6a27ff0a17df9ecee994ed06acfc2de6cbde58cf/curator-framework/src/main/java/org/apache/curator/framework/imps/GetConfigBuilderImpl.java#L41-L45

Step.4 is where this bug emerges.
https://github.com/apache/curator/blob/6a27ff0a17df9ecee994ed06acfc2de6cbde58cf/curator-framework/src/main/java/org/apache/curator/framework/imps/EnsembleTracker.java#L99-L104

I have pushed a fixup commit 6b78a3b849b0136d4a45225981c099e8350f4383 with comments to doc this.

kezhuw:
@tisonkun It is ready. I have closed and reopened to revive ci.

curator-framework/src/main/java/org/apache/curator/framework/imps/Watching.java
tisonkun: I get the point now. Let me check the code to see if we can have other less stateful/mutable solution 😆

tisonkun: Here is a patch to remove the later set https://github.com/kezhuw/curator/pull/1.

And I agree that we need some effort for https://github.com/apache/curator/pull/517 to keep the inheritance hierarchy easier to discover and extend.

tisonkun: Briefly, the root cause is `WatcherRemovalFacade::usingNamespace` drop the `removalManager`. Then by convey the manager in the method, things should work well.

kezhuw: I saw the patch. I think it is what I am trying to express in https://github.com/apache/curator/pull/517#issuecomment-2566142916.

> whether CuratorFramework::usingNamespace should inherit functionalities of this ?

It changes the behavior of `CuratorFramework::usingNamespace`, but I am ok.

tisonkun (approved):

PR #509, opened 2024-11-05
[CURATOR-721] Upgrade Guava Listenablefuture

froque:
This removes duplicate class ListenableFuture.
See more at https://github.com/google/guava/issues/7385


eolivelli:
Reopening and closing in order to see if CI kicks in

kezhuw (approved):
LGTM

I think we could even drop directy dependencies to `com.google.guava:listenablefuture`. But it could be a leftover as we probably have more to clean.

PR #510, opened 2024-12-04
CURATOR-719: Fix orSetData for parallel create calls

HoustonPutman:
https://issues.apache.org/jira/browse/CURATOR-719

This fix really just lets the `catch (KeeperException.NodeExistsException e)` clause catch exceptions thrown in the `if (createParentsIfNeeded)` clause. Previously, since the two `catch` statements were for the same try, the `KeeperException.NodeExistsException` logic would not be called if that error was thrown when creating the node after creating the parents. This case can easily be seen when trying to call create.orSetData() in parallel on the same node, while creating parents if needed. This is because the `NoNodeException` would be thrown for both parallel calls, but only one would be able to create the new node in the catch statement, and the other would get an un-catchable `NodeExistsException`.

### DIFF
--- a/curator-framework/src/test/java/org/apache/curator/framework/imps/TestCreate.java
+++ b/curator-framework/src/test/java/org/apache/curator/framework/imps/TestCreate.java
@@ -24,7 +24,11 @@ import static org.junit.jupiter.api.Assertions.*;
 import java.util.Collections;
 import java.util.List;
 import java.util.Optional;
+import java.util.concurrent.Callable;
 import java.util.concurrent.CountDownLatch;
+import java.util.concurrent.Future;
+import java.util.concurrent.LinkedBlockingQueue;
+import java.util.concurrent.ThreadPoolExecutor;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicInteger;
 import org.apache.curator.framework.CuratorFramework;
@@ -33,6 +37,7 @@ import org.apache.curator.framework.api.ACLProvider;
 import org.apache.curator.framework.api.BackgroundCallback;
 import org.apache.curator.framework.api.CreateBuilderMain;
 import org.apache.curator.framework.api.CuratorEvent;
+import org.apache.curator.framework.api.PathAndBytesable;
 import org.apache.curator.retry.RetryOneTime;
 import org.apache.curator.test.BaseClassForTests;
 import org.apache.curator.utils.CloseableUtils;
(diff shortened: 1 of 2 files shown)


kezhuw (approved):

curator-framework/src/test/java/org/apache/curator/framework/imps/TestCreate.java
```diff
@@ -540,6 +564,65 @@ public void testIdempotentCreateConnectionLoss() throws Exception {
...
+            // Check that calling the same create().orSetData() in parallel is ok
+            Callable<Exception> setData = () -> {
+                try {
+                    client.create().orSetData().creatingParentsIfNeeded().forPath(pathWithParents2, data2);
+                } catch (Exception e) {
+                    return e;
+                }
+                return null;
+            };
+            Future<Exception> f1 = executor.submit(setData);
```
eolivelli, line 612: we may be very lucky and not reproduce the problem (if the machine is "slow" and the two operations don't run concurrently)

maybe we can make this a little bit nasty and schedule many operations in parallel.
It would be great to have a way to intercept the code paths and verify that we have exercised all the possible branches. (Mockito?)

eolivelli (changes requested):
Overall LGMT

but maybe the new test may pass not reproduce the issue

curator-framework/src/test/java/org/apache/curator/framework/imps/TestCreate.java
kezhuw: > we may be very lucky and not reproduce the problem (if the machine is "slow" and the two operations don't run concurrently)

For concurrent sensitive test, I usually run with help from IDE "until failure" or "run x times". I, personally, think "reproduce" means that it will reproduce the bug given enough runs(tens to hundreds depends on context).

> Mockito?

I prefer to run multiple times over mock if the number of runs is small. Mock forces us to think different and complicate things.

I tested this after reverting the fix, it failed in the first runs without resorting to "run x times" on my 2015 macbook pro.

eolivelli: with Mockito I mean that we can use a spy or override some method to ensure that we are hitting the expected code path.


> I tested this after reverting the fix, it failed in the first runs without resorting to "run x times" on my 2015 macbook pro

let me try

eolivelli (approved):
I have run the test locally and actually without the fix it fails at the first trial.

We can keep the test as it is now.

thank you @kezhuw

HoustonPutman:
Thanks for checking y'all! I did make sure it failed without the fix, but I know how pesky these types of tests can be on different hardware and runtimes.

> looking forward to more contributions !

I will definitely have a few more in the next few weeks 🙂

PR #513, opened 2024-12-16
CURATOR-726: Improve tracing for certain operations

HoustonPutman:
https://issues.apache.org/jira/browse/CURATOR-726

In Apache Solr, we are trying to refactor all of our custom ZK interaction logic to use Curator. One of the things we do is implement metrics on our interactions with ZK. Curator allows us to do this with Tracing and CuratorListeners. However there are two things that don't give us the information we need.

1. The "Get Children" command trace does not give the information needed for the number of Children fetched.
2. The "Multi Transaction" command trace:
    - The trace does not give the number of commands that are in the transaction, which is something we track currently.
    - The MultiTransaction trace is not an AdvancedTrace, it's a normal Trace for some reason. AdvancedTracerDrivers cannot read regular traces, so we can't even see these. We need to update Curator to use an AdvancedTrace for this command.

Therefore in this PR, I have added the two pieces of information (# of Children fetched and # of transactions) and also converted the MultiTransaciton to use an AdvancedTrace.

### DIFF
--- a/curator-client/src/main/java/org/apache/curator/drivers/OperationTrace.java
+++ b/curator-client/src/main/java/org/apache/curator/drivers/OperationTrace.java
@@ -36,7 +36,7 @@ public class OperationTrace {
     private int requestTransactionCount;
     private long requestBytesLength;
     private long responseBytesLength;
-    private int responseChildrenCount;
+    private int responseChildrenCount = -1;
     private String path;
     private boolean withWatcher;
     private long sessionId;
--- a/curator-client/src/main/java/org/apache/curator/drivers/OperationTrace.java
+++ b/curator-client/src/main/java/org/apache/curator/drivers/OperationTrace.java
@@ -33,8 +33,10 @@ public class OperationTrace {
 
     private int returnCode = KeeperException.Code.OK.intValue();
     private long latencyMs;
+    private int requestTransactionCount;
     private long requestBytesLength;
     private long responseBytesLength;
+    private int responseChildrenCount;
     private String path;
     private boolean withWatcher;
     private long sessionId;
@@ -54,6 +56,11 @@ public class OperationTrace {
 
     public OperationTrace setReturnCode(int returnCode) {
         this.returnCode = returnCode;
+        return this;
+    }
+
+    public OperationTrace setRequestTransactionCount(int transactionCount) {
+        this.requestTransactionCount = transactionCount;
         return this;
     }
 
@@ -97,6 +104,11 @@ public class OperationTrace {
         return this.setResponseBytesLength(data.length);
     }
 
+    public OperationTrace setResponseChildrenCount(int responseChildrenCount) {
+        this.responseChildrenCount = responseChildrenCount;
+        return this;
+    }
+
     public OperationTrace setPath(String path) {
         this.path = path;
         return this;
@@ -124,12 +136,20 @@ public class OperationTrace {
         return this.latencyMs;
     }
 
+    public int getRequestTransactionCount() {
+        return this.requestTransactionCount;
+    }
+
     public long getRequestBytesLength() {
         return this.requestBytesLength;
     }
 
     public long getResponseBytesLength() {
         return this.responseBytesLength;
+    }
+
+    public int getResponseChildrenCount() {
+        return this.responseChildrenCount;
     }
 
     public long getSessionId() {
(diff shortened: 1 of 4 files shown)


tisonkun (approved):
Thank you!

curator-framework/src/main/java/org/apache/curator/framework/imps/GetChildrenBuilderImpl.java
```diff
@@ -165,14 +165,15 @@ public void performBackgroundOperation(final OperationAndData<String> operationA
...
                 public void processResult(int rc, String path, Object o, List<String> strings, Stat stat) {
                     watching.commitWatcher(rc, false);
+                    if (strings == null) {
+                        strings = Lists.newArrayList();
+                    }
                     trace.setReturnCode(rc)
                             .setPath(path)
                             .setWithWatcher(watching.hasWatcher())
                             .setStat(stat)
+                            .setResponseChildrenCount(strings.size())
```
kezhuw, line 175:
```suggestion
                            .setResponseChildrenCount(strings != null ? strings.size() : 0)
```

curator-framework/src/main/java/org/apache/curator/framework/imps/GetChildrenBuilderImpl.java
```diff
@@ -165,14 +165,15 @@ public void performBackgroundOperation(final OperationAndData<String> operationA
                 @Override
                 public void processResult(int rc, String path, Object o, List<String> strings, Stat stat) {
                     watching.commitWatcher(rc, false);
+                    if (strings == null) {
+                        strings = Lists.newArrayList();
+                    }
```
kezhuw, line 170:
```suggestion
```

curator-client/src/main/java/org/apache/curator/drivers/OperationTrace.java
```diff
@@ -33,8 +33,10 @@ public class OperationTrace {
 
     private int returnCode = KeeperException.Code.OK.intValue();
     private long latencyMs;
+    private int requestTransactionCount;
     private long requestBytesLength;
     private long responseBytesLength;
+    private int responseChildrenCount;
```
kezhuw, line 39: Default this to `-1` ?

`0` is a valid count for node with no children. This way, client can choose to not "record" this field in trace event base on the value.

kezhuw (approved):
LGTM

I think we probably should trace also opcode in future. The `name` is not good for programmatical decision.

HoustonPutman:
> I think we probably should trace also opcode in future. The `name` is not good for programmatical decision.

100% agreed! `CuratorEventType` already exists, so we can repurpose it for this as well.

curator-framework/src/main/java/org/apache/curator/framework/imps/GetChildrenBuilderImpl.java
HoustonPutman: I'm actually reverting this change, since I was just moving the call up a few lines. When creating the `CuratorEventImpl`, `strings` was already made to be non-null, so I didn't want to undo that.

HoustonPutman:
> > I think we probably should trace also opcode in future. The `name` is not good for programmatical decision.
> 
> 100% agreed! `CuratorEventType` already exists, so we can repurpose it for this as well.

Well that's very frustrating. `CuratorEventType` is in `curator-framework`, while `OperationTrace` is in `curator-client`, so `OperationTrace` cannot have a `CuratorEventType` variable...

curator-framework/src/main/java/org/apache/curator/framework/imps/CuratorEventImpl.java
```diff
@@ -113,7 +113,7 @@ public String toString() {
                 + opResults + '}';
     }
 
-    CuratorEventImpl(
+    oCuratorEventImpl(
```
kezhuw, line 116:
```suggestion
    CuratorEventImpl(
```

kezhuw (approved):

kezhuw:
> I think we probably should trace also opcode in future.

I guess this require a bunch of changes and worth a separated issue. Let's leave it for future improvement.

curator-framework/src/main/java/org/apache/curator/framework/imps/GetChildrenBuilderImpl.java
kezhuw: I saw that. Thank you!

PR #514, opened 2024-12-16
CURATOR-727: Allow watches to be executed asynchronously

HoustonPutman:
https://issues.apache.org/jira/browse/CURATOR-727

This gives users the ability to run all watches asynchronously, via a provided executorService.

### DIFF
--- a/curator-framework/src/main/java/org/apache/curator/framework/CuratorFrameworkFactory.java
+++ b/curator-framework/src/main/java/org/apache/curator/framework/CuratorFrameworkFactory.java
@@ -27,6 +27,7 @@ import java.util.Arrays;
 import java.util.List;
 import java.util.Objects;
 import java.util.concurrent.Executor;
+import java.util.concurrent.ExecutorService;
 import java.util.concurrent.ThreadFactory;
 import java.util.concurrent.TimeUnit;
 import org.apache.curator.CuratorZookeeperClient;
@@ -171,6 +172,7 @@ public class CuratorFrameworkFactory {
         private SchemaSet schemaSet = SchemaSet.getDefaultSchemaSet();
         private int waitForShutdownTimeoutMs = 0;
         private Executor runSafeService = null;
+        private ExecutorService asyncWatchService = null;
         private ConnectionStateListenerManagerFactory connectionStateListenerManagerFactory =
                 ConnectionStateListenerManagerFactory.standard;
         private int simulatedSessionExpirationPercent = 100;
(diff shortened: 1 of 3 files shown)


dsmiley (commented):
An alternative implementation approach to sp

curator-framework/src/main/java/org/apache/curator/framework/imps/CuratorFrameworkImpl.java
```diff
@@ -445,6 +447,18 @@ public void close() {
             unhandledErrorListeners.clear();
             connectionStateManager.close();
             client.close();
+
+            if (asyncWatchService != null) {
```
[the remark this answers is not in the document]
HoustonPutman: `after publishing the close event`.

That would happen either after `connectionStateManager.close()` or `client.close()`. So this is happening just after those are called.

But I do think we would want to skip any client connection management watches here. Those probably shouldn't be asynchronous anyways. So I'm going to look into that.

tisonkun:
cc @eolivelli @kezhuw @Randgalt @cammckenzie 

Not quite sure if such a change would break sequential assumption.

tisonkun:
Hi @HoustonPutman!

We just released Curator 5.8.0 and it causes several code conflict. To move this PR forward, I'd suggest you rebase this patch or open a new one, referring to https://github.com/apache/curator/issues/1242.

And the key blocker here is to analyze and describe how running watches in a dedicated executor service won't break current sequential implementation. That is, do we depend on the sequential assumption? Or describing the current execution order to ensure that we won't break any assumption.

HoustonPutman:
Thanks for the update @tisonkun. Yeah, there is a lot to think about there, and probably will require a more intelligent feature.

This isn't a blocker for us, so we can move forward as-is with this release in Solr. But as we transfer leader election and other big features to recipes, we can come back and rethink how to approach the watch executor.

curator-framework/src/main/java/org/apache/curator/framework/CuratorFrameworkFactory.java
```diff
@@ -519,6 +521,19 @@ public Builder runSafeService(Executor runSafeService) {
...
 
+        /**
+         * By default, watches are run sequentially.
+         * If an executor is provided here, then all watch calls will be run asynchronously via this executor.
+         * This executor service will be closed when the CuratorFramework is closed.
+         *
+         * @param asyncWatchService executorService to use for all watch calls
+         * @return this
+         */
+        public Builder asyncWatchService(ExecutorService asyncWatchService) {
```
eolivelli, line 532: is there any additional constraint on the ExecutorService ?
can operations be executed in any order ?

what happens in case of rejected execution ?
what happens if the watcher code "throws" ? should the ExecutorService be resilient ?

eolivelli: I would probably name this method "executeWatchesAsync"

eolivelli (changes requested):
coming late to this party, I am sorry

thank you for sharing your problem and a solution

Some issues issues with this patch:
- Can you please add extensive  unit tests ?
- Are there any internal mechansisms in Curator that may break when ALL the watches are executed async ?
- Maybe we should add a way in the API to let the code that register the watch to opt-in the "async" feature. Moving everything to async may introduce unexpected suprises (== bugs in production)

PR #515, opened 2024-12-17
CURATOR-724. Fix LeaderLatch recover on reconnected and missing leaderPath

tisonkun:


curator-recipes/src/main/java/org/apache/curator/framework/recipes/leader/LeaderLatch.java
```diff
@@ -607,6 +607,17 @@ private void getChildren() throws Exception {
...
                     checkLeadership(event.getChildren());
+                } else if (event.getResultCode() == KeeperException.Code.NONODE.intValue()) {
+                    // latchPath has gone - reset
+                    //
+                    // This is possible when RECONNECTED during:
+                    // (1) Scale the zk cluster to 0 nodes.
+                    // (2) Scale it back.
+                    //
+                    // See also https://issues.apache.org/jira/browse/CURATOR-724
+                    reset();
```
kezhuw, line 618: > BTW, before https://github.com/apache/curator/pull/430, the leaderLatch calls reset method after reconnection, and reset will whatever create the missing parent path nodes. So I didn't find this issue in the old versions.   -- [CURATOR-724](https://issues.apache.org/jira/browse/CURATOR-724?focusedCommentId=17901058&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-17901058)

Just for sure, so, in case of `RECONNECTED`, we have no idea what happened to the cluster and `getChildren` is a conservative option comparing to `reset`.

kezhuw (approved):
LGTM

I guess we could write a test for this.

curator-recipes/src/main/java/org/apache/curator/framework/recipes/leader/LeaderLatch.java
tisonkun: Yeah. Updated add a test.

curator-recipes/src/main/java/org/apache/curator/framework/recipes/leader/LeaderLatch.java
```diff
@@ -607,29 +607,38 @@ private void getChildren() throws Exception {
...
-
             case RECONNECTED: {
                 try {
-                    if (client.getConnectionStateErrorPolicy().isErrorState(ConnectionState.SUSPENDED)
-                            || !hasLeadership.get()) {
-                        getChildren();
+                    if (debugHandleReconnectedLatch != null) {
+                        debugHandleReconnectedLatch.await();
                     }
+                    getChildren();
```
tisonkun, line 638: Also it doesn't hurt we always call `getChildren` to check our leadership. It's not reset now anyway.

curator-recipes/src/main/java/org/apache/curator/framework/recipes/leader/LeaderLatch.java
tisonkun: Also ensure that the test failed without this fix.

tisonkun:
@xiaowuyy here is a release candidate for testing out - https://lists.apache.org/thread/5bnym8sycmmnmsz1ggq0cy1jjrrjqyoy

If it looks good, I'll call the release in the next week.

PR #517, opened 2024-12-31
CURATOR-718: Refactor CuratorFramework inheritance hierarchy by composing functionalities

kezhuw:
This closes https://github.com/apache/curator/issues/1233.

Currently, the `CuratorFrameworkImpl` hierarchy is neither simple nor composable.

Ideally, there should be only one instance of `CuratorFrameworkImpl`. Additional functionalities should be added by intercepting methods on purpose, but not by cloning through `CuratorFrameworkImpl(CuratorFrameworkImpl parent)`.

We could take CURATOR-626 and CURATOR-710 as lessons to know how brittle the currently hierarchy.

### DIFF
--- a/curator-framework/src/main/java/org/apache/curator/framework/imps/DelegatingCuratorFramework.java
+++ b/curator-framework/src/main/java/org/apache/curator/framework/imps/DelegatingCuratorFramework.java
@@ -1,0 +1,213 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+package org.apache.curator.framework.imps;
+
+import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.TimeUnit;
+import org.apache.curator.CuratorZookeeperClient;
+import org.apache.curator.framework.CuratorFramework;
+import org.apache.curator.framework.api.ACLProvider;
+import org.apache.curator.framework.api.CompressionProvider;
+import org.apache.curator.framework.api.CuratorEvent;
+import org.apache.curator.framework.api.CuratorListener;
+import org.apache.curator.framework.api.UnhandledErrorListener;
+import org.apache.curator.framework.listen.Listenable;
+import org.apache.curator.framework.schema.SchemaSet;
+import org.apache.curator.framework.state.ConnectionStateErrorPolicy;
+import org.apache.curator.framework.state.ConnectionStateListener;
+import org.apache.curator.utils.ZookeeperCompatibility;
+import org.apache.zookeeper.Watcher;
+import org.apache.zookeeper.server.quorum.flexible.QuorumVerifier;
+
+public class DelegatingCuratorFramework extends InternalCuratorFramework {
+    protected final InternalCuratorFramework client;
+
+    public DelegatingCuratorFramework(InternalCuratorFramework client) {
+        this.client = client;
+    }
+
+    @Override
+    public void start() {
+        throw new UnsupportedOperationException();
+    }
+
+    @Override
+    public void close() {
+        throw new UnsupportedOperationException();
+    }
+
+    @Override
+    public CuratorFrameworkState getState() {
+        return client.getState();
+    }
+
+    @Override
+    @Deprecated
+    public boolean isStarted() {
+        return client.isStarted();
+    }
+
+    @Override
+    public Listenable<ConnectionStateListener> getConnectionStateListenable() {
+        return client.getConnectionStateListenable();
+    }
+
+    @Override
+    public Listenable<CuratorListener> getCuratorListenable() {
+        return client.getCuratorListenable();
+    }
+
+    @Override
+    public Listenable<UnhandledErrorListener> getUnhandledErrorListenable() {
+        return client.getUnhandledErrorListenable();
+    }
+
+    @Override
+    public CuratorFramework usingNamespace(String newNamespace) {
+        return client.usingNamespace(newNamespace);
+    }
+
+    @Override
+    public CuratorZookeeperClient getZookeeperClient() {
+        return client.getZookeeperClient();
+    }
+
+    @Override
+    public ZookeeperCompatibility getZookeeperCompatibility() {
+        return client.getZookeeperCompatibility();
+    }
+
+    @Override
+    @Deprecated
+    public void clearWatcherReferences(Watcher watcher) {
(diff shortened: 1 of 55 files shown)


kezhuw:
I leave CURATOR-710 to #508.

To resolve CURATOR-710, I think we need to decide whether `CuratorFramework::usingNamespace` should inherit functionalities of `this` ? If it is yes, then we probably could fix CURATOR-710 in a straightfoward way (comparing to trick fix in #508).

curator-framework/src/main/java/org/apache/curator/framework/imps/InternalCuratorFramework.java
```diff
@@ -0,0 +1,253 @@
...
+
+/**
+ * This is the internal version of {@link CuratorFramework}.
+ *
+ * <p>Most internal codes should use {@link InternalCuratorFramework} instead of {@link CuratorFrameworkImpl}, so
+ * functionalities could be added additively by overriding methods in {@link DelegatingCuratorFramework}.
+ *
+ * <p>An instance of {@link CuratorFramework} MUST BE an instance of {@link InternalCuratorFramework}.
+ */
+public abstract class InternalCuratorFramework implements CuratorFramework {
```
kezhuw, line 57: This is the most notable change.

`NamespaceFacade` and `WatcherRemovalFacade` are now subclasses of `DelegatingCuratorFramework` and `InternalCuratorFramework` but not `CuratorFrameworkImpl`.

kezhuw (commented):
This is a huge pr. But most changes are migrating references from `CuratorFrameworkImpl` to `InternalCuratorFramework` so we can customize functionalities by overriding methods in `InternalCuratorFramework` but not `CuratorFrameworkImpl`. Now, there will be only one `CuratorFrameworkImpl` per curator client.

curator-recipes/src/test/java/org/apache/curator/framework/recipes/leader/TestLeaderSelectorEdges.java
```diff
@@ -98,15 +97,13 @@ public void flappingTest() throws Exception {
         } finally {
             try {
                 leaderSelector1.close();
-            } catch (IllegalStateException e) {
-                fail(e.getMessage());
```
kezhuw, line 102: This shadow the real cause.

See https://ge.apache.org/s/rcubepxch5ibu/tests/goal/org.apache.curator:curator-recipes:surefire:test@default-test/details/org.apache.curator.framework.recipes.leader.TestLeaderSelectorEdges/flappingTest?top-execution=1

tisonkun:
@kezhuw would you consider this is a release blocker? I may expect to call a release by the end of Feb. and now try to collaborate to converge the ongoing PRs.

kezhuw:
> would you consider this is a release blocker? 

No. It is not a even a bug.

curator-framework/src/main/java/org/apache/curator/framework/imps/DelegatingCuratorFramework.java
```diff
@@ -0,0 +1,213 @@
...
+        return client.getEnsembleTracker();
+    }
+
+    @Override
+    NamespaceFacadeCache getNamespaceFacadeCache() {
+        return client.getNamespaceFacadeCache();
+    }
+
+    @Override
+    public <DATA_TYPE> void processBackgroundOperation(
```
tisonkun, line 204:
```suggestion
    <DATA_TYPE> void processBackgroundOperation(
```

.. and may need a re-format

curator-framework/src/main/java/org/apache/curator/framework/imps/InternalCuratorFramework.java
```diff
@@ -0,0 +1,263 @@
...
```
tisonkun, line 58: Personally, I'd prefer to name it `CuratorFrameworkBase` since it's used outside (as method parameter) rather than internal.

curator-framework/src/main/java/org/apache/curator/framework/imps/DelegatingCuratorFramework.java
```diff
@@ -0,0 +1,213 @@
...
+import org.apache.curator.framework.api.UnhandledErrorListener;
+import org.apache.curator.framework.listen.Listenable;
+import org.apache.curator.framework.schema.SchemaSet;
+import org.apache.curator.framework.state.ConnectionStateErrorPolicy;
+import org.apache.curator.framework.state.ConnectionStateListener;
+import org.apache.curator.utils.ZookeeperCompatibility;
+import org.apache.zookeeper.Watcher;
+import org.apache.zookeeper.server.quorum.flexible.QuorumVerifier;
+
+public class DelegatingCuratorFramework extends InternalCuratorFramework {
```
tisonkun, line 39: Add some class comment for its purpose?

Also this seems another abstract class rather than an instantiable one. And we don't need impl `start` and `close` in such a bias.

curator-framework/src/main/java/org/apache/curator/framework/imps/DelegatingCuratorFramework.java
```diff
@@ -0,0 +1,213 @@
...
+
+    @Override
+    public void start() {
+        throw new UnsupportedOperationException();
+    }
+
+    @Override
+    public void close() {
+        throw new UnsupportedOperationException();
+    }
```
tisonkun, line 54: Or somewhat delegate to `client.start/close` also?

curator-framework/src/main/java/org/apache/curator/framework/imps/DelegatingCuratorFramework.java
tisonkun, line 39: Also can be package private.

tisonkun (commented):
Rest generally looks good to me.

curator-framework/src/main/java/org/apache/curator/framework/imps/DelegatingCuratorFramework.java
kezhuw: Currently, only `CuratorFrameworkImpl::start`/`CuratorFrameworkImpl::close` take effects, `start`/`close` in other facades throw. I think we should keep this behavior unchanged in an refactor.

curator-framework/src/main/java/org/apache/curator/framework/imps/DelegatingCuratorFramework.java
kezhuw: Done.

kezhuw: `DelegatingCuratorFramework` is now `abstract` and `start`/`close` has been moved to subclasses.

curator-framework/src/main/java/org/apache/curator/framework/imps/InternalCuratorFramework.java
kezhuw: Renamed.

By `internal`, I mean it should not be used outside of `curator`. Currently, only `impls`/`details` packages use it. Add `it is public for private usages` for caution.

kezhuw:
@tisonkun Sorry for being so late! I have pushed a fixup commit to solve the review comments.

curator-framework/src/main/java/org/apache/curator/framework/imps/DelegatingCuratorFramework.java
tisonkun: Yeah I noticed that this is the original manner. Then it's fine. (Although I think it's by accident, cc @Randgalt do you remember the reason why we don't delegate here?

kezhuw: Hmm, I don't think `NamespaceFacade::close` should close curator.

PR #518, opened 2025-01-03
CURATOR-728: Not issue ZooKeeper::create if possible in ZkPaths::mkdirs

kezhuw:
`ZkPaths::mkdir("/bar/foo")` will not run into `NoAuthException` if
"/bar/foo" exists, and we have `READ` permission to "/bar/foo" but not
`CREATE` permission to "/bar".

### DIFF
--- a/curator-client/src/main/java/org/apache/curator/utils/ZKPaths.java
+++ b/curator-client/src/main/java/org/apache/curator/utils/ZKPaths.java
@@ -282,28 +282,29 @@ public class ZKPaths {
      */
     public static void mkdirs(
             ZooKeeper zookeeper,
-            String path,
+            final String path,
             boolean makeLastNode,
             InternalACLProvider aclProvider,
             boolean asContainers)
             throws InterruptedException, KeeperException {
         PathUtils.validatePath(path);
 
-        int pos = path.length();
-        String subPath;
-
-        // This first loop locates the first parent that doesn't exist from leaf nodes towards root
+        // This first loop locates the first node that doesn't exist from leaf nodes towards root
         // this way, it is not required to have read access on all parents.
         // This is relevant after https://issues.apache.org/jira/browse/ZOOKEEPER-2590.
 
+        int pos = path.length();
         do {
+            String subPath = path.substring(0, pos);
+            if (zookeeper.exists(subPath, false) != null) {
+                break;
+            }
             pos = path.lastIndexOf(PATH_SEPARATOR_CHAR, pos - 1);
-            subPath = path.substring(0, pos);
-        } while (pos > 0 && zookeeper.exists(subPath, false) == null);
+        } while (pos > 0);
 
         // Start creating the subtree after the longest path that exists, assuming root always exists.
 
-        do {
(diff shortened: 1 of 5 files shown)


curator-client/src/main/java/org/apache/curator/utils/ZKPaths.java
```diff
@@ -282,28 +282,29 @@ public static void mkdirs(ZooKeeper zookeeper, String path, boolean makeLastNode
      */
     public static void mkdirs(
             ZooKeeper zookeeper,
-            String path,
+            final String path,
```
tisonkun, line 285: Params are by default not be mutated. I suppose we don't need final to params.

tisonkun (approved):
Rest LGTM.

curator-client/src/main/java/org/apache/curator/utils/ZKPaths.java
kezhuw, line 285:
```suggestion
            String path,
```

kezhuw:
Hmm, I pushed to `apache/curator`. I will delete this pr branch once merged. Sorry for the mistake.

curator-framework/src/test/java/org/apache/curator/framework/imps/TestCreate.java
```diff
@@ -684,21 +685,23 @@ public void testProtectedUtils() throws Exception {
...
+                    .withACL(Collections.singletonList(new ACL(ZooDefs.Perms.CREATE, ANYONE_ID_UNSAFE)))
                     .forPath("/bat");
 
             // In creation attempts where the parent ("/bat") has ACL that restricts read, creation request fails.
             try {
-                client.create().creatingParentsIfNeeded().forPath("/bat/bost");
+                client.create().creatingParentsIfNeeded().forPath("/bat/foo/bost");
                 fail("Expected NoAuthException when not authorized to read new node grandparent");
             } catch (KeeperException.NoAuthException noAuthException) {
+                assertEquals(noAuthException.getPath(), "/bat");
```
kezhuw, line 704: The old test reported `NoAuth` on creating `/bat/bost`. I think this is not the target CURATOR-715 trying to assert.

kezhuw (commented):
cc @tisonkun @pfcoperez

I made some changes to test since last review. I am going to merge this if no doubt.

pfcoperez (approved):
Thanks for this change @kezhuw . This makes the mkdirs ACLs constraints less restrictive which is a great improvement :pray:

yang-wei:
@kezhuw thank you for fixing this. It helps to unblock our upgrade to Zookeeper 3.9.3. Do you know when will be the next release that include this patch ?

kezhuw:
@yang-wei There is no schedule for next release yet. You could post mail to dev mailing list to let the community know.

tisonkun:
@yang-wei  here is a release candidate for testing out - https://lists.apache.org/thread/5bnym8sycmmnmsz1ggq0cy1jjrrjqyoy

If it looks good, I'll call the release in the next week.

PR #519, opened 2025-01-04
chore: Enable GitHub Issues

tisonkun:
Here we have several open issues:

1. @gmcdonald, will this move stop all the possibility that INFRA will help migrate JIRA tickets? I remember that I was told not to enable GitHub Issues by myself but to wait for INFRA's response. But as you can see, the ticket gets stalled, and we don't know whether it will make progress.

2. The PR title and commit message flavor may change. As we now follow `CURATOR-xxx` where `xxx` is mapped to a JIRA ticket and we have the autolink to JIRA. With GitHub issues, it can be easily linked within the PR description. And thus,
    1. I'd keep the `autolink_jira` option for all existing PRs and commits;
    2. I'd suggest all the following PR/commit use [Conventional Commits](https://www.conventionalcommits.org/en/v1.0.0/)

3. The release note generation step in release process should be updated. I suppose either GitHub releases can autogenerate one, or we maintain a CHANGELOG file and update it per PR. Either should work.

### DIFF
--- a/.asf.yaml
+++ b/.asf.yaml
@@ -26,7 +26,7 @@ github:
     - consensus
   features:
     wiki: false
-    issues: false
+    issues: true
     projects: false
   enabled_merge_buttons:
     squash:  true


kezhuw (approved):
Good to go from my side.

We can do the migration in next step. I am good for jira and github issues to coexist for a while.

tisonkun:
OK. I'll merge this PR in these days.

Let's keep the JIRA project for a while, recommend/ensure users to report new issues on GitHub. Manually migrate existing tickets to GitHub Issues, and finally archive/readonly the JIRA project.

tisonkun:
According to the conversation with the INFRA team on Slack:

> [@tisonkun] If we can turn on GitHub Issues, we can encourage/ensure new issues goes to GitHub Issues, and softly depreacted the JIRA project. But IIRC I was told not to do so because the INFRA team may finish it at once. So I don't know if, by my turning on the GitHub Issues, I'd unexpectedly break some assumption by the INFRA team.
>
> [INFRA member] I dont think it will break anything as such. PRs and Issues both get their numbering scheme from the same pool. You have over 500 PRs already so when importing issues, CURATOR-1 will start at the first available number - currrently `#520`. etc etc. The import process has a mapping system whereas it will link the Issue imported back to the original Jira ticket. If desired a final comment is also added to each Jira ticket pointing to the migrated Github Issue.

So I'd merge this PR now and let's move forward.

Issue #1246, opened 2025-02-20
Asynchronously Initialize Cache Before Reading


kotman12:
The current `CachedModeledFrameworkImpl` doesn't manage cache initialization for you. A perfect example of the kind of code you can expect to see in the wild is in the `CachedModeledFramework` tests themselves, i.e. blocking on semaphores that pin the reading thread to prevent it from certain operations that are cache dependent (but as far as I can tell exactly _which_ operations are cache dependent is not really guaranteed so this is arguably cumbersome). Either way, this is fine in a lot of cases...

However, I propose an additional `InitializedCachedModeledFramework` implementation which asynchronously waits for the cache initialization trigger and only _then_ proceeds to read from the cache. I implemented something similar for my personal use-case where I couldn't rely on, i.e. `readThrough`, to handle the uninitialized case because `readThrough` cannot disambiguate between a znode that is missing because it truly is missing from zk vs one that is missing because the cache hasn't initialized. In my case the znode wouldn't always exist in zk and so using `readThrough` would result in a lot of wasted calls to zk, greatly reducing the benefit of the cache in the first place.

To reiterate `InitializedCachedModeledFramework` has a couple benefits over the existing implementation:

1. No more possibility of misleading `NoNodeException` in the case of reading from `CachedModeledFramework` _before_ the cache has warmed. (I say misleading because the node may exist .. just not in the cache)
2. No more temptation to add blocking semaphores in front of this non-blocking interface.
3. IMO generally less [confusion about how to properly use](https://stackoverflow.com/questions/65328623/how-to-be-notified-when-cachedmodeledframework-is-initialized) this otherwise great(!) feature.

PR #1247, opened 2025-02-20
GH-1246. Asynchronously initialize cache before reading

kotman12:
Linked issue https://github.com/apache/curator/issues/1246

The current `CachedModeledFrameworkImpl` doesn't manage cache initialization for you. A perfect example of the kind of code you can expect to see in the wild is in the `CachedModeledFramework` tests themselves, i.e. blocking on semaphores that pin the reading thread to prevent it from certain operations that are cache dependent (but as far as I can tell exactly _which_ operations are cache dependent is not really guaranteed so this is arguably cumbersome). Either way, this is fine in a lot of cases...

However, I propose an additional `InitializedCachedModeledFramework` implementation which asynchronously waits for the cache initialization trigger and only _then_ proceeds to read from the cache. I implemented something similar for my personal use-case where I couldn't rely on, i.e. `readThrough`, to handle the uninitialized case because `readThrough` cannot disambiguate between a znode that is missing because it truly is missing from zk vs one that is missing because the cache hasn't initialized. In my case the znode wouldn't always exist in zk and so using `readThrough` would result in a lot of wasted calls to zk, greatly reducing the benefit of the cache in the first place.

To reiterate `InitializedCachedModeledFramework` has a couple benefits over the existing implementation:

1. No more possibility of misleading `NoNodeException` in the case of reading from `CachedModeledFramework` _before_ the cache has warmed. (I say misleading because the node may exist .. just not in the cache)
2. No more temptation to add blocking semaphores in front of this non-blocking interface.
3. IMO generally less [confusion about how to properly use](https://stackoverflow.com/questions/65328623/how-to-be-notified-when-cachedmodeledframework-is-initialized) this otherwise great(!) feature.

### DIFF
--- a/curator-x-async/pom.xml
+++ b/curator-x-async/pom.xml
@@ -65,6 +65,12 @@
         </dependency>
 
         <dependency>
+            <groupId>org.junit.jupiter</groupId>
+            <artifactId>junit-jupiter-params</artifactId>
+            <scope>test</scope>
+        </dependency>
+
+        <dependency>
             <groupId>org.slf4j</groupId>
             <artifactId>slf4j-log4j12</artifactId>
             <scope>test</scope>
--- a/curator-x-async/src/main/java/org/apache/curator/x/async/modeled/ModeledFramework.java
+++ b/curator-x-async/src/main/java/org/apache/curator/x/async/modeled/ModeledFramework.java
@@ -94,6 +94,13 @@ public interface ModeledFramework<T> {
     CachedModeledFramework<T> cached(ExecutorService executor);
 
     /**
+     * Return a cached framework which waits for cache to be initialized before accessing it.
+     *
+     * @return wrapped cached framework that waits for initialization.
+     */
+    CachedModeledFramework<T> initialized();
+
+    /**
      * Return mutator APIs that work with {@link org.apache.curator.x.async.modeled.versioned.Versioned} containers
      *
      * @return wrapped instance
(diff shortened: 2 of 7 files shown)


tisonkun (approved):
Thanks for your contribution @kotman12!

My extra consideration is whether the original manner has some benefits or we can even replace the old implement with the new one. Also compatibility is a consideration while we can bump a major version if needed and worthy.

kotman12:
@tisonkun I think I agree .. IMO in a majority of cases the "initialized" variety is what you want even though there may be some minimal overhead of going through/checking two futures instead of one. I don't know at what scale that matters, if at all, but didn't benchmark it so didn't want to jump to conclusions.

I'd be happy to refactor `CachedModeledFrameworkImpl` itself instead of adding another implementation. A benefit of this approach is that tests would be a lot simpler but I'm not sure if this is something that needs to be benchmarked? At any rate, this shouldn't introduce any backwards incompatibilities so not sure a major version bump is even needed to change `CachedModeledFrameworkImpl` to manage initialization.

A third option is to make `InitializedCachedModeledFramework` the default cached implementation returned by `ModeledFramework::cached`. This would make the user-friendlier implementation more discoverable while also giving hardcore users the option to use the original, unitialized version.

Of course I'll defer to the more experienced community here.

kotman12:
@tisonkun / @kezhuw please take a look at https://github.com/apache/curator/pull/1250 when you get a chance. I believe it is a less heavyweight implementation of the same thing we are doing here. Please don't merge this PR because while implementing https://github.com/apache/curator/pull/1250 I realized there are still gaps here. Rather than fix them I'll kindly first refer you to https://github.com/apache/curator/pull/1250 since if that is merged this one can be closed. I believe it should be backwards compatible and nearly identical from a performance standpoint.

PR #1250, opened 2025-03-05
GH-1246. Asynchronously initialize cache before reading II

kotman12:
This is an alternative, simpler implementation to https://github.com/apache/curator/pull/1247 with the same motivation, namely to solve https://github.com/apache/curator/issues/1246

### DIFF
--- a/curator-x-async/src/main/java/org/apache/curator/x/async/modeled/details/CachedModeledFrameworkImpl.java
+++ b/curator-x-async/src/main/java/org/apache/curator/x/async/modeled/details/CachedModeledFrameworkImpl.java
@@ -21,7 +21,6 @@ package org.apache.curator.x.async.modeled.details;
 
 import java.util.List;
 import java.util.Map;
-import java.util.Optional;
 import java.util.concurrent.Executor;
 import java.util.concurrent.ExecutorService;
 import java.util.function.Function;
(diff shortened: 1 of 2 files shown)


tisonkun:
Thanks for preparing this patch! I'd review it by the end of next week (sorry this week is a busy week for me >_<)

tisonkun (approved):
Looks good and very tidy.

Thanks a lot for your contribution!

tisonkun:
I'm waiting for other committers' review (cc @eolivelli @kezhuw @Randgalt @cammckenzie) but would otherwise merge this patch in the next week if there is no more objection.

kotman12:
Thanks for taking a look @tisonkun!

kezhuw (approved):

kotman12:
Hey .. I just did a walk-through of the code again and have a question. Is applying the filter:

```
e -> !e.getKey().isRoot()
                                && e.getKey().parent().equals(client.modelSpec().path())
```

safe to do from `::list()`? I realize I added that behavior by deferring to the newly created `internalChildren` from within `list` which applies this filter that wasn't applied before. I am worried this is a regression not caught by the tests.

If it is ok then it is probably ok to remove the now unused methods in `ModeledCachedImpl:

```
    ZPath basePath() {
        return basePath;
    }

    Map<ZPath, ZNode<T>> currentChildren() {
        return currentChildren(basePath);
    }
```

Either way I think a change is due ... let me know what you think!

tisonkun:
@kotman12 Let's removed the unused method in a follow-up PR.

I'm going to merge this PR after CI gree.

tisonkun:
Merged. Thanks for your contribution @kotman12!

Would you provide a follow-up to remove the unused methods mentioned above?

Issue #1251, opened 2025-03-11
Upgrade Zookeeper to 3.9.3+


aminadinari19:
Currently, curator is using 3.9.2 version of zookeeper when there is a newly-released 3.9.3 it can be upgraded to.

tisonkun:
This is possible. Would you submit a patch for this?

PR #1252, opened 2025-03-11
GH-1251: Upgrade Zookeeper to 3.9.3+

aminadinari19:


curator-test/src/main/java/org/apache/curator/test/TestingZooKeeperMain.java
tisonkun, line 1: Seems these changes cause compile fail. Why do you make these changes?

aminadinari19: @tisonkun Thank-you for taking a look at the PR. It seems like in the newer version of zookeeper, the method that was being overridden has become final, so I just altered the access method to it.

kezhuw (approved):
`TestReadOnly.testConnectionStateNewClient` was flaky after ZOOKEEPER-4508, it should be fixed by ZOOKEEPER-4921 which supposed to be shipped in 3.9.4.

I am ok for it to be merged now. We could update to 3.9.4 once it merged.

* https://github.com/apache/curator/actions/runs/15664430697/job/44126664969#step:6:735
* https://github.com/apache/curator/actions/runs/15664430697/job/44126664971#step:6:752

Issue #1258, opened 2025-03-24
PersistentTTLNode fails to remove node


chevaris:
PersistentTTLNode (Curator 5.8.0 and probably previous versions) has a corner case that prevents that ZNode is deleted when program running the recipe is stop in certain situations.

This is the sequence:
- Start the recipe with TTL of 30secs -> Container Node is created
- Stop the program (or the program crashes in production) that runs the recipe before Touch TTL node is created. This is NOT deterministic and basically a background thread is scheduled to run TTL/2 (by default). In worse case scenario de TTL node could take up to 15 secs in this example to be created

When this is happening the CONTAINER node is never deleted. One option is to increase the touchScheduleFactor, BUT still this solution looks not correct for me.

In my view the recipe should watch the Container Node itself and just when the node is created, the recipe could trigger TOUCH node creation to minimize the opportunity window in which the problem happens.

I attach a test case that shows the problem, and I include a fixed recipe that solves it.

https://github.com/chevaris/curator/commit/6da77252f24841d8f8e85572cad9ac6d86cac5e7

Anyhow, no matter how fast the touch ZNode is added the race condition will be always there, and in my view this is a limitation on the strategy used for this recipe that should be documented.

Regards,

Cheva

kezhuw:
> In my view the recipe should watch the Container Node itself and just when the node is created, the recipe could trigger TOUCH node creation to minimize the opportunity window in which the problem happens.

Is it sufficient to create the container node using "CreateMode.PERSISTENT_WITH_TTL" ?

> I attach a test case that shows the problem, and I include a fixed recipe that solves it.

PR is welcome and better to review.

kezhuw:
MAIL: https://lists.apache.org/thread/t1jcn0wj7krfc9qr3sgn3q44fn3pyotd

chevaris:
| Is it sufficient to create the container node using "CreateMode.PERSISTENT_WITH_TTL" ?

I think it is. It is even simpler solution that the one I provided in the link of the ticket. I will create a PR and include the test

Thanks for the suggestion

chevaris:
I provided a new implementation addressing your comments. Check latest commit

https://github.com/apache/curator/pull/1260/files

PR #1260, opened 2025-04-01
ISSUE 1258

chevaris:
Fix for Issue 1258

Thanks to Kezhu Wang for the suggestion in the solution

### DIFF
--- a/curator-recipes/src/main/java/org/apache/curator/framework/recipes/nodes/PersistentTtlNode.java
+++ b/curator-recipes/src/main/java/org/apache/curator/framework/recipes/nodes/PersistentTtlNode.java
@@ -159,12 +159,14 @@ public class PersistentTtlNode implements Closeable {
         this.client = Objects.requireNonNull(client, "client cannot be null");
         this.ttlMs = ttlMs;
         this.touchScheduleFactor = touchScheduleFactor;
-        node = new PersistentNode(client, CreateMode.CONTAINER, false, path, initData, useParentCreation) {
-            @Override
-            protected void deleteNode() {
-                // NOP
-            }
-        };
+        node =
+                new PersistentNode(
+                        client, CreateMode.PERSISTENT_WITH_TTL, false, path, initData, ttlMs, useParentCreation) {
+                    @Override
+                    protected void deleteNode() {
+                        // NOP
+                    }
+                };
         this.executorService = Objects.requireNonNull(executorService, "executorService cannot be null");
         childPath = ZKPaths.makePath(Objects.requireNonNull(path, "path cannot be null"), childNodeName);
     }
@@ -239,8 +241,8 @@ public class PersistentTtlNode implements Closeable {
 
     /**
      * Call when you are done with the PersistentTtlNode. Note: the ZNode is <em>not</em> immediately
-     * deleted. However, if no other PersistentTtlNode with the same path is running the node will get deleted
-     * based on the ttl.
+     * deleted. However, if no other PersistentTtlNode with the same path is running the node will get
+     * deleted based on the ttl.
      */
     @Override
     public void close() {
--- a/curator-recipes/src/test/java/org/apache/curator/framework/recipes/nodes/TestPersistentTtlNode.java
+++ b/curator-recipes/src/test/java/org/apache/curator/framework/recipes/nodes/TestPersistentTtlNode.java
@@ -134,8 +134,8 @@ public class TestPersistentTtlNode extends CuratorTestBase {
                 client.delete().deletingChildrenIfNeeded().forPath("/test/one");
                 timing.sleepABit();
 
-                // The underlying persistent node should not be able to recreate itself as the lazy parent creation is
-                // disabled
+                // The underlying persistent node should not be able to recreate itself as the lazy parent
+                // creation is disabled
                 assertNull(client.checkExists().forPath(containerPath));
                 assertNull(client.checkExists().forPath(childPath));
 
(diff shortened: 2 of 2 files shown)


curator-recipes/src/test/java/org/apache/curator/framework/recipes/nodes/TestPersistentTtlNode.java
```diff
@@ -187,4 +187,34 @@ public void childEvent(CuratorFramework client, PathChildrenCacheEvent event) th
...
+    @Test
+    public void testTouchNodeNotCreated() throws Exception {
+        try (CuratorFramework client =
+                CuratorFrameworkFactory.newClient(server.getConnectString(), new RetryOneTime(1))) {
+            client.start();
+            final long ttlMs = 1_000L;
+            try (PersistentTtlNode node = new PersistentTtlNode(client, "/test", ttlMs, new byte[0])) {
+                node.start();
+                assertTrue(node.waitForInitialCreate(timing.session(), TimeUnit.MILLISECONDS));
+                // Give some minor time for touch node to be created. Will worked after patch
```
kezhuw, line 200: I think we are testing a situation that "touch" node is not scheduled to be created. May be we can create a overrideable method, say "scheduleTouchTask", to use for test.

curator-recipes/src/test/java/org/apache/curator/framework/recipes/nodes/TestPersistentTtlNode.java
```diff
@@ -187,4 +187,34 @@ public void childEvent(CuratorFramework client, PathChildrenCacheEvent event) th
...
+                for (int i = 1; i <= 5; i++) {
+                    if (client.checkExists().forPath("/test") != null) {
+                        break;
+                    }
+                    Thread.sleep(10L);
+                }
+            }
+        }
+        try (CuratorFramework client1 =
```
kezhuw, line 209: I don't think we need yet another client.

curator-recipes/src/test/java/org/apache/curator/framework/recipes/nodes/TestPersistentTtlNode.java
```diff
@@ -187,4 +187,34 @@ public void childEvent(CuratorFramework client, PathChildrenCacheEvent event) th
...
+                CuratorFrameworkFactory.newClient(server.getConnectString(), new RetryOneTime(1))) {
+            client1.start();
+            assertTrue(client1.blockUntilConnected(2, TimeUnit.SECONDS));
+            Thread.sleep(3_000L);
+            assertNull(client1.checkExists().forPath("/test/touch"));
```
kezhuw, line 214: We could move this before `sleep` to ensure that the "touch" node is not created.

kezhuw (commented):
@chevaris Thank you for your contribution!

I left some inline comments.

kezhuw:
Refs: CURATOR-545, [ZOOKEEPER-3546](https://issues.apache.org/jira/browse/ZOOKEEPER-3546)(apache/zookeeper#1138), #1258

curator-recipes/src/test/java/org/apache/curator/framework/recipes/nodes/TestPersistentTtlNode.java
chevaris: I provided a new implementation with a new approach that is much more accurate in the second commit. Basically I will watch all lifecycle for main and  touch nodes with PersistentWatch

curator-recipes/src/test/java/org/apache/curator/framework/recipes/nodes/TestPersistentTtlNode.java
chevaris: Correct. Fixed in new implementation

curator-recipes/src/main/java/org/apache/curator/framework/recipes/nodes/PersistentTtlNode.java
```diff
@@ -68,6 +69,7 @@ public class PersistentTtlNode implements Closeable {
     private final ScheduledExecutorService executorService;
     private final AtomicReference<Future<?>> futureRef = new AtomicReference<>();
     private final String childPath;
+    private volatile boolean skipTouch; // Just for testing
```
kezhuw, line 72: I would prefer an overridable method(similar to the `deleteNode`) over unused member. Testing code could make it a nop to skip touch node creation.

kezhuw (commented):
Thank you for your contribution!

Looks good in general. I left a comment about the way to inject code for testing purpose.

kezhuw (approved):

curator-recipes/src/main/java/org/apache/curator/framework/recipes/nodes/PersistentTtlNode.java
```diff
@@ -159,44 +160,45 @@ public PersistentTtlNode(
...
-                } catch (Exception e) {
-                    if (!ThreadUtils.checkInterrupted(e)) {
-                        log.debug("Could not touch child node", e);
-                    }
-                }
-            }
-        };
+        final Runnable touchTask = () -> touch();
         Future<?> future = executorService.scheduleAtFixedRate(
                 touchTask, ttlMs / touchScheduleFactor, ttlMs / touchScheduleFactor, TimeUnit.MILLISECONDS);
```
kezhuw, line 203:
```suggestion
        Future<?> future = executorService.scheduleAtFixedRate(
                this::touch, ttlMs / touchScheduleFactor, ttlMs / touchScheduleFactor, TimeUnit.MILLISECONDS);
```

kezhuw (commented):
Would you want to take another look ? @tisonkun @eolivelli @Randgalt 

It has been raised before in CURATOR-545, but end up in a fix in zookeeper side [ZOOKEEPER-3546](https://issues.apache.org/jira/browse/ZOOKEEPER-3546).

Given the fact that `PERSISTENT_WITH_TTL` is capable as `CONTAINER` node(see also
https://github.com/apache/zookeeper/pull/1138#issuecomment-552438248). I think we could create the container node with using `PERSISTENT_WITH_TTL` so we can fix this without reconfig server cluster.

Ideally, there sould be a "touch" command to modify "mtime" of `PERSISTENT_WITH_TTL`. Sadly, [we don't have](https://issues.apache.org/jira/browse/ZOOKEEPER-2169?focusedCommentId=15461288&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-15461288), so the child touch node still sound good to me as it does not `setData` on container node after creation.

tisonkun (approved):
LGTM. Thanks for your contribution @chevaris!

PR #1261, opened 2025-04-07
Document read effects after SharedValue.setValue

### LINKED ISSUE #1254: Redundant updateValue call in SharedValue?
In SharedValue.setValue(byte[] newValue), `updateValue` is [called](https://github.com/apache/curator/blob/3828d99261efbd760f3ec50d215c80c2bc42bbd8/curator-recipes/src/main/java/org/apache/curator/framework/recipes/shared/SharedValue.java#L142) after the value on zookeeper is set. Is this redundant considering that SharedValue already has watcher that will update local value on remote change?

kezhuw:
Closes #1254.

### DIFF
--- a/curator-recipes/src/main/java/org/apache/curator/framework/recipes/shared/SharedCount.java
+++ b/curator-recipes/src/main/java/org/apache/curator/framework/recipes/shared/SharedCount.java
@@ -65,7 +65,10 @@ public class SharedCount implements Closeable, SharedCountReader, Listenable<Sha
     }
 
     /**
-     * Change the shared count value irrespective of its previous state
+     * Change the shared count value irrespective of its previous state.
+     *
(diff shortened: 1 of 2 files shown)


kezhuw:
> I think it's fair to say setValue makes a strong guarantee that after success, any call to getValue will get some more updated value, whether that value is set by the same session or a different one cannot be guaranteed though.

@ryansun96 I document these methods as you suggest. Would you mind take a look and approve if it sound good to you ?

ryansun96 (approved):

ryansun96:
@kezhuw Thank you for the PR! Looks good to me. Sorry for not having my commits ready sooner - will definitely try to contribute more in the future :)

kezhuw:
Merged. @ryansun96 Thank you for reviewing!

PR #1262, opened 2025-04-07
Fix PersistentWatcher not working with NamespaceFacade

### LINKED ISSUE #1259: PersistentWatcher no longer be used with namespaced CuratorFramework instance
We are trying to upgrade to 5.8.0 and found out that we can not use a namespaced instance of CuratorFramework together with CuratorCache.  CuratorCacheImpl uses PersistentWatcher internally that calls client.getCuratorListenable() which is hardcoded to throw an exception in NamespaceFacade.

Is this intentional or a regression?

kezhuw:
`NamespaceFacade` does not support `getCuratorListenable` while #520 use it to listen for `CuratorEventType.CLOSING` to fix CURATOR-729.

This commit exports `CuratorFrameworkBase::client` to retrieve underlying framework client to listen for for `CuratorEventType.CLOSING`.

Fixes #1259.

### DIFF
--- a/curator-framework/src/main/java/org/apache/curator/framework/imps/CuratorFrameworkBase.java
+++ b/curator-framework/src/main/java/org/apache/curator/framework/imps/CuratorFrameworkBase.java
@@ -58,6 +58,11 @@ import org.apache.zookeeper.ZooKeeper;
 public abstract class CuratorFrameworkBase implements CuratorFramework {
     abstract NamespaceImpl getNamespaceImpl();
 
+    /**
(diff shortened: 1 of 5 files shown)


kezhuw:
I think we probably could introduce `Runnable listenEvent(CuratorListener listener)` in long term to deprecate `Listenable<CuratorListener> getCuratorListenable()`. This way we gain more control over implementation. I see no semantics reason to forbide `NamespaceFacade::getCuratorListenable`.

tisonkun (approved):
Thanks for taking care of this! LGTM.

Issue #1263, opened 2025-04-08
PersistentTTLNode Thread leak


chevaris:
The recipe allows to be constructed using a provided ScheduledExecutorService  or using an internally created Executor. The code tries to be generic and when closing the recipe is only cancelling recipe touch task, BUT the Executor is never shutdown. That means that when the Executor is internally created, the Executor Thread will remain there forever.

The recipe MUST distinguish both cases, and when the Executor is internally created is MUST be shutdown when recipe is closed. When the Executor is externally provided the recipe will document that it will not shutdown when closing and it is up to the recipe user to properly shutdown the Executor

chevaris:
https://github.com/apache/curator/pull/1264

PR #1264, opened 2025-04-08
PersistentTTLNode Thread leak

chevaris:
https://github.com/apache/curator/issues/1263

### DIFF
--- a/curator-recipes/src/main/java/org/apache/curator/framework/recipes/nodes/PersistentTtlNode.java
+++ b/curator-recipes/src/main/java/org/apache/curator/framework/recipes/nodes/PersistentTtlNode.java
@@ -61,12 +61,15 @@ public class PersistentTtlNode implements Closeable {
     public static final int DEFAULT_TOUCH_SCHEDULE_FACTOR = 2;
     public static final boolean DEFAULT_USE_PARENT_CREATION = true;
 
+    private static final boolean SHUTDOWN_EXECUTOR = true;
+
     private final Logger log = LoggerFactory.getLogger(getClass());
     private final PersistentNode node;
     private final CuratorFramework client;
     private final long ttlMs;
     private final int touchScheduleFactor;
     private final ScheduledExecutorService executorService;
+    private final boolean shutdownExecutor; // If needed to shut down executorService when closing
     private final AtomicReference<Future<?>> futureRef = new AtomicReference<>();
     private final String childPath;
 
@@ -77,15 +80,7 @@ public class PersistentTtlNode implements Closeable {
      * @param initData data for the node
      */
     public PersistentTtlNode(CuratorFramework client, String path, long ttlMs, byte[] initData) {
-        this(
-                client,
-                Executors.newSingleThreadScheduledExecutor(ThreadUtils.newThreadFactory("PersistentTtlNode")),
-                path,
-                ttlMs,
-                initData,
-                DEFAULT_CHILD_NODE_NAME,
-                DEFAULT_TOUCH_SCHEDULE_FACTOR,
-                DEFAULT_USE_PARENT_CREATION);
+        this(client, path, ttlMs, initData, DEFAULT_USE_PARENT_CREATION);
     }
 
     /**
@@ -105,12 +100,14 @@ public class PersistentTtlNode implements Closeable {
                 initData,
                 DEFAULT_CHILD_NODE_NAME,
                 DEFAULT_TOUCH_SCHEDULE_FACTOR,
-                useParentCreation);
+                useParentCreation,
+                SHUTDOWN_EXECUTOR);
     }
 
     /**
      * @param client the client
      * @param executorService  ExecutorService to use for background thread. This service should be single threaded, otherwise you may see inconsistent results.
+     *                         Recipe will NOT shut down the executor when closing.
      * @param path path for the parent ZNode
      * @param ttlMs max ttl for the node in milliseconds
      * @param initData data for the node
@@ -140,6 +137,7 @@ public class PersistentTtlNode implements Closeable {
     /**
      * @param client the client
      * @param executorService  ExecutorService to use for background thread. This service should be single threaded, otherwise you may see inconsistent results.
+     *                         Recipe will NOT shut down the executor when closing.
      * @param path path for the parent ZNode
      * @param ttlMs max ttl for the node in milliseconds
      * @param initData data for the node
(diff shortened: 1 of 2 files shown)


curator-recipes/src/main/java/org/apache/curator/framework/recipes/nodes/PersistentTtlNode.java
```diff
@@ -61,12 +61,15 @@ public class PersistentTtlNode implements Closeable {
...
 
+    private static final boolean SHUTDOWN_EXECUTOR = true;
+
     private final Logger log = LoggerFactory.getLogger(getClass());
     private final PersistentNode node;
     private final CuratorFramework client;
     private final long ttlMs;
     private final int touchScheduleFactor;
     private final ScheduledExecutorService executorService;
+    private final boolean shutdownExecutor; // If needed to shut down executorService when closing
```
kezhuw, line 72: I saw `CloseableScheduledExecutorService`, and think it fit this use case pretty well.

chevaris: I think so

kezhuw: > Probably I could extend that as part of the solution. I assume that is OK

That would be good!

chevaris: Included new commit.

As commented extended CloseableScheduledExecutorService  , and added some tests there.

Also fixed the recipe  and added some tests (maybe redundant with CloseableScheduledExecutorService tests

curator-recipes/src/test/java/org/apache/curator/framework/recipes/nodes/TestPersistentTtlNode.java
```diff
@@ -237,4 +241,88 @@ void touch() {
...
+                        mainPath,
+                        ttlMs,
+                        new byte[0],
+                        PersistentTtlNode.DEFAULT_CHILD_NODE_NAME,
+                        PersistentTtlNode.DEFAULT_TOUCH_SCHEDULE_FACTOR)) {
+                    node.start();
+                    assertTrue(touchCreatedLatch.await(5 * ttlMs, TimeUnit.MILLISECONDS));
+                    node.getCloseableScheduledExecutorService()
+                            .submit(() ->
+                                    executorThreadId.set(Thread.currentThread().getId()));
```
kezhuw, line 318:
```suggestion
                    node.getCloseableScheduledExecutorService()
                            .submit(() ->
                                    executorThreadId.set(Thread.currentThread().getId())).get();
```

curator-recipes/src/test/java/org/apache/curator/framework/recipes/nodes/TestPersistentTtlNode.java
```diff
@@ -237,4 +241,88 @@ void touch() {
...
+                        touchCreatedLatch.countDown();
+                    }
+                };
+                watcher.getListenable().addListener(listener);
+                watcher.start();
+                final AtomicLong executorThreadId = new AtomicLong();
+                final String threadName = "testThreadName";
+                try (PersistentTtlNode node = new PersistentTtlNode(
+                        client,
+                        Executors.newSingleThreadScheduledExecutor(task -> new Thread(task, threadName)),
```
kezhuw, line 308: Shutdown this executor after assertions ?

curator-recipes/src/test/java/org/apache/curator/framework/recipes/nodes/TestPersistentTtlNode.java
```diff
@@ -237,4 +241,88 @@ void touch() {
...
+                            .isEmpty());
+                }
+                Thread.sleep(10L);
+                assertTrue(getThreadsWithIdAndName(executorThreadId, PersistentTtlNode.TOUCH_THREAD_NAME)
+                        .isEmpty());
+            }
+        }
+    }
+
+    private List<Thread> getThreadsWithIdAndName(final AtomicLong executorThreadId, final String name) {
```
kezhuw, line 280: There will be at most one thread. And I think "assertNotNull/assertNull" is easy to read than "assertFalse(...isEmpty())".

curator-recipes/src/test/java/org/apache/curator/framework/recipes/nodes/TestPersistentTtlNode.java
```diff
@@ -237,4 +241,88 @@ void touch() {
...
+                try (PersistentTtlNode node = new PersistentTtlNode(client, mainPath, ttlMs, new byte[0])) {
+                    node.start();
+                    assertTrue(touchCreatedLatch.await(5 * ttlMs, TimeUnit.MILLISECONDS));
+                    node.getCloseableScheduledExecutorService()
+                            .submit(() ->
+                                    executorThreadId.set(Thread.currentThread().getId()));
+                    assertFalse(getThreadsWithIdAndName(executorThreadId, PersistentTtlNode.TOUCH_THREAD_NAME)
+                            .isEmpty());
+                }
+                Thread.sleep(10L);
```
kezhuw, line 273: No sure 10ms it is enough in testing env. How about export underlying executor for tests to `awaitTermination` ?

kezhuw (commented):
@chevaris Thank you for contribution! Looks good in general. Left some inline comments in reviewing.

curator-recipes/src/test/java/org/apache/curator/framework/recipes/nodes/TestPersistentTtlNode.java
chevaris: You are right. That was missing

curator-recipes/src/test/java/org/apache/curator/framework/recipes/nodes/TestPersistentTtlNode.java
chevaris: I have my doubts here what is best, because awaitTermination assures you that tasks were executed, BUT the Javadoc does NOT confirm that the thread has been destroyed, so we can only rely that will happen eventually

After closing the PersistentTTLNode:
- When executor is internal, probably best is a loop with small sleeps to check that thread disappears. Notice also that CloseExecutorService is in another package and extracting methods for testing requires them to be public that I really do NOT like.
- When executor is provided, probably the best is to use awaitTermination and after that sleep for instance 100 msecs and check that the thread is still there

What do you think?

kezhuw: Both sound good to me!

chevaris: I realized (doing the code) that awaitTermination requires that shutdown/shutdownNow has been previously invoked , otherwise will NOT work as expected . So when executor is provided, after PersistentTTLNode has been closed I will wait for 100ms and check that executor there is still there

Maybe looking for the thread using the threadId and the name is a bit too much, when it is enough to check if the executor isShutdown or NOT.

I will provide commit that is able to check if excutor is shutdown in both cases, and also the thread checks and you provide comments. Now I am more in favour to remove all the code that looks for the thread from stacks traces, BUT just give me your opinion there

kezhuw (approved):
LGTM

chevaris:
I see that zk38 is failing BUT is not related with the changes. TesPersistentTtlNode are passing properly. Not sure how to proceed

kezhuw:
Merged. @chevaris Thank you for your contribution!

PR #1273, opened 2025-08-06
Add debug logging for client port hijack failures in ZooKeeperServerE…

bigduu:
…mbeddedAdapter

Add debug log for some specify situation to help developer identify the issue location

It's will eat the exception if can't find the `ZooKeeperServerEmbeddedImpl` class in some case, it will makes keep trying startup the zk server with same port. so the unit test will throw the address in used without any useful info.
This change add info for investigate


tisonkun:
Agree with @kezhuw.

I'm closing this as invalid. Welcome to provide more information and reopen it/open a new issue/PR if desired @bigduu.

Issue #1277, opened 2025-09-08
SafeNotifyService threads leak in CuratorFrameWorkImpl


fapifta:
[CURATOR-495](https://github.com/apache/curator/issues/1015) introduced a new runSafeService field in CuratorFrameworkImpl class, and this field is either initialized by an external ExecutorService via the builder, or it is created internally within the class.

In the CuratorFrameworkImpl#close method though, this Executor is never closed, so the threads that are opened by the instances are lingering there until the VM is closed by default. Worse, if someone specifies a thread factory to the framework implementation via the builder that produces non-daemon threads, the VM never exits due to the unstopped single thread executor.

fapifta:
I have posted a PR that adds tests to demonstrate what I think is the expected behaviour, please let me know what you think.

PR #1278, opened 2025-09-08
GH-1277 SafeNotifyService threads leak in CuratorFrameWorkImpl

fapifta:
CURATOR-495 introduced a new runSafeService field in CuratorFrameworkImpl class, and this field is either initialized by an external ExecutorService via the builder, or it is created internally within the class.

In the CuratorFrameworkImpl#close method though, this Executor is never closed, so the threads that are opened by the instances are lingering there until the VM is closed by default. Worse, if someone specifies a thread factory to the framework implementation via the builder that produces non-daemon threads, the VM never exits due to the unstopped single thread executor.


fapifta:
@kezhuw Thank you for your review, I added to the JavaDoc of the Builder#runSafeService method, and changed the approach the close waits for these executors to be closed.
I hope the tests now will go through, my understanding is that the failure was not related to the change, but let's see.

kezhuw:
Merged. Thank you all for contribution!

PR #1291, opened 2026-03-23
GH-1290. Fix barrier bypass in DistributedDoubleBarrier due to spurious wakeups or SyncConnected events

### LINKED ISSUE #1290: DistributedDoubleBarrier - Barrier bypass due to spurious wakeups or SyncConnected evenets
# DistributedDoubleBarrier - Barrier bypass due to spurious wakeups or SyncConnected events
Present since Curator 4.1.0 (CURATOR-495).

DistributedDoubleBarrier.enter() can return true before the required number of participants have arrived. A single participant can pass through the barrier with no other participants. We discovered this by tracing test flakiness in an internal software test suite.

The bug has two parts:

1. internalEnter() uses do { ... } while (false), which executes exactly once. After wait() returns, the method exits without re-checking the participant count despite Object.wait() being subject to spurious wakeups.
2. The watcher fires on any ZooKeeper event, including SyncConnected session events. Since CURATOR-495 (4.1.0), the notification is async via runSafe(), so hasBeenNotified can be set to true by a session event during wait(). Since the actual member count isn't checked, this results in a barrier participant proceeding erroneously.

The untimed enter() is not affected - it always takes an unconditional wait() path that doesn't rely on hasBeenNotified.

FluffySkyCat:
Fixes [GH-1290](https://github.com/apache/curator/issues/1290).

Replace the do { ... } while (false) in internalEnter() with a while (true) loop that re-checks the ZooKeeper children count after every wakeup. This ensures the barrier condition is always verified before returning.

Also removes the hasBeenNotified flag as redundant now that the check is performed in the while loop.

### DIFF
--- a/curator-recipes/src/main/java/org/apache/curator/framework/recipes/barriers/DistributedDoubleBarrier.java
+++ b/curator-recipes/src/main/java/org/apache/curator/framework/recipes/barriers/DistributedDoubleBarrier.java
@@ -57,7 +57,6 @@ public class DistributedDoubleBarrier {
     private final int memberQty;
     private final String ourPath;
     private final String readyPath;
-    private final AtomicBoolean hasBeenNotified = new AtomicBoolean(false);
     private final AtomicBoolean connectionLost = new AtomicBoolean(false);
     private final Watcher watcher = new Watcher() {
         @Override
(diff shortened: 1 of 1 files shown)


kezhuw (approved):
LGTM

Issue #1292, opened 2026-04-25
Memory leak in PersistentWatcher introduced in PR #520


vreuland:
Hello,

I have moved to Apache Curator 5.9 recently (from 5.6) and I have observed a memory leak using that new version.
Checking using a heap dump, the leak is clearly coming from PersistentWatcher which is registering a listener on Curator and that listener is not properly removed when the PersistentWatcher is closed.
In this example, I have 1524 registered PersistentWatcher "owned" listeners, a very big majority of them being listeners from PersistentWatcher instances long gone (closed) and discarded...
<img width="1134" height="230" alt="Image" src="https://github.com/user-attachments/assets/0b4a3791-c4b8-4b9a-ae49-af76d44a9c75" />

This issue has been introduced by CURATOR-729 (PR#520): https://github.com/apache/curator/pull/520/changes#diff-03152e1ad97a8fdbe50c309d864922eeb0bad131198922c9d865b08851c8d877R83

The listener created during PersistentWatcher#start should be removed in PersistentWatcher#close.

Best regards,
Vincent

kezhuw:
Closed by #1293.

Thank you for your contribution!

PR #1293, opened 2026-05-03
GH-1292 Fix memory leak in PersistentWatcher

vreuland:
Remove the curator listener registered during `PersistentWatcher#start` when the `PersistentWatcher` is closed to avoid a leak of listeners


eolivelli:
Thanks for your contribution 

Can you please a unit test that reproduces the issue and shows the fix ?

vreuland:
> Thanks for your contribution
> 
> Can you please a unit test that reproduces the issue and shows the fix ?

I have added the requested test. Thanks

kezhuw (approved):

